PR #27, opened 2022-03-04
feat: rename AppsyncFunction to AppsyncResolver and move arguments to $context.arguments

### LINKED ISSUE #24: AppsyncFunction Interface
I love this project and I've been trying to use it in addition with graphql codegen to generate types automatically from the schema. While trying to do this I've had issues with the API for the AppsyncFunction class

The issue I see is that the name of the function parameter influences the result of the mapping template so if I use the code from the example app
This works
```javascript
    this.getPerson = new AppsyncFunction<
      (id: string) => ProcessedPerson | null
    >((_$context, id) => {
      const person = this.personTable.getItem({
        key: {
          id: $util.dynamodb.toDynamoDB(id),
        },
        consistentRead: true,
      });
```

But this doesn't work because the name of the parameter is not the same as the name of the graphql query parameter
```javascript
    this.getPerson = new AppsyncFunction<
      (id: string) => ProcessedPerson | null
    >((_$context, userId) => { // notice the argument name has changed
      const person = this.personTable.getItem({
        key: {
          id: $util.dynamodb.toDynamoDB(userId),
        },
        consistentRead: true,
      });
```

This means that no amount of code generation will help here as even if we successfully generate a type that looks like `(id: string) => ProcessedPerson | null ` but don't use the right parameter name then you will not see a typescript error there.

Ideally I think that the interface for this function could look something like 
```javascript 
AppsyncFunction<TArgs, TResult, TSource = undefined> {
    readonly decl: FunctionDecl<F>;
    constructor(fn: ($context: $Context<Source>, args: TArgs) => TResult);
```
Then the arguments would be used with args.id so slightly less good looking but with destructuring the difference is small, but this has the benefit of better type safety as using args.userId would result in a typescript error.

As an added benefit this would make using functionless with [graphql codegen typescript resolver plugin](https://www.graphql-code-generator.com/plugins/typescript-resolvers) much easier

### LINKED ISSUE #25: $context code generation
I just realized I haven't tested the use of $context in Appsync VTL templates and it probably doesn't work since the plugin will detect it as either an argument or an external Reference. This needs to be fixed.

sam-goodwin:
Fixes #24
Fixes #25

This Pull Request makes changes to the type signature and name of an `AppsyncFunction`:

* Renames `AppsyncFunction` to `AppsyncResolver`. This aligns with AWS Appsync's terminology - technically a Resolver is a collection of 0-many Functions. It also addresses some of the naming concerns that are coming as we implement other services such as Step Functions, see https://github.com/sam-goodwin/functionless/issues/7#issuecomment-1057485056
* Changes the type signature to use an `object` to represent the Resolver's arguments instead of a TypeScript function signature and moves them to the `$context.arguments` variable. This aligns `functionless` 1:1 with Appsync's VTL environment (a tenet we are adopting for the whole project) and also aligns with existing GraphQL tooling.

## Previous Developer Experience
Before, we required the function signature be explicitly defined. This was both good and bad. Good: use a TypeScript function signature to describe the resolver, which "feels native". Bad: there is a redundant declaration of the argument, `id`, which left room for user-error, and it did not align well with exiting tooling for GraphQL resolvers and code generators, see: #24.
```ts
new AppsyncFunction<(id: string) => Person>(($context, id) => {
  return table.get({
    key: {
      id: { S: id }
    }
  });
})
```

## New Developer Experience
After this change, there are two ways to define an `AppsyncResolver`:

1. infer `TArgs` and `TResult` from the function expression:
```ts
new AppsyncResolver(($context: AppsyncContext<{id: string}>) => {
  return table.get({
    key: {
      id: { S: $context.arguments.id }
    }
  });
})
```

2. explicitly define the type and then omit it from the `$context` argument declaration:
```ts
new AppsyncResolver<{id: string>, Person>(($context) => {
  return table.get({
    key: {
      id: { S: $context.arguments.id }
    }
  });
})
```

### `TSource`

In GraphQL, there is the concept of a "source", which represents the parent object containing the field being resolved. In the previous two examples, `TSource` is omitted from the `AppsyncContext` and `AppsyncResolver` type signatures. When it is omitted, its type defaults to `undefined`, meaning no parent - the most common case for root-level fields.

To define `TSource`, you can either add it to the `AppsyncContext` or `AppsyncResolver` type signatures, depending on the developer's preference.

```ts
// option 1 - explicit types in AppsyncContext
new AppsyncResolver(($context: AppsyncContext<{id: string}, TSource>) => {});

// option 2 - explicit types in AppsyncResolver
new AppsyncResolver<{id: string>, Person, TSource>(($context) => {});
```

I personally prefer option 1 - specify it in `AppsyncContext` 

BREAKING CHANGE: AppsyncFunction is renamed to AppsyncResolver and the arguments are now defined as an object instead of as a function signature

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ For example, the below function creates an Appsync Resolver Pipeline with two st
 2. Call the `myFunction` Lambda Function
 
 ```ts
-const getItem = new AppsyncFunction<(id: string) => Item | null>(
+const getItem = new AppsyncResolver<{ id: string }, Item | null>(
   ($context, id) => {
     const item = myTable.get({
       id: $util.toDynamoDB(id),
@@ -129,7 +129,7 @@ Finally, configure the `functionless/lib/compile` TypeScript transformer plugin 
 There are three aspects your need to learn before using Functionless in your CDK application:
 
 1. Appsync Integration interfaces for `Function` and `Table`.
-2. `AppsyncFunction` construct for defining Appsync Resolver with TypeScript syntax.
+2. `AppsyncResolver` construct for defining Appsync Resolver with TypeScript syntax.
 3. Add Resolvers to an `@aws-cdk/aws-appsync-alpha.GraphQLApi`.
 
 ### Appsync Integration interfaces for `Function` and `Table`
@@ -138,7 +138,7 @@ You must wrap your CDK L2 Constructs in the corresponding wrapper class provided
 
 **Function**
 
-The `Function` wrapper annotates an `aws_lambda.Function` with a TypeScript function signature that controls how it can be called from within an `AppsyncFunction`.
+The `Function` wrapper annotates an `aws_lambda.Function` with a TypeScript function signature that controls how it can be called from within an `AppsyncResolver`.
 
 ```ts
 import { aws_lambda } from "aws-cdk-lib";
@@ -151,10 +151,10 @@ const myFunc = new Function<(name: string) => string>(
 );
 ```
 
-Within an [AppsyncFunction](#appsyncfunction), you can use the `myFunc` reference like an ordinary Function:
+Within an [AppsyncResolver](#AppsyncResolver), you can use the `myFunc` reference like an ordinary Function:
 
 ```ts
-new AppsyncFunction(() => {
+new AppsyncResolver(() => {
   return myFunc("my name");
 });
 ```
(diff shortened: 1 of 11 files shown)


mathisobadia:
@sam-goodwin I think this greatly improves the DX and makes understanding what is happening inside the resolver more straightforward. I like that there are two ways to type resolvers, I personally would use option 2 with explicit TResult type to get typescript errors in the case where the returned object is not the right type.

PR #33, opened 2022-03-09
feat: flatten chained map, forEach and reduce for vtl

### LINKED ISSUE #2: Optimize `map`, `forEach` and `reduce` chains.
As of #1 , the logic for `map`, `forEach` and `reduce` chains (such as below) results in redundant use fo `#foreach` which excessively consumes the maximum iteration count of 1000.

This expression results in `n = list.length * 3` iterations when it could be accomplished in `n = list.length`
```ts
list.map(..).map(..).forEach(..)
```

thantos:
When a map precedes a `reduce`, `forEach`, or a `map`, apply the map transformations inline in the same VTL foreach.

Example
```ts
reflect((context: AppsyncContext<{ list: string[] }>) => {
      return context.arguments.list
        .map((item) => `hello ${item}`)
        .map((item) => `hello ${item}`);
    })
```

Before:

```vtl
#set($v1 = [])
#set($v2 = [])
#foreach($item in $context.arguments.list)
#set($v3 = "hello \${item}")
$util.qr($v2.add($v3))
#end
#foreach($item in $v2)
#set($v4 = "hello \${item}")
$util.qr($v1.add($v4))
#end
...
```

After

```vtl
#set($v1 = [])
#foreach($item in $context.arguments.list)
#set($item = "hello \${item}")
#set($v2 = "hello \${item}")
$util.qr($v1.add($v2))
#end
```

Caveat: When the map, reduce, or foreach uses the array parameter, do not flatten.

Bug: Fixed a bug in reduce where the internal `previous` variable name was initialized outside of the VTL foreach and could override a later reduce which also uses the same variable name.

Example:

```ts
    reflect((context: AppsyncContext<{ list: string[] }>) => {
      return context.arguments.list
        .map((item) => `hello ${item}`)
        .reduce((newList: string[], item) => {
          return [...newList, item];
        }, [])
        .map((item) => `hello ${item}`)
        .reduce((newList: string[], item) => {
          return [...newList, item];
        }, []);
    })
```

Before (generated variable names may be wrong, sorry)
```vtl
#set($newList = [])
#foreach($item in $context.arguments.list)
#set($item = "hello \${item}")
#set($v3 = [])
$util.qr($v3.addAll($newList))
$util.qr($v3.add($item))
#set($v2 = $v3)
#set($v1 = $v2)
#end
#set($newList = [])
#foreach($item in $newList)
#set($item = "hello \${item}")
#set($v6 = [])
$util.qr($v6.addAll($newList))
$util.qr($v6.add($item))
#set($v5 = $v6)
#set($newList = $v5)
#end
```

After
```vtl
#set($v1 = [])
#foreach($item in $context.arguments.list)
#set($item = "hello \${item}")
#set($newList = $v1)
#set($v3 = [])
$util.qr($v3.addAll($newList))
$util.qr($v3.add($item))
#set($v2 = $v3)
#set($v1 = $v2)
#end
#set($v4 = [])
#foreach($item in $v1)
#set($item = "hello \${item}")
#set($newList = $v4)
#set($v6 = [])
$util.qr($v6.addAll($newList))
$util.qr($v6.add($item))
#set($v5 = $v6)
#set($v4 = $v5)
#end
```

Closes #2

### DIFF
--- a/src/vtl.ts
+++ b/src/vtl.ts
@@ -193,17 +193,15 @@ export class VTL {
 
             // Try to flatten any maps before this operation
             // returns the first variable to be used in the foreach of this operation (may be the `value`)
-            const [firstVariable, list, render] = this.flattenListMapOperations(
+            const list = this.flattenListMapOperations(
               node.expr.expr,
               value,
+              (firstVariable, list) => {
+                this.add(`#foreach(${firstVariable} in ${list})`);
+              },
               // If array is present, do not flatten the map, this option immediatly evaluates the next expression
               !!array
             );
-
-            this.add(`#foreach(${firstVariable} in ${list})`);
-
-            // Render any flatten maps before the current one
-            render?.();
 
             // Render the body
             const tmp = this.renderMapOrForEachBody(
@@ -243,31 +241,29 @@ export class VTL {
               ? `$${fn.parameters[3].name}`
               : undefined;
 
-            const [firstVariable, list, render] = this.flattenListMapOperations(
-              node.expr.expr,
-              currentValue,
-              // If array is present, do not flatten maps before the reduce, this option immediatly evaluates the next expression
-              !!array
-            );
-
             // create a new local variable name to hold the initial/previous value
             // this is becaue previousValue may not be unique and isn't contained within the loop
             const previousTmp = this.newLocalVarName();
 
-            if (initialValue !== undefined) {
-              this.set(previousTmp, initialValue);
-            } else {
-              this.add(`#if(${list}.isEmpty())`);
-              this.add(
-                `$util.error('Reduce of empty array with no initial value')`
-              );
-              this.add(`#end`);
-            }
+            const list = this.flattenListMapOperations(
+              node.expr.expr,
+              currentValue,
+              (firstVariable, list) => {
+                if (initialValue !== undefined) {
+                  this.set(previousTmp, initialValue);
+                } else {
+                  this.add(`#if(${list}.isEmpty())`);
+                  this.add(
+                    `$util.error('Reduce of empty array with no initial value')`
+                  );
+                  this.add(`#end`);
+                }
 
-            this.add(`#foreach(${firstVariable} in ${list})`);
-
-            // Render any flatten maps before the current one
-            render?.();
+                this.add(`#foreach(${firstVariable} in ${list})`);
+              },
+              // If array is present, do not flatten maps before the reduce, this option immediatly evaluates the next expression
+              !!array
+            );
 
             if (currentIndex) {
               this.add(`#set(${currentIndex} = $foreach.index)`);
@@ -487,14 +483,16 @@ export class VTL {
    * Recursively flattens map operations until a non-map or a map with `array` paremeter is found.
    * Evaluates the expression after the last map.
    *
+   * @param before a method which executes once the
    * @return [firstVariable, list variable, render function]
    */
   private flattenListMapOperations(
     expr: Expr,
     // Should start with $
     returnVariable: string,
+    before: (firstVariable: string, list: string) => void,
     alwaysEvaluate?: boolean
-  ): [string, string, (() => void) | undefined] {
+  ): string {
     if (
       !alwaysEvaluate &&
       expr.kind === "CallExpr" &&
@@ -505,26 +503,26 @@ export class VTL {
 
       const next = expr.expr.expr;
 
-      const [lastValue, list, lastRender] = this.flattenListMapOperations(
+      const list = this.flattenListMapOperations(
         next,
         value,
+        before,
         // If we find array, the next expression should be evaluated.
         // A map which relies on `array` cannot be flattened further as the array will be inaccurate.
         !!array
       );
 
-      return [
-        lastValue,
-        list,
-        () => {
-          lastRender?.();
+      this.renderMapOrForEachBody(expr, list, returnVariable, index, array);
 
-          this.renderMapOrForEachBody(expr, list, returnVariable, index, array);
-        },
-      ];
+      return list;
     }
+
+    const list = this.eval(expr);
+
+    before(returnVariable, list);
+
     // If the expression isn't a map, return the expression and return variable, render nothing
-    return [returnVariable, this.eval(expr), undefined];
+    return list;
   }
 }
 
--- a/src/vtl.ts
+++ b/src/vtl.ts
@@ -175,6 +175,7 @@ export class VTL {
         if (serviceCall) {
           return serviceCall(node, this);
         } else if (
+          // If the parent is a propAccessExpr
           node.expr.kind === "PropAccessExpr" &&
           (node.expr.name === "map" ||
             node.expr.name === "forEach" ||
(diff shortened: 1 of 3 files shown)


sam-goodwin:
Thanks for the change, this is great! Do you have some snippets of code to show the improvement?

src/vtl.ts
```diff
@@ -418,4 +437,78 @@ export class VTL {
     const __exhaustive: never = node;
     return __exhaustive;
   }
+
+  private mapForEachArgs(call: CallExpr) {
+    const fn = call.args.callbackfn as FunctionExpr;
+    return fn.parameters.map((p) => p.name);
+  }
+
+  private renderMapOrForEach(
```
sam-goodwin, line 446: Some docs would be helpful for understanding its purpose.

thantos:
> Thanks for the change, this is great! Do you have some snippets of code to show the improvement?

I added more test cases and updated the existing ones.

thantos:
Though I could add some in the PR. Will do that.

src/vtl.ts
```diff
@@ -418,4 +437,78 @@ export class VTL {
     const __exhaustive: never = node;
     return __exhaustive;
   }
+
+  private mapForEachArgs(call: CallExpr) {
```
sam-goodwin, line 441: Should this be called `getForEachArgNames`? Does it need to be a function on the class? Could it be static function?
```ts
/**
 * Returns the [value, index, array] arguments if this CallExpr is a `forEach`, `map` or `reduce` call.
 */
export function getForEachArgNames(call: CallExpr): string[] {

}
```

src/vtl.ts
```diff
@@ -418,4 +437,78 @@ export class VTL {
...
+  private renderMapOrForEach(
+    call: CallExpr,
+    list: string,
+    // Should start with $
+    returnVariable?: string,
+    index?: string,
+    array?: string
+  ) {
+    if (index) {
+      this.add(`#set($${index} = $foreach.index)`);
```
sam-goodwin, line 455: Why `$${index}` instead of `${index}`? I have been trying to adopt the convention of having variables always contain the `$`.

src/vtl.ts
```diff
@@ -418,4 +437,78 @@ export class VTL {
...
+    return tmp;
+  }
+
+  /**
+   * Recursively dives into map operations until a non-map or a map with arr paremeter is found.
+   * evaluates the expression after the last map.
+   *
+   * @return [firstVariable, list variable, render function]
+   */
+  private divePropAccessMapOperations(
```
sam-goodwin, line 478: The name of this function doesn't immediately tell me what it does.

src/vtl.ts
```diff
@@ -418,4 +437,78 @@ export class VTL {
...
+        next,
+        `$${value}`,
+        !!array
+      );
+
+      return [
+        lastValue,
+        list,
+        () => {
+          lastRender();
```
sam-goodwin, line 505: I'm confused how this function works. What does this callback do?

src/vtl.ts
```diff
@@ -418,4 +437,78 @@ export class VTL {
...
+    ) {
+      const [value, index, array] = this.mapForEachArgs(expr);
+
+      const next = expr.expr.expr;
+
+      // If we find array, stop recursing as array wouldn't be correct in the current map.
+      const [lastValue, list, lastRender] = this.divePropAccessMapOperations(
+        next,
+        `$${value}`,
+        !!array
```
sam-goodwin, line 498: What does this achieve?

src/vtl.ts
```diff
@@ -418,4 +437,78 @@ export class VTL {
...
+        list,
+        () => {
+          lastRender();
+
+          this.renderMapOrForEach(expr, list, returnVariable, index, array);
+        },
+      ];
+    }
+    // If the expression isn't a map, return the expression and return variable, render nothing
+    return [returnVariable, this.eval(expr), () => {}];
```
sam-goodwin, line 512: `() => {}`?

sam-goodwin (changes requested):

src/vtl.ts
thantos: Yeah, it could be static, will rename too.

src/vtl.ts
thantos: I didn't change this, index and array were unformatted in your code too. I can update them though.

src/vtl.ts
thantos: uhh, yeah, will think of a new name...

src/vtl.ts
sam-goodwin: Oh my bad. I changed my mind a few times, left it a bit messy it seems.

If it's easy to do so, please update. No big deal tho.

src/vtl.ts
thantos: Will document. A map, reduce, or forEach that uses `array` cannot be simplified because the array wouldn't be accurate to the point in time as `array` is the previous array, which when flattened, has yet to be computed.

src/vtl.ts
thantos: uhh, yeah, I don't want to apply the `maps` to the VTL until all of them have been resolved. This function returns the variable name to put into the vtl `foreach`, so I want to find the deepest map, build the function to add the map logic to the vtl, return the variable to put in the foreach, and then execute (render) the maps into the `vtl` later. 

I would have done this differently if this class wasn't so stateful. haha

Again, will document.

src/vtl.ts
thantos: The base case shouldn't do anything, can change it to be undefined.

src/vtl.ts
sam-goodwin: Why do you need a callback to be returned? Can you pass a flag into the recursive call to change how it renders instead of calling back?

src/vtl.ts
sam-goodwin: This smells a bit. Makes me think the callback is unnecessary. Base case should be a return from a recursive call instead of an empty callback?

src/vtl.ts
thantos: No, or at least I don't want to. This supports flattening maps for reduce and map/foreach (and maybe filter?), but this method is completely decoupled from the top level reduce/map logic including the `foreach` (map, reduce) and `initial value` (reduce) logic.

I didn't think of the function as a callback, but a closure to control when we render (similar to the body() method in reduce).

src/vtl.ts
thantos: The "callbacks" give the caller control of when the computed content is applied to the mutable state. Its not a callback, but a renderer trigger. 

Another option is to pass a callback that contains the "before" values (the foreach and other things) into this method, but I like that less because its more of a callback pattern than the late rendering closure. 

And the other option is to build an array of VTL statements instead of mutating the cental statements array, then this wouldn't be an issue because the computed statements could be returned and injected in where needed.

test/vtl.test.ts
```diff
@@ -333,18 +333,61 @@ test("chain map over list", () =>
...
 
+test("chain map over list complex", () =>
+  testCase(
+    reflect((context: AppsyncContext<{ list: string[] }>) => {
+      return context.arguments.list
+        .map((item, i, arr) => {
+          const x = i + 1;
+          return `hello ${item} ${x} ${arr.length}`;
+        })
+        .map((item2, ii) => `hello ${item2} ${ii}`);
```
sam-goodwin, line 376: what happens if you name this `i` instead of `ii`?

sam-goodwin: Will it conflict with the other chain calls?

src/vtl.ts
sam-goodwin: I thought the recursive function would recurse until it finds the terminal case. Terminal case would start the `#foreach` and return a string pointing to the result of its computation. The recursive cases would then emit their statements pointing at that.

test/vtl.test.ts
thantos: It's fine because 1. It would be the same value, map can't change i 2. It would overwrite the previous one, 3. The ts compiler will ensure there isn't a conflict.

We may be able to optimize later by proving that a set doesn't make a meaningful change, but just a non functional optimization.

src/vtl.ts
thantos: I'll give it a try.

I was trying to keep the method independent of the calling implementation, where the foreach is part of the parent, not the flattened maps. 

Looking at the outcome, it appears the only part that comes before the render is the foreach, so maybe it's fine. It'll be a bit strange because the foreach will be in the method, but the end will be outside.

test/vtl.test.ts
thantos: The previous test shows what happens.

sam-goodwin (approved):

PR #38, opened 2022-03-15
fix: improve error message when reference is used in reflect

thantos:
Previous error message: `cannot parse declaration-only function: ${text}`
New error message: `Reflect does not support function references: ${text}`

### DIFF
--- a/src/compile.ts
+++ b/src/compile.ts
@@ -58,14 +58,18 @@ export function compile(
         if (isAppsyncResolver(node)) {
           return visitAppsyncResolver(node);
         } else if (isReflectFunction(node)) {
-          return toFunction("FunctionDecl", node.arguments[0]);
+          const func = node.arguments[0];
+          if (ts.isIdentifier(func)) {
+            throw new Error("Reflect does not support function references");
+          }
+          return toFunction("FunctionDecl", func);
         }
         return ts.visitEachChild(node, visitor, ctx);
       }
 
       function isReflectFunction(node: ts.Node): node is ts.CallExpression & {
         arguments: [
-          ts.FunctionExpression | ts.ArrowFunction,
+          ts.FunctionExpression | ts.ArrowFunction | ts.Identifier,
           ...ts.Expression[]
         ];
       } {
@@ -73,6 +77,7 @@ export function compile(
           const exprType = checker.getTypeAtLocation(node.expression);
           const exprDecl = exprType.symbol?.declarations?.[0];
           if (exprDecl && ts.isFunctionDeclaration(exprDecl)) {
+            console.log(exprDecl.name?.text);
             if (exprDecl.name?.text === "reflect") {
               return true;
             }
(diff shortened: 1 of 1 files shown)


src/compile.ts
```diff
@@ -58,21 +58,28 @@ export function compile(
...
-          ts.FunctionExpression | ts.ArrowFunction,
+          ts.FunctionExpression | ts.ArrowFunction | ts.Identifier,
           ...ts.Expression[]
         ];
       } {
         if (ts.isCallExpression(node)) {
           const exprType = checker.getTypeAtLocation(node.expression);
           const exprDecl = exprType.symbol?.declarations?.[0];
           if (exprDecl && ts.isFunctionDeclaration(exprDecl)) {
+            console.log(exprDecl.name?.text);
```
sam-goodwin, line 82: Left over from debugging?

src/compile.ts
```diff
@@ -58,21 +58,28 @@ export function compile(
         if (isAppsyncResolver(node)) {
           return visitAppsyncResolver(node);
         } else if (isReflectFunction(node)) {
-          return toFunction("FunctionDecl", node.arguments[0]);
+          const func = node.arguments[0];
+          if (ts.isIdentifier(func)) {
+            throw new Error(
+              `Reflect does not support function references: ${func.text}`
```
sam-goodwin, line 64: Maybe also add, "please pass an inline function to reflect".

sam-goodwin (changes requested):
Thanks for improving the error message. One thing worth calling out is that this code should not throw at all. Instead, we need to implement a language service plugin to add custom diagnostics. 

https://github.com/sam-goodwin/functionless/issues/10

Throwing in the transformer crashes the typescript compiler and therefore the typescript language server. I've experienced problems in the IDE because of this.

src/compile.ts
sam-goodwin: Also, this requirement isn't only for reflect. It should be for anywhere we convert a function into its AST form.

src/compile.ts
thantos: I removed it in the next commit :-)

src/compile.ts
thantos: Will update the other places to reflect this, also I think a few other statements could act as a reference like Call or property access as you mention in the document. 

How much did you looking into resolving the references as opposed to needing to be inline? (I really want to pass the arrow functions to a test method, but ran into this).

sam-goodwin: I hope we could solve it also - I think this is part of a larger problem of "closure serialization". For example, we would also want to be able to serialize a Lambda Function implementation, including all of the data it captures in the closure and any dependent functions.

```ts
import dep from "some-module";

class Service extends Construct {
  constructor(scope: Construct, id: string, props: Props) {
    new LambdaFunction(this, "Func", (name: string) => {
      return dep.someFunction(name, props.someCapturedClosureData);      
    });
  }
}
```

The ideal state is that all valid NodeJS applications are valid Functionless applications.

sam-goodwin:
Any update here?

sam-goodwin (approved):

PR #39, opened 2022-03-16
feat: event bridge

### LINKED ISSUE #37: AWS Event Bridge
- [x] Operate on a type safe event
- [ ] Define rule
    - [x] Using predicate on the event
    - [ ] Using conditionals inline with targets
    - [ ] Using an existing CDK rule
- [x] Define function to transform event to target service
   - [ ] Using conditionals for rules and many targets
   - [x] Using standalone targets for a single rule
   - [ ] Integrations 
      - [x] Event Bridge
      - [ ] AppSync - https://aws.amazon.com/blogs/mobile/appsync-eventbridge/
      - [ ] ApiGateway #23 
      - [ ] Step Functions #7 
      - [x] Function 
      - [ ] Dynamo 
- [ ] Event bus as service target to publish events
   - [ ] From Step Functions #7
   - [ ] From AppSync - https://github.com/awsed/AppSync2EventBridge/blob/master/index.ts
   - [ ] From API Gateway #23 

### Option 1: As Rule and Target

construct that accepts a bus and a predicate, returns a rule.

```ts
interface MyEventDetail {
    value: string 
}
type MyEvent = EventBusEvent<MyEventDetail, "my-event-detail-type">;

const rule = new EventBridgeRule<MyEvent>(bus, (event): boolean => {
   return event.source == 'lambda' && event.detail.value.startsWith('something');
})
const function = new Function(...);
const target = new EventBridgeTarget<MyEvent>(rule, (event) => {
    function(event.detail)
});
```

### Option 2: As Single Integration

```ts
interface MyEventDetail {
    value: string 
}
type MyEvent = EventBusEvent<MyEventDetail, "my-event-detail-type">;

const function = new Function(...);
new EventBus<MyEvent>(bus, (event, $util) => {
    if($util.matchType<MyEvent>()) {
          function(event.detail);
    }
})
```

### As AppSync target

```ts
import {aws_events} from 'aws-cdl-lib';
import { EventBus } from 'functionless';

cosnt eventBus = new EventBus(new aws_events.EventBus());

const postEvent = new AppsyncResolver<{ id: string }, Item | null>(($context) => {
    eventBus.putEvent($context.arguments);
})
```

thantos:
Not finished... messy, etc

- [x] Patterns
   - [x] Equals (string, number, boolean)
   - [x] Number Range (single)
   - [x] Number Range (aggregated)
   - [x] Prefix (string)
   - [x] List Includes
   - [x] Boolean simplification
   - [x] Null check
   - [x] Undefined/Exists
   - [x] Negation of simple expressions
   - [x] Negation of aggregations
   - [x] && Aggregations
   - [x] || Aggregations
   - [x] Referencing locale scope variables
   - ~~Referencing external scope constants~~ close in favor of #51
- [x] Input Transforms
   - [x] Utility to support [Predefined variables](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-transform-target-input.html#eb-transform-input-predefined)
   - [x] Referencing locale scope variables
   - ~~Referencing external scope constants~~ close in favor of #51
   - [x] Support all default event fields.
- [x] Compile
   - [x] EventBus.when
   - [x] EventBusRule.map  
- [x] Integrations (out of EB, new PR)
   - ~~Dynamo~~ closing in favor of #49
   - [x] Lambda
   - ~~Step Functions~~ closing in favor of #48
   - ~~API Gateway~~ closing in favor of #50
   - ~~AppSync~~ closed in favor of #53
   - [x] Event Bridge (bus to bus)
- ~~Integration Target (into EB)~~
   - ~~AppSync~~ closed in favor of #53
   - ~~API Gateway~~ closed in favor of #50
   - ~~Step Functions~~ closed in favor of #48 
- [x] Update README
- [x] Add complete example to test-app

Closes #37


sam-goodwin:
The title of the Pull Request must follow conventional commit messages. The title will be the commit message when we merge.

src/compile.ts
```diff
@@ -236,14 +248,22 @@ export function compile(
...
+              ? ts.factory.createStringLiteral(checker.typeToString(type))
+              : ts.factory.createIdentifier("undefined"),
           ]);
         } else if (ts.isElementAccessExpression(node)) {
+          const type = checker.getTypeAtLocation(node.argumentExpression);
           return newExpr("ElementAccessExpr", [
             toExpr(node.expression),
             toExpr(node.argumentExpression),
+            type
+              ? ts.factory.createStringLiteral(checker.typeToString(type))
```
sam-goodwin, line 265: Should we convert this to an object instead of a string?

thantos: Yeah, probably. I was just happy to get some type data at all :-). I also don't know if its the right value, this returns `string[]`, `number[]`, `boolean`. An object would allow us to add more forms of the type later, so object seems right.

sam-goodwin: What do we need the types for? We could convert types to Json schema. Typescript's structural typing maps well.

thantos:
> The title of the Pull Request must follow conventional commit messages. The title will be the commit message when we merge.

Uhh yeah, realized it after pushing :-).

src/compile.ts
thantos: Just need to know the type of the property being interacted with. 

`bool` https://github.com/sam-goodwin/functionless/pull/39/files#diff-39df29cf96e7006909d1516b2738e52ab55c0e7e65f10a5e2dee5be1dc96157eR304-R309
`*[]` to know if includes should work
`number` to do numeric ranges.

src/eventbridge.ts
```diff
@@ -0,0 +1,68 @@
...
+  source: Source;
+  detailType: DetailType;
+  detail: T;
+}
+
+type EventHandlerFunction<E extends EventBusEvent = EventBusEvent<any>> = (
+  event: E
+) => void;
+
+export type EventPredicateFunction<
```
sam-goodwin, line 21: What's the difference between this and `EventHandlerFunction`?

src/eventbridge.ts
```diff
@@ -0,0 +1,68 @@
...
+  E extends EventBusEvent = EventBusEvent<any>
+> = (event: E) => void;
+
+export class EventBusRule<T extends EventBusEvent> extends Construct {
```
sam-goodwin, line 25: tsdoc

test/eventpattern.test.ts
```diff
@@ -0,0 +1,540 @@
...
+  deep: {
+    value: string;
+  };
+  bool: boolean;
+  array: string[];
+  numArray: number[];
+  boolArray: boolean[];
+}>;
+
+describe("event pattern", () => {
```
sam-goodwin, line 19: Great unit tests!

src/eventbridge.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export class EventBus<E extends EventBusEvent> extends Construct {
+  /**
+   * This static property identifies this class as an EventBus to the TypeScript plugin.
+   */
+  public static readonly FunctionlessType = "EventBus";
+
+  public readonly decl: FunctionDecl;
+
+  constructor(
```
sam-goodwin, line 54: I wonder if this class should instead just wrap the `aws_events.EventBus`? Does it need to be a Construct? That's what `Table` and `Function` do.

src/eventbridge.ts
```diff
@@ -0,0 +1,68 @@
...
+    id: string,
+    private readonly bus: aws_events.EventBus,
+    readonly fn: EventHandlerFunction
+  ) {
+    super(scope, id);
+    // The Functionless TSC compiler plugin will transform the `EventHandlerFunction` into a FunctionDecl.
+    this.decl = fn as unknown as FunctionDecl;
+  }
+
+  when(id: string, predicate: EventPredicateFunction<E>) {
```
sam-goodwin, line 65: Can we add an example of this to `test-app/src/people-db.ts`? Curious to see how it all comes together in an app along-side the AppSync Resolvers.

sam-goodwin (changes requested):

src/eventbridge.ts
thantos: Handler function will be for the input transformer (to). They may end up the same and I'll remove one.

src/eventbridge.ts
thantos: It was handy to be a construct because it can be the parent of all of the rules and the eb could be from another stack/imported. I'll play around with it once I get things actually working. You mentioned the step function class is a construct?

src/eventbridge.ts
thantos: yeah!

src/eventbridge.ts
sam-goodwin: Yeah it is. I've been flipping and flopping. Good point on it being in a different stack, but couldn't that be achieved by making the rule a Construct?

src/eventbridge.ts
thantos: https://github.com/sam-goodwin/functionless/pull/39/commits/04bde7229e112e66382a74a8052e434b71d63ddd#diff-3bb1081149ef42a1526cc7ecaa1af0a20897ac5e897324e066afca37d8ef383dR48-R67

src/eventbridge.ts
thantos: See latest commit. Rule is a construct, takes scope and id. Event bus is just a wrapper.

https://github.com/sam-goodwin/functionless/pull/39/commits/04bde7229e112e66382a74a8052e434b71d63ddd#diff-3bb1081149ef42a1526cc7ecaa1af0a20897ac5e897324e066afca37d8ef383dR48-R67

src/eventbridge/eventpattern/index.ts
```diff
@@ -0,0 +1,798 @@
...
+  INCLUDES_SEARCH_ELEMENT,
+  OPERATIONS,
+  STARTS_WITH_SEARCH_STRING,
+} from "./constants";
+import { invertBinaryOperator } from "./utils";
+
+/**
+ * https://github.com/sam-goodwin/functionless/issues/37#issuecomment-1066313146
+ */
+export const synthesizeEventPattern = (
```
sam-goodwin, line 64: Is it good practice to have logic in `index.ts`? Took me a while to find this funciton. Would prefer a file, `synth.ts`

src/eventbridge/eventpattern/index.ts
```diff
@@ -0,0 +1,798 @@
...
+      const mergedPattern =
+        pattern ??
+        andMergePattern(key, classDocument1.doc[key], classDocument2.doc[key]);
+      return {
+        doc: {
+          ...doc.doc,
+          [key]: mergedPattern,
+        },
+      };
+    }, {} as PatternDocument);
```
sam-goodwin, line 145: You can also put the type on the argument:
```ts
allKeys.reduce((doc: PatternDocument, key) => ..)
```

sam-goodwin (changes requested):

src/compile.ts
```diff
@@ -58,14 +58,20 @@ export function compile(
         if (isAppsyncResolver(node)) {
           return visitAppsyncResolver(node);
         } else if (isReflectFunction(node)) {
-          return toFunction("FunctionDecl", node.arguments[0]);
+          const func = node.arguments[0];
+          if (ts.isIdentifier(func)) {
+            throw new Error(
+              `Reflect does not support function references: ${func.text}`
```
sam-goodwin, line 64: Should we move this logic to the `toFunction` call since the same restrictions applies to all cases.

src/compile.ts
```diff
@@ -58,14 +58,20 @@ export function compile(
...
+          }
+          return toFunction("FunctionDecl", func);
         }
         return ts.visitEachChild(node, visitor, ctx);
       }
 
       function isReflectFunction(node: ts.Node): node is ts.CallExpression & {
         arguments: [
-          ts.FunctionExpression | ts.ArrowFunction,
+          ts.FunctionExpression | ts.ArrowFunction | ts.Identifier,
```
sam-goodwin, line 74: Why did we add support for this if it's an error?

src/compile.ts
```diff
@@ -219,7 +227,11 @@ export function compile(
           ]);
         } else if (ts.isIdentifier(node)) {
           if (node.text === "undefined" || node.text === "null") {
-            return newExpr("NullLiteralExpr", []);
+            return newExpr("NullLiteralExpr", [
+              ts.factory.createIdentifier(
+                node.text === "undefined" ? "true" : "false"
```
sam-goodwin, line 232: I think you mean:
```ts
ts.factory.createTrue()
ts.factory.createFalse()
```

Interesting idea here - use `true` to identify `undefined` and `false` to identify `null`?  Should we instead just add an `UndefinedLiteral`?

src/compile.ts
sam-goodwin: Cool, we can start with this.

src/compile.ts
```diff
@@ -286,12 +306,22 @@ export function compile(
             toExpr(node.right),
           ]);
         } else if (ts.isPrefixUnaryExpression(node)) {
-          if (node.operator !== ts.SyntaxKind.ExclamationToken) {
+          if (
+            ![
+              ts.SyntaxKind.ExclamationToken,
+              ts.SyntaxKind.MinusToken,
+            ].includes(node.operator)
```
sam-goodwin, line 313: I think it's simpler to write this logic as:
```ts
if (node.operator !== ts.SyntaxKind.ExclamationToken && node.operator !== ts.SyntaxKind.MinusToken) {
  throw ..
}
```

Also, I don't think `.includes` would narrow the type like the `if` would.

src/compile.ts
```diff
@@ -286,12 +306,22 @@ export function compile(
...
+          ) {
             throw new Error(
               `invalid Unary Operator: ${ts.tokenToString(node.operator)}`
             );
           }
-          return newExpr("UnaryExpr", [toExpr(node.operand)]);
+          return newExpr("UnaryExpr", [
+            ts.factory.createStringLiteral(
+              ts.tokenToString(node.operator) ?? ""
```
sam-goodwin, line 321: Hmm, this default `""` seems wrong?

src/compile.ts
```diff
@@ -385,6 +417,8 @@ export function compile(
           ]);
         } else if (ts.isBreakStatement(node)) {
           return newExpr("BreakStmt", []);
+        } else if (ts.isParenthesizedExpression(node)) {
```
sam-goodwin, line 420: Hmm, this is interesting. To ensure precedence is maintained, we should maintain the parentheses with our own `ParanthesizedExpr`.

src/eventbridge/eventpattern/index.ts
```diff
@@ -0,0 +1,798 @@
...
+      return handleUnaryExpression(expr);
+    } else if (isCallExpr(expr)) {
+      return handleCall(expr);
+    } else {
+      throw new Error(`${expr.kind} is unsupported`);
+    }
+    // assertNever(expr);
+  };
+
+  const handleSmnt = (stmt: Stmt): PatternDocument => {
```
sam-goodwin, line 84: typo: `Stmt`

src/eventbridge/eventpattern/index.ts
```diff
@@ -0,0 +1,798 @@
...
+
+/**
+ * https://github.com/sam-goodwin/functionless/issues/37#issuecomment-1066313146
+ */
+export const synthesizeEventPattern = (
+  predicate: FunctionDecl
+): fnls_event_bridge.FnLsEventPattern => {
+  const eventDecl = predicate.parameters[0];
+
+  const handleExpr = (expr: Expr): PatternDocument => {
```
sam-goodwin, line 69: is `eval` a better name than `handle`? We should try and align the terminology in each of the interpreters.

src/eventbridge/eventpattern/index.ts
```diff
@@ -0,0 +1,798 @@
...
+
+  const handleAnd = (expr: BinaryExpr): PatternDocument => {
+    const left = handleExpr(expr.left);
+    const right = handleExpr(expr.right);
+
+    return andDocuments(left, right);
+  };
+
+  const andDocuments = (
+    classDocument1: PatternDocument,
```
sam-goodwin, line 118: what does `classDocument` mean?

src/eventbridge/eventpattern/index.ts
```diff
@@ -0,0 +1,798 @@
...
+    classDocument2: PatternDocument
+  ): PatternDocument => {
+    const allKeys = [
+      ...new Set([
```
sam-goodwin, line 122: TIL: you can use spread on a `Set`.

src/eventbridge/eventpattern/index.ts
```diff
@@ -0,0 +1,798 @@
...
+  };
+
+  const handleAnd = (expr: BinaryExpr): PatternDocument => {
+    const left = handleExpr(expr.left);
+    const right = handleExpr(expr.right);
+
+    return andDocuments(left, right);
+  };
+
+  const andDocuments = (
```
sam-goodwin, line 117: Need some docs to explain the logic.

src/eventbridge/eventpattern/index.ts
```diff
@@ -0,0 +1,798 @@
...
+      return {
+        doc: {
+          ...doc.doc,
+          [key]: mergedPattern,
+        },
+      };
+    }, {} as PatternDocument);
+  };
+
+  const andMergePattern = (
```
sam-goodwin, line 148: Docs to explain the logic.

src/eventbridge/eventpattern/index.ts
```diff
@@ -0,0 +1,798 @@
...
+    // both are documents, lets merge the documents
+    if (isPatternDocument(pattern1) && isPatternDocument(pattern2)) {
+      return andDocuments(pattern1, pattern2);
+      // both are patterns, merge the patterns
+    } else if (!isPatternDocument(pattern1) && !isPatternDocument(pattern2)) {
+      if (isNumericAggregationPattern(pattern1)) {
+        if (isNumericAggregationPattern(pattern2)) {
+          return andNumericAggregation(pattern1, pattern2);
+        }
+        if (isNumericRangePattern(pattern2)) {
```
sam-goodwin, line 162: nit: would prefer to use `else if` even when the branch returns - i think it groups related clauses together better than a separate `if` on a new line.

src/eventbridge/eventpattern/index.ts
```diff
@@ -0,0 +1,798 @@
...
+   */
+  const handleOr = (expr: BinaryExpr): PatternDocument => {
+    const left = handleExpr(expr.left);
+    const right = handleExpr(expr.right);
+
+    return orDocuments(left, right);
+  };
+
+  /**
+   * Event bridge does not support OR across fields, only AND logic
```
sam-goodwin, line 198: Much wow! :/

sam-goodwin: Is there a technical reason or is EB dumb?

src/eventbridge/eventpattern/index.ts
```diff
@@ -0,0 +1,798 @@
...
+    }
+
+    throw Error(`Unsupported numeric range operation: ${op}.`);
+  };
+
+  /**
+   * Any call that we encouter.
+   * Event bridge pattern match calls must all be results at build time.
+   *
+   * Limited to just Includes and StartsWith.
```
sam-goodwin, line 352: This will need to be clearly documented in the `README.md`. Anywhere we can document in the code so it will end up in `tsdoc` generated documentation?

src/eventbridge/eventpattern/index.ts
```diff
@@ -0,0 +1,798 @@
...
+
+    if (!eventReference) {
+      throw new Error("Includes operation must be on a property of the event.");
+    }
+
+    if (
+      isPropAccessExpr(expr.expr.expr) ||
+      isElementAccessExpr(expr.expr.expr)
+    ) {
+      if (expr.expr.expr.type === "number[]") {
```
sam-goodwin, line 409: Nice, this is pretty cool. But also makes me wish we could use the `ts.TypeChecker` and its AST in the CDK code. We'll very quickly want more type information than strings. At the very least, we will need to create Expressions for representing type information in a structured form. Maybe building on top of JSON schema as the underlying structural type system>

I wonder, what happens if someone is aliasing a `string` type?

```ts
type UserId = string
```

What will the type be when converted to a string? `string` or `UserId`?

src/eventbridge/eventpattern/index.ts
```diff
@@ -0,0 +1,798 @@
...
+    }
+  }
+}
+
+/**
+ * Retrieves a string, number, boolean, undefined, or null constant from the given expression.
+ * Wrap the value to not be ambiguous with the undefined value.
+ * When one is not found, return undefined (not wrapped).
+ *
+ * TODO: Support following references.
```
sam-goodwin, line 685: Do you have some examples?

src/eventbridge/eventpattern/index.ts
```diff
@@ -0,0 +1,798 @@
...
+ */
+const getConstant = (
```
sam-goodwin, line 687: this looks like an `eval` function - it evals the `Expr` to a literal value? Call it `evalToConstant`?

src/eventbridge/eventpattern/index.ts
```diff
@@ -0,0 +1,798 @@
...
+  expr: Expr
+): { value: string | number | boolean | undefined | null } | undefined => {
+  if (
+    isStringLiteralExpr(expr) ||
+    isNumberLiteralExpr(expr) ||
+    isBooleanLiteral(expr)
+  ) {
+    return { value: expr.value };
+  } else if (isNullLiteralExpr(expr)) {
+    return { value: expr.undefined ? undefined : null };
```
sam-goodwin, line 697: can you rename this to `expr.isUndefined`?

src/eventbridge/eventpattern/index.ts
```diff
@@ -0,0 +1,798 @@
...
+ *   * documents with more than one field in in - do not support OR accross multiple fields, inverting multiple AND fields is OR
+ *   * aggregate pattern (OR logic) - do not support AND within a field, inverting OR statements becomes AND, the one exception are numeric ranges.
+ */
+const negateDocument = (classDocument: PatternDocument): PatternDocument => {
+  if (Object.keys(classDocument).length > 1) {
+    throw Error(
+      "Can only negate simple statements like equals, doesn't equals, and prefix."
+    );
+  }
+  const [key = undefined] = Object.keys(classDocument.doc);
```
sam-goodwin, line 730: what does this syntax do?
```ts
const [key = undefined]
```

Never seen that before.

src/eventbridge/eventpattern/pattern.ts
```diff
@@ -0,0 +1,244 @@
...
+    if (isPositiveSingleValueRange(pattern)) {
+      return [pattern.lower.value];
+    } else if (isNegativeSingleValueRange(pattern)) {
+      return [{ "anything-but": pattern.lower.value }];
+    }
+    return [
+      {
+        number: [
+          ...(pattern.lower.value !== Number.NEGATIVE_INFINITY
+            ? [pattern.lower.inclusive ? ">=" : ">", pattern.lower.value]
```
sam-goodwin, line 213: having a hard time understanding what the logic is here?

src/eventbridge/eventpattern/types.ts
```diff
@@ -0,0 +1,69 @@
...
+  | AnythingButPattern
+  | NumberPattern
+  | PrefixPattern
+  | ExistsPattern
+)[];
+
+export type Pattern = PatternList | undefined | SubPattern;
+
+/**
+ * As of CDK 2.14, Event Bridge doesn't support the content matchers.
```
sam-goodwin, line 61: Doesn't our `peerDependency` prevent this from happening?

src/eventbridge/index.ts
```diff
@@ -0,0 +1,78 @@
+import { aws_events } from "aws-cdk-lib";
```
sam-goodwin, line 1: Can we adopt the code tenet of reserving `index.its` for `export * from "./file"` instead of putting logic in there?

src/eventbridge/index.ts
```diff
@@ -0,0 +1,78 @@
...
+export class EventBusTarget<T extends EventBusEvent> {
+  target: aws_events.IRuleTarget;
+
+  constructor(func: EventHandlerFunction<T>) {
+    const decl = func as unknown as FunctionDecl;
+    this.target = synthesizeEventBridgeTarget(decl);
+  }
+}
+
+export class EventBusRule<T extends EventBusEvent> extends Construct {
```
sam-goodwin, line 35: I think this belongs in a file `rule.ts` instead of `index.ts.

src/eventbridge/index.ts
```diff
@@ -0,0 +1,78 @@
...
+    target: EventHandlerFunction<E> | EventBusTarget<E>
+  ): EventBusRule<T> {
+    const _target =
+      target instanceof EventBusTarget ? target : new EventBusTarget<E>(target);
+    this.rule.addTarget(_target.target);
+    return this;
+  }
+}
+
+export class EventBus<E extends EventBusEvent> {
```
sam-goodwin, line 67: Let's make sure we thoroughly document these classes with high quality documentation, including examples. The docs here are what will be generated with `tsdoc` and included on the `functionless.org` website.

src/eventbridge/index.ts
```diff
@@ -0,0 +1,78 @@
+import { aws_events } from "aws-cdk-lib";
+import { Rule } from "aws-cdk-lib/aws-events";
+import { Construct } from "constructs";
+import { FunctionDecl } from "../declaration";
+import { synthesizeEventPattern } from "./eventpattern";
+import { synthesizeEventBridgeTarget } from "./targets";
+
+export interface EventBusEvent<
```
sam-goodwin, line 8: Move to `event-bus-event.ts`?

sam-goodwin: Also, thorough documentation with exhaustive examples.

src/eventbridge/index.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    this.rule = new Rule(this, "rule", {
+      eventPattern: pattern as aws_events.EventPattern,
+      eventBus: bus,
+    });
+  }
+
+  target<E extends T>(target: EventHandlerFunction<E>): EventBusRule<T>;
+  target<E extends T>(target: EventBusTarget<E>): EventBusRule<T>;
+  target<E extends T>(
```
sam-goodwin, line 57: Thorough documentation.

src/eventbridge/index.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export class EventBus<E extends EventBusEvent> {
+  /**
+   * This static property identifies this class as an EventBus to the TypeScript plugin.
+   */
+  public static readonly FunctionlessType = "EventBus";
+
+  constructor(private readonly bus: aws_events.EventBus) {}
+
+  when(scope: Construct, id: string, predicate: EventPredicateFunction<E>) {
```
sam-goodwin, line 75: Maybe `target` should be renamed to `then`? "when-then"?

test/util.ts
```diff
@@ -1,4 +1,6 @@
 import { FunctionDecl } from "../src";
+import { synthesizeEventPattern } from "../src/eventbridge/eventpattern";
+import { FnLsEventPattern } from "../src/eventbridge/eventpattern/types";
```
sam-goodwin, line 3: What does `FnLs` mean?

test-app/yarn.lock
```diff
@@ -1416,7 +1416,7 @@ function-bind@^1.1.1:
   resolved "https://registry.yarnpkg.com/function-bind/-/function-bind-1.1.1.tgz#a56899d3ea3c9bab874bb9773b7c5ede92f4895d"
   integrity sha512-yIovAzMX49sF8Yl58fSCWJ5svSLuaibPxXQJFLmBObTuCr0Mf1KiPopGM9NiFjiYBCbfaa2Fh6breQ6ANVTI0A==
 
-"functionless@file:..":
+functionless@../:
```
sam-goodwin, line 1419: hmm, why did this change?

test/eventpattern.test.ts
```diff
@@ -0,0 +1,797 @@
+import { ebEventPatternTestCase, ebEventPatternTestCaseError } from "./util";
+import { EventBusEvent, EventPredicateFunction } from "../src/eventbridge";
+import { reflect } from "../src";
+
+type TestEvent = EventBusEvent<{
```
sam-goodwin, line 5: We should encourage `interface`s instead of `type` aliases. It's good practice to use them since they statically store the type definition in the type system instead of lazily computing the definition. This produces better error messages and other introspection messages and is more efficient - something we need to be careful of.

test/eventpattern.test.ts
```diff
@@ -0,0 +1,797 @@
...
+        {
+          source: ["lambda"],
+        }
+      );
+    });
+
+    test("double equals", () => {
+      ebEventPatternTestCase(
+        reflect<EventPredicateFunction<TestEvent>>(
+          (event) => event.source == "lambda"
```
sam-goodwin, line 35: Can we provide a union type of string literals or an enum to provide more guidance here on the allowed values while also leaving it open for all types of integrations?

test/eventpattern.test.ts
```diff
@@ -0,0 +1,797 @@
...
+        );
+      });
+
+      test("num array not includes", () => {
+        ebEventPatternTestCase(
+          reflect<EventPredicateFunction<TestEvent>>(
+            (event) => !event.detail.numArray.includes(1)
+          ),
+          {
+            detail: { numArray: [{ "anything-but": 1 }] },
```
sam-goodwin, line 204: Wow, pretty cool that you can handle such complex cases

test/eventpattern.test.ts
```diff
@@ -0,0 +1,797 @@
...
+        reflect<EventPredicateFunction<TestEvent>>(
+          (event) =>
+            (event.detail.num > 300 || event.detail.num < 200) &&
+            event.detail.num > 400
+        )
+      );
+    });
+
+    test("numeric range nil range error upper", () => {
+      ebEventPatternTestCaseError(
```
sam-goodwin, line 766: Can we also include the error message in the test case here? I would like to know what error message is reported to the user in this case.

sam-goodwin (changes requested):

test-app/src/app.ts
```diff
@@ -44,3 +45,15 @@ peopleDb.deletePerson.addResolver(api, {
...
+  value: string;
+}>;
+
+new EventBus<MyEvent>(new aws_events.EventBus(stack, "bus"))
+  .when(stack, "aRule", (event) => event.detail.value === "hello")
+  .map<Person>((event) => ({
+    id: event.source,
+    name: event.detail.value,
+  }))
+  .pipe(peopleDb.computeScore);
```
sam-goodwin, line 59: This looks amazing!!

src/compile.ts
thantos: So far the tests seem to show the parenthesis are correctly applied. the contents are already their own expression, curious when the OoO would be incorrect...

src/eventbridge/eventpattern/index.ts
thantos: Seems to work?

```ts
type MyString = string;
interface MyTest extends EventBusRuleInput<{ s: MyString }> {}

test("non-string type", () => {
  ebEventTargetTestCase<MyTest>(
    reflect((event) => event.detail.s + event.detail.s),
    aws_events.RuleTargetInput.fromObject(
      `${EventField.fromPath("$.detail.s")}${EventField.fromPath("$.detail.s")}`
    )
  );
});
```

src/eventbridge/eventpattern/index.ts
thantos: I've always used the `index.ts` to define the top level module. The same as if it as `eventpattern.ts`. I am fine with extracting it out into a new fine with explicit `index.ts` exports.

src/compile.ts
thantos: Because it is reality. Unless reflect fails, it can return all of the other types and the output should try to reflect reality, not intention.

src/compile.ts
thantos: We can, I didn't want to refactor all of the code, but I like this. will cut a PR.

src/compile.ts
thantos: Updated to an assertion and error. This should never happen.

src/eventbridge/eventpattern/index.ts
thantos: I can switch to eval.

src/eventbridge/eventpattern/index.ts
thantos: Eventually I want to refactor these cases to support something like
1. If same type -> transform, error, continue
2. if either one is type -> transform, error, continue

But didn't want to start that yet...

src/eventbridge/eventpattern/index.ts
thantos:
```ts
{
   field1: ["value"],
   field2: ["value"]
}
```

this represents `&&`  between field1 and field2. There is no way to do a `||` without creating multiple rules that are mutually exclusive

```ts
// one
{
   field1: [{ anything-but: "value"}],
   field2: ["value"]
}

// other
{
   field1: ["value"],
   field2: [{ anything-but: "value"}]
}

// both
{
   field1: ["value"],
   field2: ["value"]
}
```

and XOR

```ts
// one
{
   field1: [{ anything-but: "value"}],
   field2: ["value"]
}

// other
{
   field1: ["value"],
   field2: [{ anything-but: "value"}]
}
```

thantos: and multiple rules that point to the same target would generate multiple events, which is why the rules would need to stay perfectly mutually exclusive. 

This, pointing to the same target, will likely send duplicate events when both conditions are met.

```ts
// one
{
   field2: ["value"]
}

// other
{
   field1: ["value"],
}
```

src/eventbridge/eventpattern/index.ts
thantos: It would need to be at the `EventBusRule` level.

src/eventbridge/eventpattern/index.ts
thantos: first element of the array, initialize to undefined if there is no first element.

```ts
const key = Object.keys(classDocument.doc).length > 0 ? Object.keys(classDocument.doc)[0] : undefined.
```

src/eventbridge/eventpattern/numeric.ts
```diff
@@ -0,0 +1,261 @@
...
+  | NumericAggregationPattern
+  | NumericRangePattern
+  | EmptyPattern => {
+  const joinedRanges = aggregation1.ranges.reduce((ranges, range) => {
+    const joinedRanges = aggregation2.ranges
+      .map((r) => joinNumericRange(r, range, true))
+      .filter((r) => validateNumericRange(r));
+    if (joinedRanges.length === 0) {
+      throw new Error(
+        `At least one OR statement in a numeric range must be valid with the other side`
```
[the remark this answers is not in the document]
thantos: In typing out some examples, I think 95% of cases that are desired and logical will work. Many of the edge cases are in themselves illogical and will be caught by the standard compiler. And the logical ones may just be a missing test case. 

That said... I agree, there are a lot of caveats that will fall into three categories:
1. Unsupported natively by the service
2. Too complex to implement on the first pass
3. Missed test cases/combinations/bugs

1. we'll need to document and shame
2. we'll need to document and prioritize
3. we need more real world testing...

src/eventbridge/eventpattern/types.ts
thantos: Ohh, I meant that as of the current version of CDK, there is no support for content matchers in the types. There is no version that supports it now. I can remove the version to reduce confusion.

src/eventbridge/index.ts
thantos: Sure.

test-app/yarn.lock
thantos: The other path didn't install for me, but I think they are the same?

test/eventpattern.test.ts
thantos: https://palantir.github.io/tslint/rules/interface-over-type-literal/

;-)

thantos: And yeah, was going back and forth on how to generate the event structure and I didn't change this one after changing to a generic type.

test/eventpattern.test.ts
thantos: You mean for source?

test/eventpattern.test.ts
thantos: Yeah, I was planning to get back and set them all once things settled down.

test/eventpattern.test.ts
sam-goodwin: yes for `event.source` - is that a closed set of values?

src/eventbridge/eventpattern/index.ts
thantos: Had been calling `Patterns` "Classifications", but have changed them to `Patterns` everywhere now.

test/eventpattern.test.ts
thantos: Actually, looks like interfaces cannot extend types with type arguments, so type it is.

test/eventpattern.test.ts
thantos: No, it can be any string value, and some AWS services send their ID. I'd say we can add a list later, but we'll always allow for a set of strings.

test/util.ts
thantos: Functionless, need a better acronym and give up and use the full word...

test-app/yarn.lock
thantos: Renamed.

test/util.ts
thantos: Renamed

src/compile.ts
thantos: #52

src/eventbridge/eventpattern/index.ts
sam-goodwin: Love that.

src/eventbridge/bus.ts
```diff
@@ -0,0 +1,114 @@
+import { aws_events } from "aws-cdk-lib";
+import { Construct } from "constructs";
+import { EventBusRule, EventPredicateFunction } from "./rule";
+import { EventBusRuleInput } from "./types";
+
+/**
```
sam-goodwin, line 6: Fantastic docs!

sam-goodwin:
Curious if we can clean up the folder structure?
image

Do we need the nested folder, `eventpattern`? I think it adds unnecessary nesting - each file name is expressive enough on its own.

Can we rename `eventbridge` to `event-bridge` - I think it's easier to read.

src/eventbridge/bus.ts
```diff
@@ -0,0 +1,114 @@
...
+ * A Functionless wrapper for a AWS CDK {@link aws_events.EventBus}.
+ *
+ * Filtering events and sending them to Lambda.
```
sam-goodwin, line 9: Can we elaborate here? 

> Wrap your {@link aws_events.EventBus} instance with this class, specify a type to represent the events passing through the EventBus, and then use the `.when`, `.map` and `.pipe` functions to express EventBus Rules and Targets with native TypeScript syntax.

src/eventbridge/bus.ts
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * ```ts
+ * interface Payload {
+ *    value: string;
+ * }
+ *
+ * // An event with the payload
+ * type myEvent = EventBusRuleInput<Payload>;
```
sam-goodwin, line 17: probably better to use an `interface` here

https://blog.logrocket.com/types-vs-interfaces-in-typescript/

src/eventbridge/bus.ts
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * const myAwsFunction = new aws_lambda.Function(this, 'myFunction', { ... });
+ * // A function that expects the payload.
+ * const myLambdaFunction = new Function<Payload, void>();
```
sam-goodwin, line 21: this should wrap `myAwsFunction`

Also, should you use `functionless.Function` so it's clear that it is imported?

src/eventbridge/bus.ts
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * // Some AWS Event Bus, may be new, may be default or imported.
+ * const awsBus = new aws_events.EventBus(this, "mybus");
```
sam-goodwin, line 24: Could be simpler like:

```ts
// instantiate an aws_events.EventBus Construct
const awsBus = new aws_events.EventBus(this, "mybus");
// you can also import it
const awsBus = aws_events.EventBus.fromArn(arn)
```

src/eventbridge/bus.ts
```diff
@@ -0,0 +1,114 @@
...
+export class EventBus<E extends EventBusRuleInput> {
+  /**
+   * This static property identifies this class as an EventBus to the TypeScript plugin.
+   */
+  public static readonly FunctionlessType = "EventBus";
+
+  constructor(readonly bus: aws_events.EventBus) {}
+
+  /**
+   * Filter events using as a EventBus Rule using a Functionless predicte function.
```
sam-goodwin, line 55: Bit confusing to read:
> .. using as a EventBus Rule using a ..

src/eventbridge/bus.ts
```diff
@@ -0,0 +1,114 @@
...
+   *
+   * Equals
+   *
+   * ```ts
+   * when(this, 'rule', (event) => event.source === "lambda")
```
sam-goodwin, line 60: Explain the `this` and `'rule'`.

Also, use `"` instead of `'` to align with our ESLint configuration.

src/eventbridge/bus.ts
```diff
@@ -0,0 +1,114 @@
...
+   * when(this, 'rule', (event) => event.source === "lambda" && event['detail-type'] === "SUCCESS")
+   * ```
+   *
+   * Array Includes
+   *
+   * ```ts
+   * when(this, 'rule', (event) => event.detail.list.includes("someValue"))
+   * ```
+   *
+   * Unsupported by Event Bridge
```
sam-goodwin, line 99: How exhaustive are the above examples of supported syntax?

src/eventbridge/rule.ts
```diff
@@ -0,0 +1,164 @@
...
+  EventBusTargetProps,
+  EventBusTargetResource,
+  LambdaTargetProps,
+  pipe,
+} from "./target";
+
+/**
+ * A function interface used by the {@link EventBus}'s when function to generate a rule.
+ *
+ * event is every event sent to the bus to be filtered. This argument is optional.
```
sam-goodwin, line 19: Are you missing `@param event`?

src/eventbridge/rule.ts
```diff
@@ -0,0 +1,164 @@
...
+ */
+export type EventPredicateFunction<
+  E extends EventBusRuleInput = EventBusRuleInput<any>
+> = (event: E) => boolean;
+
+/**
+ * Represents a set of events filtered by the when predicate using event bus's EventPatterns.
+ * https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-event-patterns.html
+ *
+ * See {@link EventBus}.when for more details on filtering events.
```
sam-goodwin, line 29: This should be `{@link EventBus.when}` - notice how if you change it to this, you can use jump-to-definition to go straight to that method.

sam-goodwin: Should we provide some extra documentation on this class? 

> To create EventBusRules, use the {@link EventBus.when} helper instead of instantiating this class directly.
> ```
> (single example with .when, .map, .transform)
> ```

src/eventbridge/rule.ts
```diff
@@ -0,0 +1,164 @@
...
+   *
+   * Local varaibles
+   *
+   * ```ts
+   * .map(event => {
+   *    const formatted = `${event.id}_${event.detail.id}`;
+   *    return ({
+   *       formatted
+   *    })
+   * })
```
sam-goodwin, line 108: Impressive!

src/eventbridge/rule.ts
```diff
@@ -0,0 +1,164 @@
...
+   * ```
+   *
+   * Unsupported by EventBridge:
+   * * String slicing/sub string
+   * * Arithmatic
```
sam-goodwin, line 113: Typo
> Arithmetic

src/eventbridge/rule.ts
```diff
@@ -0,0 +1,164 @@
...
+   * * Boolean logic `event.source === "lambda"`)
+   * * Formatting the whole event `event: ${event}`
+   * * More...
```
sam-goodwin, line 116: Can we document the caveats exhaustively?

src/eventbridge/rule.ts
```diff
@@ -0,0 +1,164 @@
...
+   *
+   * Unsupported by Functionless:
+   * * Variables from outside of the function scope
```
sam-goodwin, line 119: link to issue?

src/eventbridge/rule.ts
```diff
@@ -0,0 +1,164 @@
...
+
+  /**
+   * Defines a target of the {@link EventBusTransform}'s rule using this TargetInput.
+   *
+   * The event is sent to the target verbatim unless .map is used first.
+   *
+   * Event buses do not support TargetInput transforms.
+   * https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-bus-to-bus.html
+   *
+   * Send to lambda
```
sam-goodwin, line 133: Can we make the headings stand out better?

> Ex. Send To Lambda:

src/eventbridge/rule.ts
```diff
@@ -0,0 +1,164 @@
...
+  map<P>(transform: EventTransformFunction<T, P>): EventBusTransform<T, P> {
+    return new EventBusTransform<T, P>(transform, this);
+  }
+
+  /**
+   * Defines a target of the {@link EventBusTransform}'s rule using this TargetInput.
+   *
+   * The event is sent to the target verbatim unless .map is used first.
+   *
+   * Event buses do not support TargetInput transforms.
```
sam-goodwin, line 130: Reword suggestion:
> EventBus does not support TargetInput transforms

Also, it's not obvious to me what `TargetInput` is.

src/eventbridge/rule.ts
```diff
@@ -0,0 +1,164 @@
...
+   * https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-bus-to-bus.html
```
sam-goodwin, line 131: use `@see` 

> @see https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-bus-to-bus.html

src/eventbridge/target.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export type EventBusTargetResource<T extends EventBusRuleInput, P> =
+  | Function<P, any>
+  | LambdaTargetProps<P>
+  | EventBus<T>
+  | EventBusTargetProps<T>;
+
+  /**
+   * Add a target to the run based on the configuration given.
+   */
```
sam-goodwin, line 22: formatting - be sure to run prettier

src/eventbridge/target.ts
```diff
@@ -0,0 +1,53 @@
...
+export const pipe = <T extends EventBusRuleInput, P>(
```
sam-goodwin, line 23: any reason why this isn't in the `rule.ts`? Is it used anywhere else?

src/eventbridge/transform.ts
```diff
@@ -0,0 +1,61 @@
...
+    ruleArn: string;
+    ruleName: string;
+    ingestionTime: string;
+    eventJson: string;
+  };
+}
+
+/**
+ * Represents an event that has been transformed using Target Input Transformers.
+ * https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-transform-target-input.html
```
sam-goodwin, line 34: use `@see`

src/eventbridge/transform.ts
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * See {@link EventBusRule}.map for more details on transforming event details.
```
sam-goodwin, line 36: fix:
``` 
{@link EventBusRule.map}
```

src/eventbridge/transform.ts
```diff
@@ -0,0 +1,61 @@
...
+    const decl = func as unknown as FunctionDecl;
+    this.targetInput = synthesizeEventBridgeTargets(decl);
+  }
+
+  /**
+   * Defines a target of the {@link EventBusTransform}'s rule using this TargetInput.
+   *
+   * EventBus is not a valid pipe target for transformed events.
+   *
+   * See {@link EventBusRule}.pipe for more details on pipe.
```
sam-goodwin, line 54:
```
{@link EventBusRule.pipe}
```

sam-goodwin (changes requested):

thantos:
I think we should try to keep each of the different target formats somewhat separate.

In event bridge, the pattern and target input formats are mostly unrelated to each other.

Pattern

```ts
{
    source: ["lambda"]
}
```

Target Input

```ts
{
    inputPathMap: {
         value: '$.source'
    },
    inputTemplate: '<value>'
}
```

-----

Fine with renaming the folders.

Should we update the other folders too? We have `appsync` for example.

src/eventbridge/transform.ts
thantos: :+1: wasn't sure if that would work!

src/eventbridge/transform.ts
thantos: Nice, will keep this handy for more rich docs.

https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#see

src/eventbridge/bus.ts
thantos: Woops...
> EventBus Rules can filter events using Functionless predicate functions.

src/eventbridge/target.ts
thantos: ha... it runs on explicit save, I guess I didn't hit save on that file. Good catch.

src/eventbridge/target.ts
thantos: Yeah, `pipe` is used in `rule.ts` and `transform.ts` because we can `pipe` directly from a rule (`.when().pipe()`, no transform) or after a transform (`.when().map().pipe()`, with transform). `pipe` generalizes the logic between the two, including failing if there is an event target input (transform), but the target is event bus which doesn't support transforming.

thantos:
This leaves the format as...

```
event-bridge - The public classes for wrapping EventBus, `when`, `map`, `pipe`
- event-pattern - everything needed to compile/synth a FunctionDecl into an Event Pattern
- target-input - everything needed to compile/synth a FunctionDecl into an Event Target Input 
```

src/compile.ts
```diff
@@ -55,38 +58,103 @@ export function compile(
...
-              ts.factory.createIdentifier(error.name),
-              undefined,
-              [ts.factory.createStringLiteral(error.message)]
-            ),
-          ]) as unknown as Statement;
-        }
-        return ts.visitEachChild(node, visitor, ctx);
+        };
+        // keep processing the children of the updated node.
+        return ts.visitEachChild(_visit(), visitor, ctx);
```
sam-goodwin, line 86: Why visit the transformed code? This would be expensive and unnecessary.

sam-goodwin (changes requested):

src/event-bridge/bus.ts
```diff
@@ -0,0 +1,122 @@
...
+ * const anotherEventBus = aws_event.EventBus.fromEventBusArn(...);
+ *
+ * new functionless.EventBus<myEvent>(awsBus)
+ *    // when the payload is equal to some value
+ *    .when(this, 'rule2', event => event.detail.value === "some value")
+ *    // send verbatim to the other event bus
+ *    .pipe(anotherEventBus);
+ * ```
+ */
+export class EventBus<E extends EventBusRuleInput> {
```
sam-goodwin, line 54: Should the file be named `event-bus.ts`?

sam-goodwin (changes requested):

src/compile.ts
```diff
@@ -55,38 +58,103 @@ export function compile(
...
+
+      /**
+       * Checks to see if a node is of type EventBusRule.
+       * The node could be any kind of node that returns an event bus rule.
+       */
+      function isEventBusRule(node: ts.Node) {
+        return isFunctionlessClassOfKind(node, EventBusRule.FunctionlessType);
+      }
+
+      function isFunctionlessClassOfKind(node: ts.Node, kind: string) {
```
sam-goodwin, line 146: Running this multiple times per node is wasteful.

sam-goodwin (changes requested):

src/compile.ts
thantos: Ehh? what part of it? Seems like early optimization honestly. We could assign some form of name to a node type, but there could be multiple layers of validation that differ between node types, including which level needs to be checked.

For example, `isEventBusMapFunction` checks it for `node.expression.expression` only when `node.expression` is named map. 

`isNewEventBusRule` check it for `node.expression` only when the `node` is a `NewExpression`.

Probably orders of magnitude less expensive than the tree exploration done in the step functions code or the re-computation of lexical scope.

src/compile.ts
thantos: because the child of the transformed class is not transformed and may need to be.

```ts
bus
   .when(event => event)
   .map(event => event.detail)
```

The current visit is on the `map` method. The contents of map are updated and returned, but the fact that it was chained off a `when` was not considered during `visitEventBusMapFunction`. 

We could add the exploration down the children to each of the visit methods, but that seems unnecessary.

This is something you wouldn't hit when always transforming the top of the statement (`new AppResolver()`), but once you need to transform a method on a class its needed.

thantos:
> > Should we update the other folders too? We have appsync for example.
> 
> If I had my way, we would have no folders lol.
> 
> If we decide to, let's do one at a time in separate PRs

hmm, not sure what you are looking for here, did you want all of the event bus transforms and classes merged back into a single file or did you want the structure to be flattened?

```
... everything else in src...
event-bus.ts
event-bus-utils.ts
event-bus-rule.ts
event-bus-transform.ts
event-bus-target.ts
event-bus-types.ts
event-bus-pattern-synth.ts
event-bus-pattern-numeric.ts
event-bus-pattern-pattern.ts
event-bus-pattern-types.ts
event-bus-pattern-utils.ts
event-bus-target-synth.ts
```

README.md
```diff
@@ -219,7 +224,7 @@ new AppsyncResolver(() => {
 });
 ```
 
-### `AppsyncResolver` construct for defining Appsync Resolver with TypeScript syntax
+#### `AppsyncResolver` construct for defining Appsync Resolver with TypeScript syntax
```
sam-goodwin, line 227: Why did this change?

README.md
```diff
@@ -308,6 +313,238 @@ getPerson.addResolver(api, {
...
+  name: string;
+  age: number;
+  interests: string[];
+}
+
+interface UserEvent
+  extends functionless.EventBusRuleInput<
+    UserDetails,
+    // We can provide custom detail-types to match on
+    "Create" | "Update" | "Delete"
```
sam-goodwin, line 351: Love it

sam-goodwin: Is this optional?

README.md
```diff
@@ -308,6 +313,238 @@ getPerson.addResolver(api, {
...
+  > {}
+```
+
+And wrap our CDK `EventBus`
```
sam-goodwin, line 355: Explain why? Also, I think we should add an ability to instantiate our constructs without wrapping - we can use constructor overloads to allow both?

README.md
```diff
@@ -308,6 +313,238 @@ getPerson.addResolver(api, {
...
+
+```ts
+const bus = functionless.EventBus<UserEvent>(awsEventBus);
+```
+
+#### Create Rules (`EventBusRule`) on a Event Bus to match incoming events.
+
+Now that you have a wrapped `EventBus`, lets add some rules.
```
sam-goodwin, line 363: The wrapping concept is quickly becoming cumbersome, do you agree? Maybe we should follow the cdk's import pattern?

README.md
```diff
@@ -308,6 +313,238 @@ getPerson.addResolver(api, {
...
+
+Functionless lets you write logic in Typescript on the type safe event.
+
+Lets match all of the create or updated events with one rule and another rule for deletes.
```
sam-goodwin, line 367: `"Create"`

README.md
```diff
@@ -308,6 +313,238 @@ getPerson.addResolver(api, {
...
+    event["detail-type"] === "Create" || event["detail-type"] === "Update"
+);
+const deleteEvents = bus.when(
+  this,
+  "deleteRule",
+  (event) => event["detail-type"] === "Delete"
+);
+```
+
+We also want to do something special when we get a new cat lover who is between 18 and 30 years old, lets make another rule for those.
```
sam-goodwin, line 383: Lol cat lover

README.md
```diff
@@ -308,6 +313,238 @@ getPerson.addResolver(api, {
...
+
+```ts
+const youngAdultCatLoversEvents = bus.when(
```
sam-goodwin, line 386: This name raises my eyebrows lol

sam-goodwin (changes requested):

README.md
```diff
@@ -308,6 +313,238 @@ getPerson.addResolver(api, {
...
+
+const createOrUpdate = functionless.Function<CreateOrUpdate, void>(createOrUpdateFunction);
+const delete = functionless.Function<Delete, void>(createOrUpdateFunction);
+```
+
+The events from before do not match the formats from before, so lets transform them to the structures match.
+
+```ts
+const createOrUpdateEventsTransformed =
+  createOrUpdateEvents.map<CreateOrUpdate>((event) => ({
```
sam-goodwin, line 427: I love the `.map<CreateOrUpdate>` - I usually try to avoid explicit type arguments but this solves the problem optimally.

README.md
```diff
@@ -308,6 +313,238 @@ getPerson.addResolver(api, {
...
+
+#### Target other AWS services like Lambda and other Event Buses
+
+Now that we have created rules on our event buses using `when` and transformed those matched events using `map`, we need to send the events somewhere.
+
+We can `pipe` the transformed events to the lambda functions we defined earlier.
+
+```ts
+createOrUpdateEventsTransformed.pipe(createOrUpdate);
+deleteEventsTransformed.pipe(delete);
```
sam-goodwin, line 448: Avoid the `delete` keyword.

README.md
```diff
@@ -646,6 +883,320 @@ $util.qr($v1.put($b, true))
 #end
 ```
 
+## Typescript -> Event patterns
```
sam-goodwin, line 886: Let put this in a collapsible section to save space? We need to later move these to dedicated markdown files for docusaurus. Readme is huge.

sam-goodwin (commented):
Amazing documentation man. Thanks so much for the effort.

README.md
thantos: I think I bumped all of those headers down a level.
image

It was

```
## Usage
### `AppsyncResolver` construct for defining Appsync Resolver with TypeScript syntax
```

and I made it 

```
## Usage
### App Sync
#### `AppsyncResolver` construct for defining Appsync Resolver with TypeScript syntax
### Event Bridge
#### EB Stuff
```

README.md
thantos: Yeah, detail-type is optional, and source

```ts
interface EventBusRuleInput<Detail, DetailType extends string = string, Source extends string = string> {} 
```

README.md
thantos: Let me try

README.md
thantos:
```ts
functionless.EventBus.fromBus(aws_events.EventBus.fromEventBusArn(...));
new functionless.EventBus(this, 'myBus');
```

README.md
sam-goodwin: We can do the constructor change in a separate pr

thantos: Already did it! But had to solve a problem where the functionless static members don't propagate to declarations automatically. 

Will push tonight

src/compile.ts
```diff
@@ -162,16 +242,63 @@ export function compile(
         }
       }
 
-      function isFunctionlessClassOfKind(node: ts.Node, kind: string) {
+      /**
+       * Checks if a type, class, or interface extends or implements a type, class, or interface name.
+       */
+      function extendsType(type: ts.Type | undefined, exts: string): boolean {
```
sam-goodwin, line 248: Can we simplify this by just adding a string literal property that we look for?

```ts
interface EventBusRule {
  Kind: "EventBusRule";
}

type.getProperty("Kind").type.isStringLiteral()
```

sam-goodwin: This looks expensive and I'm not sure why it's necessary. The structural type system and a basic property-heuristic would be faster and simpler. I've also experimented with using a `Symbol` if you're worried about clashes.

src/event-bridge/event-bus.ts
```diff
@@ -0,0 +1,161 @@
...
+ *    // send verbatim to the other event bus
+ *    .pipe(anotherEventBus);
+ * ```
+ */
+export class EventBus<E extends EventBusRuleInput> extends EventBusBase<E> {
+  constructor(scope: Construct, id: string, props?: aws_events.EventBusProps) {
+    super(new aws_events.EventBus(scope, id, props));
+  }
+
+  static fromBus<E extends EventBusRuleInput>(
```
sam-goodwin, line 150: Docs.

src/compile.ts
thantos: > Can we simplify this by just adding a string literal property that we look for?

I don't like this, it is mixing the build time and runtime behavior of the interface. We'd need to check the type of the field in the interface and anyone who creates the interface will need to add that property.

We'll still need to explore the hierarchy of types to find the interface anyways because we cannot check the runtime value.

When I started to extend/implement classes and interfaces, the properties were no longer present and had to explore the `heritageClauses`.

sam-goodwin: > We'd need to check the type of the field in the interface and anyone who creates the interface will need to add that property.

These are internal classes and interfaces, so i don't think this is a big deal.

> When I started to extend/implement classes and interfaces, the properties were no longer present and had to explore the heritageClauses.

How were you getting the property? Were you using `ts.Type` or the raw AST, `ts.InterfaceDeclaration`?

thantos: https://ts-ast-viewer.com/#code/JYOwLgpgTgZghgYwgAgJLmvJyDeAoZQ5Aa1ABMAuZAInUlkQmrwF889R6sUBhAC2AAbMsggAPSCDIBnNBgbZ8RZADc4ggK4Qq1aQHsAthDACQAc2Zs8CQXGmz+Qsj1v3kwAwAdBEI+AcCwrjsLEA

There are no members on the class

image

and even the `Child` interface is missing the `kind` member

image

The `__FunctionlessBase` has no runtime impact at all, but its consistently discoverable.

sam-goodwin: Use checker.getType* and use ts.Type, instead of the ast.

sam-goodwin: Add a type discriminator property just like we do for union types

thantos: Not worried about clashing, didn't want instances to need to add the property. Will try to get type instead of the AST (but I am using `getTypeAtLocation` and the type is what is missing the members).

test-app/cdk.context.json
```diff
@@ -0,0 +1,5 @@
+{
+  "acknowledged-issue-numbers": [
+    19179
```
sam-goodwin, line 3: Hmm what's this?

test-app/src/app.ts
```diff
@@ -53,3 +55,17 @@ peopleDb.deletePerson.addResolver(api, {
   typeName: "Mutation",
   fieldName: "deletePerson",
 });
+
+type MyEvent = EventBusRuleInput<{
```
sam-goodwin, line 59: interface instead of type? Which way are we going to encourage?

test-app/src/people-events.ts
```diff
@@ -0,0 +1,124 @@
...
+          event["detail-type"] === "Create" &&
+          event.detail.interests.includes("CATS") &&
+          event.detail.age >= 18 &&
+          event.detail.age < 30
+      )
+      .pipe(catPeopleBus);
+
+    // a function to view the events which make it to the cat bus
+    catPeopleBus
+      .when(this, "catSinkRule", () => true)
```
sam-goodwin, line 109: Hmm, do we need some other helpers, such as .forAll?

test/eventbus.test.ts
```diff
@@ -0,0 +1,114 @@
...
+  const rule = busBus.when(stack, "rule", () => true);
+  rule.pipe(busBus);
+
+  expect((rule.rule as any).targets.length).toEqual(1);
+  expect(
+    (rule.rule as any).targets[0] as aws_events.IRuleTarget
+  ).toHaveProperty("arn");
+});
+
+test("new bus with when map pipe function aa", () => {
```
sam-goodwin, line 70: `aa` ?

test/eventpattern.test.ts
```diff
@@ -0,0 +1,1517 @@
+import { ebEventPatternTestCase, ebEventPatternTestCaseError } from "./util";
+import { EventBusRuleInput, EventPredicateFunction } from "../src/event-bridge";
+import { reflect } from "../src";
+
+type TestEvent = EventBusRuleInput<{
```
sam-goodwin, line 5: I'm wondering why we need the developer to always use this wrapper? Can't they instead just declare their type and then our constructs wrap it?

sam-goodwin (changes requested):

src/event-bridge/rule.ts
```diff
@@ -0,0 +1,207 @@
...
+   * * String slicing/sub string
+   * * Arithmatic
+   * * Boolean logic `event.source === "lambda"`)
+   * * Formatting the whole event `event: ${event}`
+   * * More...
+   *
+   * Unsupported by Functionless:
+   * * Variables from outside of the function scope
+   */
+  map<P>(transform: EventTransformFunction<T, P>): EventBusTransform<T, P>;
```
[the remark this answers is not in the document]
thantos: Payload

src/event-bridge/target-input.ts
```diff
@@ -0,0 +1,300 @@
...
+ * Generates a RuleTargetInput from a typescript function.
+ *
+ * Transforms an input event `event` to the return type `P`.
+ *
+ * TargetInputs interact with the input event using JSON Path.
+ *
+ * https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-transform-target-input.html
+ */
+export const synthesizeEventBridgeTargets = (
+  decl: FunctionDecl | Err | any
```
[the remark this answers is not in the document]
thantos: Updated to `unknown`

I wanted to force checking because of how easy it is to not get a valid input (ex: `ts-patch` is missing) and throw a logical error.

test-app/cdk.context.json
thantos: CDK was complaining, this acknowledges the "risk" so the CLI stops complaining.

src/event-bridge/utils.ts
```diff
@@ -0,0 +1,359 @@
...
+  StringLiteralExpr,
+  TemplateExpr,
+  UnaryExpr,
+} from "../expression";
+import { isReturn, isVariableStmt, Stmt, VariableStmt } from "../statement";
+
+/**
+ * Returns a string array representing the property access starting from a named identity.
+ *
+ * Does not return the itentity name given.
```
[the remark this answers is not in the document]
thantos: so many spelling issues...

src/step-function.ts
```diff
@@ -326,6 +326,7 @@ abstract class BaseStepFunction<F extends AnyFunction>
   implements aws_stepfunctions.IStateMachine
 {
   readonly kind = "StepFunction";
+  readonly functionlessKind = "StepFunction";
```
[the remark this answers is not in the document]
thantos: This is partially laziness. I had a constant, but it breaks the compiler code that expects a string literal. Seemed like small tradeoff to get this shipped.

test-app/src/app.ts
thantos: interfaces are not supported in this form. 

> An interface can only extend an identifier/qualified-name with optional type arguments. ts(2499)

The options are

```ts
interface MyEventDetails {
  value: string;
}

interface MyEvent extends EventBusRuleInput<MyEventDetails>{};
```

or 

```ts
type MyEvent = EventBusRuleInput<{
   value: string;
}>;
```

I went with the less verbose way, but fine with the former.

test-app/src/people-events.ts
thantos: We could just add `pipe` to the bus with a default rule. I'll create a ticket.

thantos: https://github.com/sam-goodwin/functionless/issues/89

test/eventbus.test.ts
thantos: ha... my lazy way to break the compile cache was to add letters to the test names. I missed one or two I guess.

src/step-function.ts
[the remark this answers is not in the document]
sam-goodwin: Can't you get the type of the property? The type should be a StringLitersl regardless of where it came from. Avoid the ast if you can

test/eventpattern.test.ts
thantos: My initial thought was allowing the developer to provide rich types for any of the fields in the wrapper.

```
type MyEvent = EventBusRuleInput<{
   value: string
}> & {
    version: 1 | 2 | 3
}
```

and to support rich unions

```
type MyEvent = EventBusRuleInput<{ someField: string }, 'customDetailType'>;

new EventBus<StepFunctionEvent | MyEvent>( )
   .when(event => event['detail-type'] === customDetailType && event.detail.someField)
```

Thoughts on how to handle this more gracefully?

sam-goodwin: Let's open an issue and continue the discussion.

sam-goodwin (approved):
Did you update the front page of the readme to reflect the StepFunction change for arguments? Also, is this a breaking change?

thantos:
This PR does not include the Step Functions change, PR will be up when this one ships.

test/eventpattern.test.ts
thantos: #94

PR #40, opened 2022-03-19
fix: improve VTL synth to support if-statements after a service call and inline dereferencing of service call

sam-goodwin:
There were some bugs in how we synthed VTL:

1. if-statements occurring after the last service call were ignored.
```ts
const result = serviceCall()
// below did not result in VTL
if (result.cond) {
  return result.a;
} else {
  return result.b;
}
```
2. in-lined PropAccessExpr on a service call was ignored and broke the request mapping template
```ts
return serviceCall().prop
```

The request mapping template would mistakenly place the `.prop` reference on the request:
```
$util.toJson($v1.prop)
```

Now, the response mapping template contains that reference:
```
#set($context.stash.return__val = $context.result.prop)
```

3. the NoneDataSource required that the request mapping template was an object with `version` and `payload` properties.
```
{
  "version": "2018-05-29",
  "payload": null
}
```

### DIFF
--- a/src/appsync.ts
+++ b/src/appsync.ts
@@ -1,6 +1,6 @@
 import * as appsync from "@aws-cdk/aws-appsync-alpha";
 import { AppSyncResolverEvent } from "aws-lambda";
-import { CallExpr } from "./expression";
+import { CallExpr, Expr } from "./expression";
 import { AnyLambda } from "./function";
 import { VTL } from "./vtl";
 import { AnyTable, isTable } from "./table";
@@ -49,6 +49,22 @@ export type ResolverFunction<
   Result,
   Source
 > = ($context: AppsyncContext<Arguments, Source>) => Result;
+
+export interface MaterializedResolverProps extends appsync.ResolverProps {
+  templates: string[];
+}
+
+export interface MaterializedResolver extends appsync.Resolver {
+  readonly templates: string[];
+}
+
+export function materialize(
+  templates: string[],
+  resolver: appsync.Resolver
+): MaterializedResolver {
+  (resolver as any).templates = templates;
+  return resolver as MaterializedResolver;
+}
 
 /**
  * An AWS AppSync Resolver Function derived from TypeScript syntax.
@@ -125,11 +141,13 @@ export class AppsyncResolver<
       appsync.BaseResolverProps,
       "typeName" | "fieldName" | "cachingConfig"
     >
-  ): appsync.Resolver {
+  ): MaterializedResolver {
     let stageIt = 0;
 
     const resolverCount = countResolvers(this.decl);
+    const templates: string[] = [];
 
+    let resolver: MaterializedResolver;
     if (resolverCount === 0) {
       // mock integration
       const [pipelineConfig, responseMappingTemplate] = synthesizeFunctions(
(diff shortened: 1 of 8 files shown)


sam-goodwin:
Also improved the coverage of our test cases to actually create the appsync resolver rather than just testing the VTL class. This caught some other bugs which i fixed.

sam-goodwin:
Deployed a stack and tested everything still works.

src/appsync.ts
```diff
@@ -50,6 +50,22 @@ export type ResolverFunction<
...
 
+export interface MaterializedResolverProps extends appsync.ResolverProps {
+  templates: string[];
+}
+
+export interface MaterializedResolver extends appsync.Resolver {
+  readonly templates: string[];
+}
+
+export function materialize(
```
thantos, line 57: Docs? I am not sure I understand all of the templates in the return array

src/appsync.ts
```diff
@@ -210,63 +250,89 @@ export class AppsyncResolver<
...
+                `#set( $context.stash.${stmt.name} = ${getResult(
+                  stmt.expr
+                )} )\n{}`
+              );
             } else {
               throw new Error(
                 `only a 'VariableDecl', 'Call' or 'Return' expression may call a service`
               );
             }
+            function getCall(expr: Expr): CallExpr {
```
thantos, line 270: doc? Searched up the tree for a call, or ref, or something. Also this function seems really specific to the current integrations of lambda and dynamo tables.

thantos (changes requested):
I don't see a test for if-statements, is that change in here?

sam-goodwin:
> I don't see a test for if-statements, is that change in here?

Lol good point! I went to add the test and realized I had to refactor tests before I could even test that case. Forgot to add it once I was done.

src/appsync.ts
```diff
@@ -142,52 +160,74 @@ export class AppsyncResolver<
...
+            requestMappingTemplate
+          ),
+          responseMappingTemplate: appsync.MappingTemplate.fromString(
+            responseMappingTemplate
+          ),
+        })
+      );
+      templates.push(responseMappingTemplate);
+
+      return resolver;
```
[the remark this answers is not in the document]
sam-goodwin: Agree. This needs to be cleaned up. First priority is to fix critical bugs and enable unit test coverage of the entire pipeline.

src/appsync.ts
sam-goodwin: Will add docs and explain the cases it is solving. I want to clean this all up but my first priority is to fix bugs and merge step functions before diverging too far.

test/util.ts
```diff
@@ -8,9 +11,30 @@ export function returnExpr(varName: string) {
 #return($context.stash.return__val)`;
 }
 
-export function testCase(decl: FunctionDecl, expected: string) {
-  const vtl = new VTL();
-  vtl.eval(decl.body);
-  const actual = vtl.toVTL();
+export function testCase(decl: FunctionDecl, ...expected: string[]) {
```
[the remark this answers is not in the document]
sam-goodwin: Renamed to `appsyncTestCase`

thantos (approved):

PR #43, opened 2022-03-20
fix: update lambda function to accept json payload

### LINKED ISSUE #42: Lambda Json Payload
Currently, functionless turns a Lambda payload into a array of arguments in the callable `Function`.

```ts
new Function<(arg: string, arg2: string) => string>(...)
```

which actually represent a payload of the form

```ts
{
   arg: string;
   arg2: string;
}
```

This is problematic because 
1. Lambda supports any valid json, including strings and numbers as the payload
2. All integrations with lambda need to use an array of args instead of the more natural object

### Proposal:

Accept any type payload and output to the function as arbitrary json.

```
new Function<{ arg: string, arg2: string }, string>(...); 
```

thantos:
Fixes #42

Previously, Lambda Functions were defined with a type parameter that represented the function signature. Arguments were turned into objects using their names.

```ts
const myFunction = new Function<(arg1: string, arg2: number) => string>(...);

myFunction("some string", 10);

// actual json into the lambda
{
   arg1: "some string:",
   arg2: 10
}
```

This PR updates the function to accept a single Payload argument, matching the public lambda contract that accepts any single JSON input as a parameter.

```ts
const myFunction = new Function<{ arg1: string, arg2: number }, string>(...)

myFunction({ arg1: "some string", arg2: 10 });
```

The payload passed to the lambda function is the same as the above example.

The new representation allows for bare strings, numbers, and nulls.

```ts
const myFunction = new Function<string, string>(...)

myFunction(some string"");
```

This example is passed to the lambda functions as expected, as a bare string.

----

BREAKING CHANGE: Function signature changed from argument list to one or zero arguments for payload as json. Function type parameter is now a payload type and a return type. Function return type defaults to void.

### DIFF
--- a/src/function.ts
+++ b/src/function.ts
@@ -5,11 +5,11 @@ import { VTL } from "./vtl";
 // @ts-ignore - imported for typedoc
 import type { AppsyncResolver } from "./appsync";
 
-export function isFunction(a: any): a is Function {
+export function isFunction(a: any): a is Function<any, any> {
   return a?.kind === "Function";
 }
 
-export type AnyLambda = Function;
+export type AnyLambda = Function<any, any>;
 
 /**
  * Wraps an {@link aws_lambda.Function} with a type-safe interface that can be
@@ -17,7 +17,7 @@ export type AnyLambda = Function;
  *
  * For example:
  * ```ts
- * const getPerson = new Function<(key: string) => Person>(
+ * const getPerson = new Function<string, Person>(
  *   new aws_lambda.Function(..)
  * );
  *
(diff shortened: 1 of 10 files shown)


src/function.ts
```diff
@@ -42,34 +40,32 @@ export type AnyLambda = Function<AnyFunction>;
  *
  * Make sure to implement your Lambda Function entry-points accordingly.
  */
-export class Function<F extends AnyFunction> {
+export class Function<P = undefined, O = void> {
```
sam-goodwin, line 43: Do we need default types? I think we should encourage the user to specify them. Also the defaults represent a pretty uselsss function haha

src/function.ts
```diff
@@ -42,34 +40,32 @@ export type AnyLambda = Function<AnyFunction>;
...
+  constructor(readonly resource: aws_lambda.IFunction) {
     return Object.assign(lambda, this);
 
     function lambda(call: CallExpr, vtl: VTL): string {
-      const payload = vtl.var(`{}`);
-      for (const [argName, argVal] of Object.entries(call.args)) {
-        vtl.qr(`${payload}.put('${argName}', ${vtl.eval(argVal)})`);
-      }
+      // first argument is the payload
+      const [payloadKey] = Object.keys(call.args);
```
sam-goodwin, line 51: Could it be undefined?

sam-goodwin (changes requested):

src/function.ts
thantos: no, because its now the definition of the Function, so I was always getting `args`.

src/function.ts
thantos: is it useless?

`start()`
`record()`
`increment()`
`submit()`

But sure, we could get rid of the defaults.

sam-goodwin: I guess not useless then. Useful for timers and alike. But probably not common?

src/function.ts
sam-goodwin: What about when P is undefined? Didn't you change the signature to have no args in that case?

thantos: it still returns `args` as a property name (the tests seems to work).

thantos: Upon more testing, the args map is always returning `args` mapped to either `nullliteral` or a prop accessor, base on 0 or one arguments. Given two arguments it still returns only one.

Putting function brand back and set to the computed `ConditionalFuntion` has returned a more logical result. Will refacto r :+1:

src/function.ts
thantos: Removed the defaults, always can add it back later.

src/function.ts
sam-goodwin: Yes function brand is required to give the type of the resolved function to the compiler.

sam-goodwin (dismissed):
When you merge, please make sure the `BREAKING CHANGE:` section is included as the commit message.

sam-goodwin:
Also, can you please update the description to include some examples of how things changed and what customers should do to adapt? Think of the PR description as the content developers will find when their code breaks - it should help them fix it.

sam-goodwin:
Looks like we forgot to update the `README.md`

sam-goodwin (changes requested):
Please make sure to update the README.md with the new syntax.

sam-goodwin (approved):

sam-goodwin (approved):

PR #46, opened 2022-03-28
feat: compile returns error node instead of throwing

### LINKED ISSUE #11: Transformer throwing Errors kills the TypeScript language server
Currently, when the TypeScript transformer encounters a node it does not understand, we just throw an Error and display the syntax (text) that could not be processed. This is not too big of a deal when running `tsc`, but it kills the language server when editing in VS code.

Instead, the transformer should not throw and instead output a `NullLiteral` or `ErrorNode` to the AST.

For errors, we will rely on a language service plugin (see: #10) to add domain-specific diagnostics instead of throwing an error.

thantos:
Compile will no longer throw when re-writing nodes in the ts-tranform. Instead a node of type `Err` is returned. Methods whichs expected the new Functionless nodes should check for `Err` when they also check for the new node types (generally `FunctionDecl`). They can throw the error provided to them from the `Err` node or handle the error how they would like.

Closes #11

### DIFF
--- a/src/appsync.ts
+++ b/src/appsync.ts
@@ -9,9 +9,10 @@ import {
   ToAttributeMap,
   ToAttributeValue,
 } from "typesafe-dynamodb/lib/attribute-value";
-import { FunctionDecl } from "./declaration";
+import { FunctionDecl, isFunctionDecl } from "./declaration";
 import { Literal } from "./literal";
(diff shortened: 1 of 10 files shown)


sam-goodwin (approved):

sam-goodwin:
This is a breaking change since the contract of reflect changed.

PR #54, opened 2022-03-29
fix: support excluding files from functionless transform, error nodes injection location, transforming children

thantos:
Pulling forward some changes from #39 and fixing issues caused by #42

* `Err` node was being injected over wrapping node like `AppResolver`, instead we should only replace the `FunctionDecl` using the new `errorBoundary` wrapper.
* Support the ability to exclude file paths during transform using a glob pattern. 
   * Use Case: Functionless source files are transformed during functionless testing. This doesn't normally happen when someone consumes functionless from npm.
* Pulling forward a change from #39 that fixes issue where the "children" of a transformed node are not correctly output. 
  * ex: `bus.when(event => event.source === "lambda").map(event => event)`. The `when` method would not be transformed because the `map` was found first.

Excluding some paths from functionless transform.

```json
  {
     "transform": "functionless/lib/compile",
     "exclude": "./src/{,**}/*"
  }
```

### DIFF
--- a/src/compile.ts
+++ b/src/compile.ts
@@ -276,7 +276,7 @@ export function compile(
             toExpr(node.expression),
             ts.factory.createArrayLiteralExpression(
               node.arguments.map((arg, i) =>
-                newExpr("ArgumentExpr", [
+                newExpr("Argument", [
                   toExpr(arg),
                   // the arguments array may not match the signature or the signature may be unknown
                   signature?.parameters?.[i]?.name
--- a/src/compile.ts
+++ b/src/compile.ts
@@ -5,7 +5,7 @@ import { AnyTable } from "./table";
 import { AnyLambda } from "./function";
 import { FunctionlessNode } from "./node";
 import { AppsyncResolver } from "./appsync";
-import glob from "glob";
+import minimatch from "minimatch";
 import { assertDefined } from "./assert";
 
 export default compile;
@@ -17,7 +17,7 @@ export interface FunctionlessConfig extends PluginConfig {
   /**
    * Glob to exclude
    */
-  exclude?: string;
+  exclude?: string[];
 }
 
 /**
@@ -35,10 +35,9 @@ export function compile(
   _config?: FunctionlessConfig,
   _extras?: TransformerExtras
 ): ts.TransformerFactory<ts.SourceFile> {
-  const ignore = _config?.exclude
-    ? new Set(glob.sync(_config.exclude, { absolute: true }))
-    : new Set();
-
+  const excludeMatchers = _config?.exclude
+    ? _config.exclude.map((pattern) => minimatch.makeRe(pattern))
+    : [];
   const checker = program.getTypeChecker();
   return (ctx) => {
     const functionless = ts.factory.createUniqueName("functionless");
@@ -55,7 +54,7 @@ export function compile(
       );
 
       // Do not transform any of the files matched by "exclude"
-      if (ignore.has(sf.fileName)) {
+      if (excludeMatchers.some((matcher) => matcher.test(sf.fileName))) {
         return sf;
       }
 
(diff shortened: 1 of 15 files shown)


src/compile.ts
```diff
@@ -470,7 +552,7 @@ function getOperator(op: ts.BinaryOperatorToken): BinaryOp | undefined {
   return OperatorMappings[op.kind as keyof typeof OperatorMappings];
 }
 
-const OperatorMappings = {
+const OperatorMappings: Record<number, BinaryOp> = {
```
sam-goodwin, line 555: Why explicit type? I think I was purposely using an inferred type.

src/expression.ts
```diff
@@ -98,20 +101,27 @@ export class ElementAccessExpr extends BaseNode<"ElementAccessExpr"> {
   }
 }
 
+export const isArgumentExpr = typeGuard("ArgumentExpr");
+
+export class ArgumentExpr extends BaseNode<"ArgumentExpr"> {
```
sam-goodwin, line 106: Duplicate change in two PRs?

thantos:
> I don't understand this use-case, can you elaborate?
> 
> > Use Case: Functionless source files are transformed during functionless testing. This doesn't normally happen when someone consumes functionless from npm.

When Jest funs, all of the source tree of Functionless goes through the functionless transform. 

I found this because I added the ability to directly use EventBusRule

```ts
new EventBusRule(this, 'rule', bus, (event) => event.source === "lambda")
```

However, the EventBus `.when` also creates an instance, but is passed an already transformed `FunctionDecl`.

This was causing all of the tests that use `EventBus` to fail because the compilers was trying to transformer the invalid from in `EventBus.when`.

Really, nothing in `/src` should ever be compiled, so I wanted to be able to skip it during transformation and I believe others may have code they want to be ignored too (maybe a performance thing, maybe they have another use case).

src/expression.ts
thantos: Yeah, this is on top of the other PR #47. I could set the other PR branch as the parent of this PR. Not sure the right way to handle concurrent PRs.

src/compile.ts
thantos: So the compiler fails when the operator isn't supported by `BinaryOp` so that other places have a chance to handle the value. For example, I added `!=` in #39, but function that is trying to handle all BinaryOps may be missed if it doesn't know about the new operator. 

Note: this was pulled forward from #39, could remove it from this PR.

src/compile.ts
```diff
@@ -1,15 +1,24 @@
...
 
 /**
  * Configuration options for the functionless TS transform.
  */
-export interface FunctionlessConfig extends PluginConfig {}
+export interface FunctionlessConfig extends PluginConfig {
+  /**
+   * Glob to exclude
+   */
+  exclude?: string;
```
sam-goodwin, line 20: should this be an array?

src/compile.ts
```diff
@@ -26,6 +35,10 @@ export function compile(
   _config?: FunctionlessConfig,
   _extras?: TransformerExtras
 ): ts.TransformerFactory<ts.SourceFile> {
+  const ignore = _config?.exclude
+    ? new Set(glob.sync(_config.exclude, { absolute: true }))
```
sam-goodwin, line 39: is this the most efficient way to do this? This could be a huge set?

sam-goodwin (approved):

src/compile.ts
thantos: My original thought was that glob supports crazy things, so keep it singular, but looks like multiple is easy, so lets do that. And it'll match `tsconfig`.
https://www.typescriptlang.org/tsconfig#include

src/compile.ts
thantos: Updated to use `minimatch` (used by npm internally) which turns globs into regex, can be cached, and can match a single value.

sam-goodwin (approved):

Issue #48, opened 2022-03-28
Event Bridge + Step Functions


thantos:
Dependent on #36 and #39 

Event bridge integrations with step functions. Start a step function integration using the typesafe pipe from the Functionless `EventBus`.

```ts
bus
   .when(...)
   .map(...)
   .pipe(myStepFunction)

bus
   .when(...)
   .pipe(myStepFunction)
```

Event bridge events send from step functions.

```ts
const createPerson = new ExpressStepFunction((arg: string): Person | undefined => {
   const person = {...};
   myBus({ "detail-type": "PersonCreated",  detail: person });
});
```

thantos:
### Hard Problem - When putting an event from step functions, the wrapper does not match the schema of event bridge.

Event Bridge Event Wrapper

```ts
export interface EventBusRuleInput<
  T = any,
  DetailType extends string = string,
  Source extends string = string
> {
  source: Source;
  "detail-type": DetailType;
  detail: T;
  version: string;
  id: string;
  account: string;
  time: string;
  region: string;
  resources: string[];
}
```  

However the task expects
```ts
{
          Resource: "arn:aws:states:::events:putEvents",
          Type: "Task",
          Parameters: {
            Entries: [
              {
                Detail: "string",
                DetailType: "string",
                EventBusName: "string",
                Resources: ["string"],
                Source: "string",
                Time: number,
                TraceHeader: "string",
              },
            ],
          },
        }
```

Now we have a few issues:

* All arguments need to be re-mapped to their capitalized or `dash-free` version
* We need to inject the bus arn into the object
* All arguments are technically optional, however step functions does not support missing json paths. This means that we either need some complex permutations using choice and pass in step functions (runtime) or to understand which keys will be present at build time.

### Short Term Solution

1. For users to give whole objects and not references
```ts
// Not supported
const myEvent = { "detail-type": "someType" }
bus(myEvent)

// Supported
bus({ "detail-type": "someType" })
```

### Followups

* Support resolution of constant references. If the reference is always a constant object which can be determined at build time, lets resolve the reference and use it.
* Support runtime determination of missing fields using Choice and Pass. This issue also exists for all other step function integrations.
* Complain to AWS some more.

PR #92, opened 2022-04-05
feat: step function object parameter, start step function invocation from event bridge

### LINKED ISSUE #91: Step Function interface as an Object
Current step function interface is an array of parameters. 

```ts
const func = new StepFunction((arg1: string, arg2: string): string => { return arg1 + arg2; });

// in app sync
const result = func("hello", "world");
```

### Problems

1. This is problematic for things like event bridge where the natural representation is an object.

```ts
new EventBus().when(...).map(event => [source]).pipe();
```

Returning an argument is awkward, additionally the integration between Step Functions and event bridge send an object, so magical transformation is needed.

2. At some point, step functions may accept a primitive value or array instead of an object, the argument syntax would be ambiguous in this situation.
3. Lambda uses object arguments and not argument arrays.

### Proposal

Accept an object instead of an array of arguments.

```ts
const func = new StepFunction<{ arg1: string, arg2: string}, string>((input)=> { return input.arg1 + input.arg2; });

// in app sync
const result = func({ arg1: "hello", arg2: "world"});

new EventBus().when().map(event => ({ arg1: event.id, arg2: "world" }).pipe(func );
```

### LINKED ISSUE #93: Event Bridge: Support Constants outside of the closure.
Part of #51 

The ability to reference data outside of the Event Bridge closure functions.

```ts
const stepFunction = new StepFunction();
const testSource = "someSource";

new EventBridge().when(event => event.resources.includes(stepFunction.stepFunctionArn) && event.source === testSource);
```

thantos:
Closes #91,  Closes #93
Part of #48 (missing the ability to send an event from a step function)

* [x] Change node === undefined to return undefined identity and not UndefinedExpression
* [x] Change SFN to accept an object and not an array of parameters
* [x] Change argument expression to be maybe undefined
* [x] Add ability to pipe to Event Bridge
* [x] Reference data outside of the event bus closures (`map` and `when`)
* [ ] Add ability to send event bridge notification in step functions

### Before Step Functions Parameter change

```ts
const func = new StepFunction((arg1: string, arg2: string): string => { return arg1 + arg2; });

// in app sync
const result = func("hello", "world");
```

### After Step Functions Parameter change #91

```ts
const func = new StepFunction<{ arg1: string, arg2: string}, string>((input)=> { return input.arg1 + input.arg2; });

// in app sync
const result = func({ arg1: "hello", arg2: "world"});

new EventBus().when().map(event => ({ arg1: event.id, arg2: "world" }).pipe(func );
```

### Support Invoking Step Functions From Event Bridge Events #48 

```ts
const myFunction = new StepFunction();

new EventBus().when(...).map(...).pipe(myFunction);
```

### Event Bridge Support Outside Constants #93

```ts
const stepFunction = new StepFunction();
const testSource = "someSource";

new EventBridge().when(event => event.resources.includes(stepFunction.stepFunctionArn) && event.source === testSource);
```

BREAKING CHANGE: Change to the signature of `StepFunction` and `ExpressStepFunction`.  Step Functions accept an `Input` object and a `Output` object type parameter. The input type is given to the StepFunction function as a single Object instead of an array of parameters.

### DIFF
--- a/src/asl.ts
+++ b/src/asl.ts
@@ -3,6 +3,7 @@ import { aws_iam, aws_stepfunctions } from "aws-cdk-lib";
 
 import { assertNever } from "./assert";
 import {
+  Argument,
   CallExpr,
   ElementAccessExpr,
   Expr,
@@ -35,9 +36,12 @@ import {
   WhileStmt,
 } from "./statement";
 import { anyOf, findFunction } from "./util";
-import { FunctionDecl } from "./declaration";
+import { FunctionDecl, isParameterDecl, isFunctionDecl } from "./declaration";
 import { FunctionlessNode } from "./node";
 import { visitEachChild } from "./visit";
+import { isFunction } from "./function";
+import { isTable } from "./table";
+import { isStepFunction } from "./step-function";
 
 export function isASL(a: any): a is ASL {
   return (a as ASL | undefined)?.kind === ASL.ContextName;
@@ -681,13 +685,15 @@ export class ASL {
         throw new Error(`the expr of a ThrowStmt must be a NewExpr`);
       }
 
-      const error = (stmt.expr as NewExpr).args.reduce(
-        (args: any, arg) => ({
-          ...args,
-          [arg.name!]: ASL.toJson(arg.expr),
-        }),
-        {}
-      );
+      const error = (stmt.expr as NewExpr).args
+        .filter((arg): arg is Argument & { expr: Expr } => !!arg.expr)
+        .reduce(
+          (args: any, arg) => ({
+            ...args,
+            [arg.name!]: ASL.toJson(arg.expr),
+          }),
+          {}
+        );
 
       const throwTransition = this.throw(stmt);
       if (throwTransition === undefined) {
@@ -1232,8 +1238,10 @@ function hasBreak(loop: ForInStmt | ForOfStmt | WhileStmt | DoStmt): boolean {
 }
 
 export namespace ASL {
-  export function toJson(expr: Expr): any {
-    if (expr.kind === "Argument") {
+  export function toJson(expr?: Expr): any {
+    if (!expr) {
+      return undefined;
+    } else if (expr.kind === "Argument") {
       return toJson(expr.expr);
     } else if (expr.kind === "CallExpr") {
       if (isSlice(expr)) {
@@ -1293,11 +1301,11 @@ export namespace ASL {
       return expr.value ?? null;
     } else if (expr.kind === "ReferenceExpr") {
       const ref = expr.ref();
-      if (ref.kind === "Function") {
+      if (isFunction(ref)) {
         return ref.resource.functionArn;
-      } else if (ref.kind === "StepFunction") {
+      } else if (isStepFunction(ref)) {
         return ref.stateMachineArn;
-      } else if (ref.kind === "Table") {
+      } else if (isTable(ref)) {
         return ref.resource.tableName;
       }
     } else if (expr.kind === "ElementAccessExpr") {
@@ -1325,6 +1333,17 @@ export namespace ASL {
         return filterToJsonPath(expr);
       }
     } else if (expr.kind === "Identifier") {
+      const ref = expr.lookup();
+      // If the identifier references a parameter expression and that parameter expression
+      // is in a FunctionExpr or FunctionDecl and that Function is at the top (no parent).
+      if (
+        ref &&
+        isParameterDecl(ref) &&
+        anyOf(isFunctionExpr, isFunctionDecl)(ref.parent) &&
+        ref.parent.parent === undefined
+      ) {
+        return "$";
+      }
       return `$.${expr.name}`;
     } else if (expr.kind === "PropAccessExpr") {
       return `${toJsonPath(expr.expr)}.${expr.name}`;
@@ -1347,8 +1366,9 @@ export namespace ASL {
 
     const end = expr.getArgument("end")?.expr;
     if (
-      end?.kind !== "NullLiteralExpr" &&
-      end?.kind !== "UndefinedLiteralExpr"
+      !!end?.kind &&
+      end.kind !== "NullLiteralExpr" &&
+      end.kind !== "UndefinedLiteralExpr"
     ) {
       if (end?.kind !== "NumberLiteralExpr") {
         throw new Error(
@@ -1622,12 +1642,14 @@ export namespace ASL {
           isLiteralOrTypeOfExpr(expr.left) ||
           isLiteralOrTypeOfExpr(expr.right)
         ) {
-          const [literalOrTypeOf, val] =
-            isLiteralOrTypeOfExpr(expr.left) || isLiteralOrTypeOfExpr(expr.left)
-              ? [expr.left, expr.right]
-              : [expr.right, expr.left];
+          // typeof x === "string" -> left: typeOf, right: literal
+          // "string" === typeof x -> left: literal, right: typeOf
+          // x === 1 -> left: identifier, right: literal
+          const [literalExpr, val] = isLiteralExpr(expr.left)
+            ? [expr.left, expr.right]
+            : [expr.right, expr.left];
 
-          if (literalOrTypeOf.kind === "TypeOfExpr") {
+          if (val.kind === "TypeOfExpr") {
             const supportedTypeNames = [
               "undefined",
               "boolean",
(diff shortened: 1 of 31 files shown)


src/compile.ts
```diff
@@ -746,6 +771,46 @@ export function compile(
...
+       *
+       * return { value: () => val };
+       */
+      function outOfScopeIdentifierToRef(
+        symbol: ts.Symbol,
+        scope: ts.Node
+      ): ts.NewExpression | undefined {
+        if (symbol) {
+          if (symbol.valueDeclaration) {
+            // Identities in Shorthand Property Assignment value declarations return the shorthand prop assignment and not the value.
```
sam-goodwin, line 790: Should this be:

> Identifies if

src/step-function.ts
```diff
@@ -413,14 +434,16 @@ abstract class BaseStepFunction<F extends AnyFunction>
...
-            Input: call.args.reduce(
-              (args, arg) => ({
-                ...args,
-                [`${arg.name!}${isVariableReference(arg.expr) ? ".$" : ""}`]:
-                  ASL.toJson(arg.expr),
-              }),
-              {}
-            ),
+            Input: call.args
+              .filter((arg): arg is Argument & { expr: Expr } => !!arg.expr)
```
sam-goodwin, line 438: hmm, why not `call.args[0]`?

src/step-function.ts
```diff
@@ -915,17 +939,28 @@ export type SyncExecutionResult<T> =
   | SyncExecutionFailedResult
   | SyncExecutionSuccessResult<T>;
 
-export interface ExpressStepFunction<F extends AnyFunction> {
-  (...args: Parameters<F>): SyncExecutionResult<ReturnType<F>>;
+type ConditionalMachine<P, O> = P extends undefined
```
sam-goodwin, line 942: How about just replace this with the entire `AWS.StepFunctions.StartSyncExecution` type, except substitute `Input` with the user's type?

sam-goodwin (changes requested):

src/step-function.ts
thantos: Nevermind, I think we need to handle all of the paths...

Creating the variable signature in typescript is messy and I can't get it to work... also leaves the string parameters ambiguous.

thantos: Going to start without support for the object input form, that way we can build time check the keys, can add the object form later.

src/asl.ts
```diff
@@ -1622,12 +1642,14 @@ export namespace ASL {
...
-          const [literalOrTypeOf, val] =
-            isLiteralOrTypeOfExpr(expr.left) || isLiteralOrTypeOfExpr(expr.left)
-              ? [expr.left, expr.right]
-              : [expr.right, expr.left];
-
-          if (literalOrTypeOf.kind === "TypeOfExpr") {
+          // typeof x === "string" -> left: typeOf, right: literal
+          // "string" === typeof x -> left: literal, right: typeOf
+          // x === 1 -> left: identifier, right: literal
+          const [literalExpr, val] = isLiteralExpr(expr.left)
```
sam-goodwin, line 1648: what happened to the typeof logic?

src/assert.ts
```diff
@@ -32,6 +33,38 @@ export function assertDefined<T>(
   return value as Exclude<T, undefined>;
 }
 
+export function assertPrimitive(val: any, message?: string): PrimitiveValue {
+  if (isPrimitive(val === "string")) {
```
sam-goodwin, line 37: Huh? this will always be a boolean.

src/compile.ts
```diff
@@ -490,6 +496,23 @@ export function compile(
             return ref(node);
           }
 
+          const symbol = checker.getSymbolAtLocation(node);
+          /**
+           * If the identifier is not within the closure, we attempt to enclose the reference in its own closure.
+           * cosnt val = "hello";
```
sam-goodwin, line 502: nit: typo

src/expression.ts
```diff
@@ -200,13 +205,13 @@ export class ElementAccessExpr extends BaseExpr<"ElementAccessExpr"> {
 export const isArgument = typeGuard("Argument");
 
 export class Argument extends BaseExpr<"Argument", CallExpr | NewExpr> {
-  constructor(readonly expr: Expr, readonly name?: string) {
+  constructor(readonly expr?: Expr, readonly name?: string) {
```
sam-goodwin, line 208: since when did expr become optional?

test/appsync.test.ts
```diff
@@ -0,0 +1,293 @@
...
+    }
+  }
+}`,
+      "{}",
+      VTL.CircuitBreaker
+    );
+
+    const templates = getAppSyncTemplates(func);
+
+    appsyncVelocityJsonTestCase(
```
sam-goodwin, line 212: this is great!

test/step-function.test.ts
```diff
@@ -3912,20 +4000,24 @@ test("try { list.forEach(item => task(item)) }", () => {
...
 test("try { list.forEach(item => throw) }", () => {
   const { stack } = initStepFunctionApp();
-  const definition = new ExpressStepFunction(stack, "fn", (list: string[]) => {
-    try {
-      return list.forEach(() => {
-        throw new Error("cause");
-      });
-    } catch {
-      return null;
+  const definition = new ExpressStepFunction<{ list: string[] }, void | null>(
```
sam-goodwin, line 4003: do you prefer this way of modeling the function signature? I would much prefer it on the actual function. I think most people would prefer that also since it is more familiar?

```ts
(input: {list: string[]}): void | null => ..
```

test/step-function.test.ts
```diff
@@ -5348,3 +5583,139 @@ test("call Step Function from another Step Function", () => {
...
+  const { stack } = initStepFunctionApp();
+  const machine1 = new ExpressStepFunction<{ value: string }, string>(
+    stack,
+    "machine1",
+    () => {
+      return "hello";
+    }
+  );
+
+  expect(
```
sam-goodwin, line 5643: We should look into refactoring the Step Function test cases into a utility function that just takes the closure - that way we can migrate all of these test cases to use https://localstack.cloud one day.

sam-goodwin (changes requested):

test/step-function.test.ts
thantos: #97

test/step-function.test.ts
thantos: I actually don't think its more familiar because we generally do not need to put types on arrow function signatures unless necessary

I think it makes more sense to add it on the class because other types and methods may rely on the generic types than just the closure method.

That said, I am fine with either. For some reason I didn't think the inference would work, but it does.

```ts
  const definition = new ExpressStepFunction(
    stack,
    "fn",
    (input: { input: { id: string } }) => {
      return input.input.id;
    }
  ).definition;
```

thantos: I think a developer is more likely to add it to the type parameters, but I am glad we support both.

src/expression.ts
thantos: This is because of the signature processing for CallExprs. 

All arguments in the signature are added to the array, but were creating an undefined literal expression for all unprovided arguments. So I changed it to instead provide `undefined` for unprovided arguments. It maintains the named argument still.

src/asl.ts
thantos: I switched the logic, you had `literalOrTypeOf`. It was kind of awkward because the expression in typeof is the typeof and the literal is the other side. For a literal, it is the opposite, the literal is the literal and the expression is the other side.

So I look for the literal, which is `string`, `number`, etc. With the opposite side being an expression or a `typeof`.

There was a failing use case (see the change to the typeof test) where `"string" === typeof x` didn't work. 

So now the logic is simple, 1) look for the literal, 2) decide if one side is a `typeof`, 3) process.

src/assert.ts
thantos: haha... bad refactor.. thanks

src/asl.ts
```diff
@@ -1325,6 +1333,17 @@ export namespace ASL {
...
       }
     } else if (expr.kind === "Identifier") {
+      const ref = expr.lookup();
+      // If the identifier references a parameter expression and that parameter expression
+      // is in a FunctionExpr or FunctionDecl and that Function is at the top (no parent).
+      if (
+        ref &&
+        isParameterDecl(ref) &&
+        anyOf(isFunctionExpr, isFunctionDecl)(ref.parent) &&
+        ref.parent.parent === undefined
```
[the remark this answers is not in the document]
thantos: Nice, I wasn't sure.

src/asl.ts
```diff
@@ -1293,11 +1301,11 @@ export namespace ASL {
       return expr.value ?? null;
     } else if (expr.kind === "ReferenceExpr") {
       const ref = expr.ref();
-      if (ref.kind === "Function") {
+      if (isFunction(ref)) {
```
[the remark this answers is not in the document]
thantos: So this is in part because ref is no longer just for services. It can now return anything, so the type essentially `unknown`.

```ts
export type CanReference =
  | AnyTable
  | AnyLambda
  | AnyStepFunction
  | typeof $AWS
  | unknown;
```

The guards work for `unknown`, but `.kind` does not.

```ts
const sfn = new StepFunction();
new EventBridge.when(event => sfn.stepFunctionArn);
```

=>

```ts
propAccess(ref(() => sfn), "stepFunctionArn")
```

sam-goodwin (approved):

thantos:
Found more bugs, new and existing, adding tests and fixing...

Related to step function invoke dynamic paths, describe execution parameters, and describe execution not being supported by expression step functions.

Issue #23, opened 2022-03-03
API Gateway mapping templates


garretcharp:
Is it currently possible to use this for api gateway mapping templates as well? If not support for this would be nice!

sam-goodwin:
Yes, this is totally something we should do. We can make use of the existing TS->VTL code!

Let's define the requirements for a new construct, `ApiIntegration`:

1. type-safe representation of the API Gateway's velocity template environment

Taking a lesson from this feedback (https://github.com/sam-goodwin/functionless/issues/24), we should first implement a layer that is un-opinionated and 1:1 with the underlying input to an API Gateway Velocity Template.

This means defining static types for the `$util`, `$input` and `$context` global variables. 
- [ ] `$input` - https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-mapping-template-reference.html#input-variable-reference
- [ ] `$context` - https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-mapping-template-reference.html#context-variable-reference
- [ ] `$util` - https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-mapping-template-reference.html#util-template-reference

The `$input` type depends on the Method's path, query params and body. How can the developer safely define types for `$input`? 

Simplest idea is to define an interface with `params`, `body` and `path`. This maps 1:1 with the API Gateway configuration.
```ts
new ApiIntegration<{
  body: BodyType,
  params: ParamsType,
  path: PathType
}>(($input) => {
  $input
});
```

Perhaps we can do better by specifying the query params and path as a string, see https://twitter.com/mpocock1/status/1499002040168636420?s=20&t=j1cgOZeaFkevSnG4nOHl5A, but this may be going too far?

`$util` unfortunately collides with Appsync's `$util` object, so perhaps we should move them into namespaces:

```ts
appsync.$util
apigw.$util
```

2. API Gateway V1 integration

If you use the AWS CDK to generate APIs, developers create hierarchies of `Resource`s and add `Method`s to them (i.e. GET, POST, PUT, etc.). See: https://docs.aws.amazon.com/cdk/api/v1/docs/aws-apigateway-readme.html

`functionless`'s `ApiIntegration` can implement the [`Integration` interface](https://docs.aws.amazon.com/cdk/api/v1/docs/aws-apigateway-readme.html#integration-targets) 
```ts
const api = new apigateway.RestApi(this, "hello-api");
const person = api.root.addResource("people");

[...]

PR #98, opened 2022-04-06
chore: add graphql codegen typescript generation to test app

mathisobadia:
This pull request adds graphql codegen to the test-app, this allows generating typescript code for all resolvers in the schema automatically.
Main benefit is that this ensures accurate types for the resolvers on any schema change

Discussion: 
I have added some documentation on this in the readme but I'm honestly not sure the end of the file is the best place to put this. Maybe creating a new readme in the test-app would make more sense?

We discussed this PR a while ago so I don't know if it still has it's place in this project or not but since I had already done the work I though I would still open it and discuss to see if it needs changes or if it would better live in an example repo I create.

### DIFF
--- a/README.md
+++ b/README.md
@@ -398,9 +398,7 @@ interface Delete {
 const createOrUpdateOperation = functionless.Function<CreateOrUpdate, void>(
   createOrUpdateFunction
 );
-const deleteOperation = functionless.Function<Delete, void>(
-  deleteFunction
-);
+const deleteOperation = functionless.Function<Delete, void>(deleteFunction);
 ```
 
 The events from before do not match the formats from before, so lets transform them to the structures match.
@@ -1319,7 +1317,7 @@ Two plugins are necessary to generate resolver types:
 - [typescript](https://www.graphql-code-generator.com/plugins/typescript)
 - [typescript-resolver](https://www.graphql-code-generator.com/plugins/typescript-resolvers)
 
-Both of those plugins need to be configured to generate types that can be easily imported into your app.
+Both of those plugins need to be configured by creating a [codegen.yml](./src/test-app/codegen.yml) file.
 
 ```yaml
 overwrite: true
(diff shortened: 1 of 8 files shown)


mathisobadia:
Yes for sure generating the GQL schema from typescript would result in an even better experience, as typescript types are more expressive and composable.  Then open questions there are similar to API gateway, one of those being how to we map typescript types to [scalar types](https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html) maybe we create a typescript type for each of those scalars that is exported by functionless and can be used in the type definitions.
Such as 

``` javascript
import { ID, AWSDateTime } from 'functionless'
interface Person = {
  id: ID,
  createdAt: AWSDateTime | undefined
}
```
 
which would result in 
```
 type Person {
  id: ID!
  createdAt: AWSDateTime
}
```

I'm wondering what exact solution will end up being chosen in #23 as we would want to choose something similar

(to answer that last question yes it's just to show how they can be used together so I agree that it's debatable wether this belongs in this repo, I though it was interesting to see how these two code generation tools could be used together in the test-app)

sam-goodwin:
> Wouldn't the patterns of functionless say we should generate the .gql file from typescript and not generate typescript from the GQL?

We should support both patterns equally. Lots of existing projects already have a schema and we want to support them!

test-app/src/generated-types.ts
```diff
@@ -0,0 +1,182 @@
+import { GraphQLResolveInfo } from 'graphql';
```
sam-goodwin, line 1: What's best practice here? Check in the file or not?

test-app/tsconfig.json
```diff
@@ -28,7 +28,10 @@
       {
         "transform": "functionless/lib/compile"
       }
-    ]
+    ],
+    "paths": {
+      "functionless": ["../"]
```
sam-goodwin, line 33: Hmm what does this do? I've been using yarn link, does this improve that experience?

sam-goodwin:
Thanks so much for this change, so great to show how easy it is to integrate with existing tooling practices.

sam-goodwin:
I wonder if a projen template could automate the set up of this for people? Something for us to look into later.

test-app/src/generated-types.ts
mathisobadia: So I just saw this [Paul Swail post](https://serverlessfirst.com/emails/graphql-codegen-for-appsync-projects/) that advises against checking in the generated types. In my projects I've personally checked them in because I like to see type changes in pull requests. I think we could choose not to check them but we would need to make sure that the codegen script is run before build and deploy, maybe even on yarn install? So that the user does not need to run yarn codegen manually after cloning the repo.

test-app/tsconfig.json
mathisobadia: Oh there might be something wrong with my environment then because I kept having typescript errors here until I added this. What yarn version should I use with this project?

test-app/src/generated-types.ts
sam-goodwin: I understand what you mean - like a snapshot test. Let's go with it - it'll be big on the first commit but changes should be irregular.

sam-goodwin: Also, it's good for our sample app to show them for people browsing.

test-app/tsconfig.json
sho-ai-main: I removed it with latest commit, I think it was something specific to my setup

README.md
```diff
@@ -1309,3 +1309,44 @@ See the following files to understand the structure of the Abstract Syntax Tree:
...
+Functionless can be used together with [graphql code generator](https://www.graphql-code-generator.com/) to automatically generate types from the schema.
+
+Two plugins are necessary to generate resolver types:
+
+- [typescript](https://www.graphql-code-generator.com/plugins/typescript)
+- [typescript-resolver](https://www.graphql-code-generator.com/plugins/typescript-resolvers)
+
+Both of those plugins need to be configured to generate types that can be easily imported into your app.
+
+```yaml
```
sam-goodwin, line 1324: What's the name of this file?

README.md
```diff
@@ -1309,3 +1309,44 @@ See the following files to understand the structure of the Abstract Syntax Tree:
...
+      customResolverFn: "{ args: TArgs; context: TContext; result: TResult; source: TParent;}"
+      # appsync allows returnning undefined instead of null only when a type is optional
+      maybeValue: T | null | undefined
+      # typename is not really usefull for resolvers and can cause clashes in the case where a type extends another type but have different names
+      skipTypename: true
+```
+
+you can then use `npx graphql-codegen --config codegen.yml` to generate the types, you should regenerate them any time you update your schema
+
+The generated types will include a `Resolvers` type that can be used to reference the types of all resolvers and get access to the type of parameters, result and source. e.g. `type GetPersonQueryArgs = Resolvers["Query"]["getPerson"]['args']`
```
sam-goodwin, line 1350: Maybe have one example as a snippet here?

sam-goodwin (changes requested):

README.md
mathisobadia: I've updated the doc to add a code snippet and some links to relevant files, let me know what you think!

sam-goodwin (approved):

PR #103, opened 2022-04-22
feat: native typescript lambda function code

### LINKED ISSUE #153: Native Lambda Functions
issue to track #103 as part of the beta v1 project. See #103 for details.

thantos:
Closes #153 

Depends on #109 and #108 

Function closure serialization!

- [x] Basic Function serialization
- [x] Add constructor to Function to pass in a closure, scope, and id
- [x] ~~Exhaustive~~ tests
- [x] Support embedded client calls
   - [x] Step Functions
   - [x] Lambda
   - ~~App Sync~~ include in #119
   - ~~Api Gateway Endpoints~~ Not implemented - include in #23
   - [x] Event Bridge
   - [x] Dynamo
      - Only within `$AWS.DynamoDB.*` see #132
- [x] Support arbitrary token data as env variables
- [x] Split localstack integration into a separate PR #109 
- [x] AppSync integration changes #113 
- Dynamically simplify native integration references #122 

```ts
new Function(this, 'myFunction', async (event) => {
   return Object.keys(event).length;
})
```


src/declaration.ts
```diff
@@ -37,6 +37,31 @@ export class FunctionDecl<F extends AnyFunction = AnyFunction> extends BaseDecl<
   }
 }
 
+export const isHoistedFunctionDecl = typeGuard("HoistedFunctionDecl");
+
+/**
+ * Instead of serializing this function to AST, this function has been hoisted to a module export with a random name.
+ */
+export class HoistedFunctionDecl<
```
sam-goodwin, line 45: How will we determine the dependencies and auto wire environment variables and iam policies? Walk the FunctionDecl like we do for the functionless integrations?

sam-goodwin:
Do we want to pull out pulumi's closure serializer into its own package to avoid the massive dependency ?

src/declaration.ts
thantos: Ohh, ha, this needs to be cleaned up, from before using Pulumi.

thantos:
> Do we want to pull out pulumi's closure serializer into its own package to avoid the massive dependency ?

Its just a build time dependency, does it matter?

I'd like to write our own in the longer term.

sam-goodwin:
Can do it later, sure. But I also noticed it has a dependency on their concept of a Resource for logging. I was wondering if that mattered at all.

thantos:
> Can do it later, sure. But I also noticed it has a dependency on their concept of a Resource for logging. I was wondering if that mattered at all.

Looking into this. I am not a fan of copying/forking libraries, especially complex ones. But its not OK if there is any Pulumi specific solutions.

.projenrc.js
```diff
@@ -1,4 +1,7 @@
 const { typescript } = require("projen");
+
+const MIN_CDK_VERSION = "2.20.0";
```
sam-goodwin, line 3: Do we need to specify the version or can we just use default. Projen defaults to latest.

README.md
```diff
@@ -558,6 +559,31 @@ bus
...
+Normal use of the CDK through the CLI should work just fine without doing anything special.
+
+The problem comes when using the explicit `app.synth()` method. This is a common case if trying to test your CDK code through testing tooling like `jest`.
+
+CDK does not natively support async code through the constructs, however, Functionless is using `pulumi`'s `serializeFunction` which is an async function. 
+
+When using `.synth()` programmatically, use the provided `asyncSynth` method to wrap your app.
+
+```ts
+const cloudAssembly = asyncSynth(app, options);
```
sam-goodwin, line 584: Maybe show a test case for clarity. Also, should this be `await asyncSynth`?

src/declaration.ts
```diff
@@ -37,6 +37,31 @@ export class FunctionDecl<F extends AnyFunction = AnyFunction> extends BaseDecl<
...
+
+/**
+ * A function declaration which contains the original closure instead of Functionless expressions.
+ */
+export class NativeFunctionDecl<
+  F extends AnyFunction = AnyFunction
+> extends BaseDecl<"NativeFunctionDecl", undefined> {
+  readonly _functionBrand?: F;
+  constructor(
+    readonly parameters: ParameterDecl[],
```
sam-goodwin, line 50: Why don't we also include the `FunctionDecl`? We can use it to auto-wire IAM permissions and environment variables.

src/function.ts
```diff
@@ -30,11 +38,50 @@ export type AnyLambda = Function<any, any>;
...
+  public static readonly FunctionlessType = "Function";
+  readonly resource: aws_lambda.IFunction;
+
+  static promises: Promise<any>[] = [];
 
   // @ts-ignore - this makes `F` easily available at compile time
   readonly __functionBrand: ConditionalFunction<P, O>;
 
-  constructor(readonly resource: aws_lambda.IFunction) {
+  constructor(
```
sam-goodwin, line 49: We should add thorough code documentation with examples for each variant.

src/function.ts
```diff
@@ -80,3 +127,53 @@ export interface Function<P, O> {
     ConditionalFunction<P, O>
   >;
 }
+
+export class CallbackLambdaCode extends aws_lambda.Code {
+  constructor(private func: AnyFunction) {
+    super();
+  }
+
+  bind(scope: Construct): aws_lambda.CodeConfig {
```
sam-goodwin, line 136: nit: `public bind`

src/function.ts
```diff
@@ -30,11 +38,50 @@ export type AnyLambda = Function<any, any>;
  */
 export class Function<P, O> {
   readonly kind = "Function" as const;
+  public static readonly FunctionlessType = "Function";
+  readonly resource: aws_lambda.IFunction;
+
+  static promises: Promise<any>[] = [];
```
sam-goodwin, line 44: Should we make sure this is a global singleton? Or do we want to rely on the `functionless` module being a singleton because of `peerDeps`? 

My preferred method is with a symbol:

```ts
const PromisesSymbol = Symbol.for("functionless.Function.promises");

public static readonly promises = (global[PromisesSymbol] = global[PromisesSymbol] ?? []);
```

src/function.ts
```diff
@@ -80,3 +127,53 @@ export interface Function<P, O> {
...
+      },
+    });
+
+    const funcResource = scope.node.findChild(
+      "Resource"
+    ) as aws_lambda.CfnFunction;
+
+    const codeConfig = asset.bind(scope);
+
+    funcResource.code = {
```
sam-goodwin, line 169: Let's give a link to cloudy as a thankyou

sam-goodwin (changes requested):

.projenrc.js
thantos: As a library, we should pin to the version we suggest in the peer deps. I was having trouble keeping them all the same version as I added new CDK deps, the latest would be installed, but the older ones were not being upgrading, existentially the one in alpha. Having this constant made it easier to keep them the same version.

src/declaration.ts
thantos: I had a feeling of need it eventually, but wasn't sure. This is probably where we'd end up.

src/declaration.ts
thantos: There are a few options here, once is to use the serialize function like Cloudy does, another is to use codegen. I will figure this out today.

README.md
thantos: Not sure if I want to show a full test, actually deploying the template is verbose. Will link to an example.

src/function.ts
thantos: In a future PR, we should import for the existing case instead of using the constructor.

```ts
new Function(this, id, async () => {})

Function.fromFunction(new aws_lambda.Function(...));
```

src/function.ts
thantos: Good idea.

thantos:
Current concerns with using Pulumi:
1. Customization - Pulumi is not a hookable solution. It is possible to use the `serialize` function to ignore variables, but not rewrite them. Best you can do is ignore them.
2. Ignoring nodes is all or nothing - In `serialize`, I can ignore a node like all CDK Constructs, but I cannot get a property off of that object to use later.
3. Serialize doesn't return top level nodes in the closure - `(event) => event` never sends the `event` object to the serialize function.
4. Logging - There is a logging Resource parameter, which is Pulumi's `Resource`, still not sure the impact of this yet.

thantos:
Caveat: References to functions in namespaces should be simplified to keep the resulting native function bundle small.

### Impacts:
1. Bundle Size
2. Invalid Bundles during tests

### Example

```ts
new Function(..., () => {
   $AWS.EventBridge.putEvents(...);
})
```

Should be

```ts
const putEvents = $AWS.EventBridge;
new Function(..., () => {
   putEvents(...);
})
```

### Reason

1. Pulumi will serialize the entirety of $AWS and all sub-variables when using the `$AWS.EventBridge.putEvents` form. 
   1. Using the referenced from has reduced bundle sizes by over 10x
2. Invalid function are created, specifically during tests, when Pulumi tries to include code generated by jest for global functions.
   1. The referenced form seems less likely to have this issue.

### Mitigation
1. Users can follow the pattern above
2. Support Automatic Reference of native integrations (future PR: )
3. Consider avoiding integrations in keep objects? `import { awsEventBridgePutEvents } from 'functionless'` instead of `import $AWS from 'functioness'` ?

package.json
```diff
@@ -31,7 +31,7 @@
     "@types/fs-extra": "^9.0.13",
     "@types/jest": "^27.5.0",
     "@types/minimatch": "^3.0.5",
-    "@types/node": "^14",
+    "@types/node": "^12",
```
sam-goodwin: hmm what's going on here? Node 12 has recently reached end of life. 

https://endoflife.date/nodejs

src/compile.ts
```diff
@@ -1251,6 +1204,47 @@ export function compile(
         return;
       }
 
+      function getOutOfScopeValueNode(
```
sam-goodwin, line 1207: can you add a quick comment with some examples? What is this being used for?

src/compile.ts
```diff
@@ -1251,6 +1204,47 @@ export function compile(
...
+              scope
+            );
+          }
+        }
+        return undefined;
+      }
+
+      function isSymbolOutOfScope(symbol: ts.Symbol, scope: ts.Node): boolean {
+        if (symbol) {
+          if (symbol.valueDeclaration) {
```
sam-goodwin, line 1232: nit: `if (symbol?.valueDeclaration)`

test/function.localstack.test.ts
```diff
@@ -239,32 +239,54 @@ localstackTestSuite("functionStack", (testResource, _stack, app) => {
...
+      const func = new Function(parent, "function", async () => {
+        const busbus = bus;
+        busbus({
+          "detail-type": "anyDetail",
+          source: "anySource",
+          detail: {},
+        });
+      });
+
+      return { outputs: { function: func.resource.functionName } };
```
sam-goodwin, line 257: nit: forgetting prettier again?

test/function.localstack.test.ts
```diff
@@ -239,32 +239,54 @@ localstackTestSuite("functionStack", (testResource, _stack, app) => {
...
+    (parent) => {
+      const bus = new EventBus<EventBusRuleInput>(parent, "bus");
+      const func = new Function(parent, "function", async () => {
+        const busbus = bus;
+        busbus({
+          "detail-type": "anyDetail",
+          source: "anySource",
+          detail: {},
+        });
+      });
```
sam-goodwin, line 255: Why does it fail? Seems fine to me - we're just referencing a value.

sam-goodwin (changes requested):

src/aws.ts
```diff
@@ -361,4 +361,40 @@ export namespace $AWS {
...
+      ) => AWS.EventBridge.Types.PutEventsResponse,
+      "EventBridge.putEvent"
+    >({
+      kind: "EventBridge.putEvent",
+      native: {
+        // Access needs to be granted manually
+        bootstrap: () => {},
+        /* istanbul ignore next */
+        preWarm: (prewarmContext: NativePreWarmContext) => {
+          /* istanbul ignore next */
```
sam-goodwin, line 381: what is this instanbul thing?

src/compile.ts
```diff
@@ -233,6 +251,18 @@ export function compile(
...
+      //   return ts.isCallExpression(node) && isEventBus(node.expression);
+      // }
+
+      // function isFunctionCall(node: ts.Node): node is ts.CallExpression {
+      //   return ts.isCallExpression(node) && isFunction(node.expression);
+      // }
+
+      // function isFunction(node: ts.Node) {
+      //   return isFunctionlessClassOfKind(node, Function.FunctionlessType);
+      // }
```
sam-goodwin, line 264: remove?

src/compile.ts
```diff
@@ -269,6 +299,41 @@ export function compile(
...
+      //     (s) => s.name === "Construct"
+      //   );
+
+      //   return (
+      //     (!constructExportSymbol ||
+      //       type.symbol === constructExportSymbol ||
+      //       type.getBaseTypes()?.some((t) => isCDKConstruct(t))) ??
+      //     false
+      //   );
+      // }
```
sam-goodwin, line 335: remove?

src/compile.ts
```diff
@@ -407,6 +472,291 @@ export function compile(
...
+      interface NativeExprContext {
+        preWarmContext: ts.Identifier;
+        closureNode: TsFunctionParameter;
+        registerIntegration: (node: ts.Expression) => void;
+      }
+
+      /**
+       * Native Functions do not allow the creation of resources, CDK constructs or functionless.
+       */
+      function toNativeFunction(
```
sam-goodwin, line 503: Provide a comment that shows an example of how the code is transformed. I'm particularly interested in preWarmContext.

src/compile.ts
```diff
@@ -407,6 +472,291 @@ export function compile(
...
+              node.expression,
+              nativeExprContext.closureNode
+            );
+
+            if (!outOfScopeIntegrationReference) {
+              throw Error(
+                "Integration Variable must be defined out of scope: " +
+                  node.expression.getText()
+              );
+            }
```
sam-goodwin, line 598: Why?

src/compile.ts
```diff
@@ -407,6 +472,291 @@ export function compile(
...
+                  ]
+                )
+              ),
+            ];
+          }
+        } else if (ts.isNewExpression(node)) {
+          const newType = checker.getTypeAtLocation(node);
+          // cannot create new resources in native runtime code.
+          const functionlessKind = getFunctionlessTypeKind(newType);
+          if (getFunctionlessTypeKind(newType)) {
```
sam-goodwin, line 635: Should this be expanded to any class that extends Construct?

src/compile.ts
```diff
@@ -407,6 +472,291 @@ export function compile(
...
+            throw Error(
+              `Cannot initialize new resources in a native function, found ${functionlessKind}.`
+            );
+          }
+          // TODO: check for CDK constructs
+          // const type = checker.getTypeAtLocation(node.expression);
+          // if (isCDKConstruct(type)) {
```
sam-goodwin, line 642: Ah I see. What's the hold up here? Not sure how to determine the type?

src/compile.ts
```diff
@@ -407,6 +472,291 @@ export function compile(
...
+          // }
+        } else if (ts.isReturnStatement(node)) {
+          const child = node.expression
+            ? toNativeExpr(node.expression, context, nativeExprContext)
+            : undefined;
+          if (!child) {
+            return node;
+          } else if (Array.isArray(child)) {
+            // if multiple nodes are returned, assume the last one is the value to assign and the rest should be injected as siblings.
+            const [last, ...rest] = child.reverse();
```
sam-goodwin, line 655: I don't understand?

sam-goodwin: Need comments with examples.

src/compile.ts
```diff
@@ -407,6 +472,291 @@ export function compile(
...
+              ),
+            ];
+          } else {
+            return context.factory.updateReturnStatement(
+              node,
+              child as ts.Expression
+            );
+          }
+        } else if (ts.isVariableDeclarationList(node)) {
+          // bubble any extra nodes to the top of the variable declaration list.
```
sam-goodwin, line 670: why? I don't understand.

test/function.localstack.test.ts
thantos: Pulumi tries to serialize the entire construct and finds some native functions it can't serialize.

src/compile.ts
```diff
@@ -407,6 +472,291 @@ export function compile(
...
+              node,
+              node.name,
+              node.exclamationToken,
+              node.type,
+              child.length > 0
+                ? (child[child.length - 1] as ts.Expression)
+                : undefined
+            );
+            return [
+              ...(child.length > 1 ? child.slice(0, child.length - 1) : []),
```
sam-goodwin, line 713: Wha tis going on here? Why do we need to rearrange things like this?

src/event-bridge/event-bus.ts
```diff
@@ -120,9 +124,37 @@ abstract class EventBusBase<E extends EventBusRuleInput>
...
     this.eventBusArn = bus.eventBusArn;
+
+    // Closure event bus base
+    const eventBusName = this.eventBusName;
+    this.native = <NativeIntegration<EventBusBase<E>>>{
+      bootstrap: (context: Function<any, any>) => {
+        this.bus.grantPutEventsTo(context.resource);
+      },
+      preWarm: (prewarmContext: NativePreWarmContext) => {
+        prewarmContext.eventBridge();
```
sam-goodwin, line 140: hmm, this looks interesting. Does NativePreWarmContext have all AWS SDK?

src/event-bridge/event-bus.ts
```diff
@@ -120,9 +124,37 @@ abstract class EventBusBase<E extends EventBusRuleInput>
...
+  readonly native: NativeIntegration<EventBusBase<E>>;
+
   constructor(readonly bus: aws_events.IEventBus) {
     this.eventBusName = bus.eventBusName;
     this.eventBusArn = bus.eventBusArn;
+
+    // Closure event bus base
+    const eventBusName = this.eventBusName;
+    this.native = <NativeIntegration<EventBusBase<E>>>{
+      bootstrap: (context: Function<any, any>) => {
```
sam-goodwin, line 136: should this be called `configure` or `synth`? `bootstrap` may be a bit vague/ambiguous. Although, funnily enough, it's exactly what I called it in punchcard haha

test/function.localstack.test.ts
thantos: prettier runs on save and on commit. Seems like it is fine with the format. 

image

SAVE

image


image


SAVE


image


image

SAVE

image

src/event-bridge/event-bus.ts
sam-goodwin: Any reason why we don't have `preWarm` instantiate any clients it needs and return them? `prewarmContext` can provide a cache which can be used to avoid duplicate clients?

test/function.localstack.test.ts
thantos: Is there something specific about the single form you don't like? Should the prettier rules be updated to reflect that?

src/function.ts
```diff
@@ -1,43 +1,94 @@
...
+        ).Payload?.toString();
+        return response ? JSON.parse(response) : undefined;
+      },
+      /**
+       * Code that runs once per lambda invocation
+       * The pre-warm client supports singleton invocation of clients (or other logic) across all integrations in the caller function.
+       */
+      preWarm: (preWarmContext: NativePreWarmContext) => {
+        preWarmContext.lambda();
+      },
```
sam-goodwin, line 89: nit: place above `call`

src/function.ts
```diff
@@ -1,43 +1,94 @@
...
+       */
+      bootstrap: (context: Function<any, any>) => {
+        this.resource.grantInvoke(context.resource);
+      },
+      /**
+       * This method is called from the calling runtime lambda code (context) to invoke this lambda function.
+       */
+      call: async (args, prewarmContext) => {
+        const [payload] = args;
+        const lambdaClient = prewarmContext.lambda();
```
sam-goodwin, line 71: does this prewarmContext properly tree-shake? If i only use `lambda()`, will it also serialize other client dependencies?

src/function.ts
```diff
@@ -61,28 +112,347 @@ export class Function<P, O> implements Integration {
...
+    // retrieve and bootstrap all found native integrations. Will fail if the integration does not support native integration.
+    const nativeIntegrationsPrewarm = integrations
+      .map((i) => new IntegrationImpl(i))
+      .map((integration) => {
+        const integ = integration.native;
+        integ.bootstrap(this);
+        return integ;
+      })
+      .filter((i) => i.preWarm)
+      .map((i) => i.preWarm);
```
sam-goodwin, line 233: nit: I personally think this can be simplified.

```ts
const nativeIntegrationsPrewarm = integrations.flatMap(i => {
  const integ = new IntegrationImpl(i).native;
  integ.bootstrap(this);
  return integ.preWarm ? [integ] : [];
});
```

src/function.ts
```diff
@@ -61,28 +112,347 @@ export class Function<P, O> implements Integration {
...
+      Function.promises.push(
+        callbackLambdaCode.generate(nativeIntegrationsPrewarm)
+      );
+  }
+
+  public static fromFunction<P, O>(
+    func: aws_lambda.IFunction
+  ): ImportedFunction<P, O> {
+    return new ImportedFunction<P, O>(func);
+  }
```
sam-goodwin, line 245: static functions above constructor

src/function.ts
```diff
@@ -61,28 +112,347 @@ export class Function<P, O> implements Integration {
...
+}
+
+export class ImportedFunction<P, O> extends FunctionBase<P, O> {
```
sam-goodwin, line 248: docs

package.json
thantos: Hmm, it was upgraded recently, downgraded during the merge from live, and restored on my commit from a while ago, running yarn default. Should be fine now.

src/function.ts
```diff
@@ -61,28 +112,347 @@ export class Function<P, O> implements Integration {
...
+  constructor(func: aws_lambda.IFunction) {
+    return super(func) as unknown as ImportedFunction<P, O>;
+  }
+}
+
+type ConditionalFunction<P, O> = P extends undefined
```
sam-goodwin, line 254: nice

src/function.ts
```diff
@@ -61,28 +112,347 @@ export class Function<P, O> implements Integration {
...
+    private func: (preWarmContext: NativePreWarmContext) => AnyFunction
+  ) {
+    super();
+  }
+
+  public bind(scope: Construct): aws_lambda.CodeConfig {
+    this.scope = scope;
+    // Lets give the function something lightweight while we process the closure.
+    return aws_lambda.Code.fromInline(
+      "If you are seeing this in your lambda code, ensure generate is called and then consult the README."
```
sam-goodwin, line 284: Can we link to an issue?

src/function.ts
```diff
@@ -61,28 +112,347 @@ export class Function<P, O> implements Integration {
...
+            const reversed =
+              Tokenization.reverse(obj, { failConcat: false }) ??
+              Tokenization.reverseString(obj).tokens;
+            if (!Array.isArray(reversed) || reversed.length > 0) {
+              if (Array.isArray(reversed)) {
+                tokens = [...tokens, ...reversed.map((s) => s.toString())];
+              } else {
+                tokens = [...tokens, reversed.toString()];
+              }
+            }
```
sam-goodwin, line 325: What's this do? Why do we need to reverse things?

src/function.ts
```diff
@@ -61,28 +112,347 @@ export class Function<P, O> implements Integration {
...
+          }
+          return true;
+        },
+      }
+    );
+
+    const tokenContext = tokens.map((t) => {
+      const id = new RegExp("\\${Token\\[TOKEN\\.([0-9]*)\\]}", "g").exec(
```
sam-goodwin, line 333: nit: why not use native syntax?
```
/\${Token\..etc.../g
```

src/function.ts
```diff
@@ -61,28 +112,347 @@ export class Function<P, O> implements Integration {
...
+      return {
+        id,
+        token: t,
+        // env variables must start with a alpha character
+        env: `env__functionless${id}`,
+      };
+    });
+
+    const resultText = tokenContext.reduce(
+      // TODO: support templated strings
```
sam-goodwin, line 345: i don't understand.

src/function.ts
```diff
@@ -61,28 +112,347 @@ export class Function<P, O> implements Integration {
...
+      (r, t) => r.split(`"${t.token}"`).join(`process.env.${t.env}`),
+      result.text
+    );
+
+    // console.log(resultText);
```
sam-goodwin, line 350: remove

src/function.ts
```diff
@@ -61,28 +112,347 @@ export class Function<P, O> implements Integration {
...
+
+    const asset = aws_lambda.Code.fromAsset("", {
```
sam-goodwin, line 352: does the empty string have any negative consequences?

src/function.ts
```diff
@@ -61,28 +112,347 @@ export class Function<P, O> implements Integration {
...
+      s3ObjectVersion: codeConfig.s3Location?.objectVersion,
+      zipFile: codeConfig.inlineCode,
+      imageUri: codeConfig.image?.imageUri,
+    };
+
+    asset.bindToResource(funcResource);
+  }
+}
+
+export interface NativeIntegration<F extends AnyFunction> {
```
sam-goodwin, line 393: docs

src/function.ts
```diff
@@ -61,28 +112,347 @@ export class Function<P, O> implements Integration {
...
+    args: Parameters<F>,
+    preWarmContext: NativePreWarmContext
+  ) => Promise<ReturnType<F>>;
+  /**
+   * Method called outside of the handler to initialize things like the PreWarmContext
+   */
+  preWarm?: (preWarmContext: NativePreWarmContext) => void;
+}
+
+/* istanbul ignore next */
```
sam-goodwin, line 410: why do i keep seeing this istanbul thing everywhere?

Also, this class seems important - let's add docs.

src/function.ts
```diff
@@ -61,28 +112,347 @@ export class Function<P, O> implements Integration {
...
+    this.cache = {};
+  }
+
+  public lambda(): Lambda {
+    return this.getOrInit("LAMBDA", () => new Lambda());
+  }
+
+  public eventBridge(): EventBridge {
+    return this.getOrInit("EVENT_BRIDGE", () => new EventBridge());
+  }
```
sam-goodwin, line 424: I like how this class caches things, but is it a good idea to embed the individual clients in here? Won't that harm tree-shaking?

src/function.ts
```diff
@@ -61,28 +112,347 @@ export class Function<P, O> implements Integration {
...
+      );
+    }
+    return this.getOrInit<T>(key, init);
+  }
+
+  public custom<T>(key: string): T {
+    return this.getOrFail<T>(key);
+  }
+
+  private getOrInit<T>(key: string, create: () => T): T {
```
sam-goodwin, line 443: Why not just make this public?

src/function.ts
```diff
@@ -61,28 +112,347 @@ export class Function<P, O> implements Integration {
...
+    init: () => T,
+    allowExists: boolean = false
+  ): T {
+    if (key in this.cache && !allowExists) {
+      throw Error(
+        `Prewarm Context key ${key} already exists. Use context.custom to get the object or set allowExists to true.`
+      );
+    }
+    return this.getOrInit<T>(key, init);
+  }
```
sam-goodwin, line 437: Is this over-engineering? Seems like `getOrInit` and `getOrFail` are enough?

test-app/src/message-board.ts
```diff
@@ -376,15 +371,11 @@ interface TestDeleteEvent
...
-`),
-    runtime: aws_lambda.Runtime.NODEJS_14_X,
-    handler: "index.handler",
-  })
+  stack,
+  "sendNotification",
+  async (event) => {
+    console.log("notification: ", event);
+  }
 );
```
sam-goodwin, line 379: syntax highlighting is broken by the string literal types in the Comment interface - consider moving them to the bottom to aid legibility in GitHub

test/function.localstack.test.ts
```diff
@@ -0,0 +1,332 @@
...
+    (parent) => {
+      const create = () => {
+        const val = "a";
+        return new Function(parent, "function", async () => val);
+      };
+
+      return { outputs: { function: create().resource.functionName } };
+    },
+    async (context) => {
+      await testFunction(context.function, {}, "a");
```
sam-goodwin, line 49: This seems boiler-platey and confusing. Can we simplify somehow?

```ts
const func = new Function(.., (param1, param2) => ..);

expect (await synth(func)(param1, param2)).toEqual(result);
```

test/function.localstack.test.ts
```diff
@@ -0,0 +1,332 @@
...
+      );
+    }
+  );
+
+  testResource(
+    "Call Lambda AWS SDK put event to bus with reference",
+    (parent) => {
+      const bus = new EventBus<any>(parent, "bus");
+
+      // Necessary to keep the bundle small and stop the test from failing.
```
sam-goodwin, line 191: How significantly does this effect the bundle size?

test/function.localstack.test.ts
```diff
@@ -0,0 +1,332 @@
...
+      // See https://github.com/sam-goodwin/functionless/pull/103#issuecomment-1116396779
```
sam-goodwin, line 192: Should we link to an issue instead of a comment?

test/function.localstack.test.ts
```diff
@@ -0,0 +1,332 @@
...
+      });
+
+      return { outputs: { function: func.resource.functionName } };
+    },
+    async (context) => {
+      await testFunction(context.function, {}, 0);
+    }
+  );
+
+  // See https://github.com/sam-goodwin/functionless/pull/103#issuecomment-1116396779
```
sam-goodwin, line 215: Issue instead of comment?

test/function.localstack.test.ts
```diff
@@ -0,0 +1,332 @@
...
+  testResource.skip(
+    "Call Lambda AWS SDK put event to bus without reference",
+    (parent) => {
+      const bus = new EventBus<EventBusRuleInput>(parent, "bus");
+
+      const func = new Function(parent, "function", async () => {
+        const result = $AWS.EventBridge.putEvents({
```
sam-goodwin, line 222: Is this really not supported?

sam-goodwin (changes requested):

test/function.localstack.test.ts
sam-goodwin: We are going to need to fork pulumi?

test/function.localstack.test.ts
sam-goodwin: No worries, just looked like it wasn't formatted because everything was on one line.

sam-goodwin:
How would someone write unit tests for their Lambda Functions that are coded this way? Do they have to mock much? Is there a pattern we can encourage?

thantos:
> How would someone write unit tests for their Lambda Functions that are coded this way? Do they have to mock much? Is there a pattern we can encourage?

Good question, right now there isn't an easy way. We could just expose the closure through the function and let them call it. Obviously it would have the problem that it assumes the serialization was lossless. Or we could serialize the function, import it again (or evaluate it) and then call the serialized from.


```ts
new Function(..., () => {}).invoke();
```

test/function.localstack.test.ts
thantos: It is supported, but it breaks in jest.

Essentially, pulumi serializes the entire `aws.js` file, which has `jest`'s injection. Then `global[ts-jest]` doesn't exist and fails in the lambda when its run on localstack.

Doing this outside of tests works, but does bloat the output bundle. #122 is supposed to reduce the change of this case.

test/function.localstack.test.ts
thantos: 3mb vs 100kb (including the serialized coverage code).

test/function.localstack.test.ts
thantos: Was trying to reduce the number of stack deployments, it also supports automatic creation of outputs and passes them back to the test function. Deployments to `localstack` are not fast and would require either a unique state, or for the same stack to incrementally deploy during each test. 

The `outputs: { function: ... }` part is boilerplaty, but this wrapper supports any resource (there is example that uses the bus arns below). 

I'll see if I can make a special wrapper for the lambda tests to make them smaller.

What does `synth` do here?

test/function.localstack.test.ts
thantos: Maybe? I wanted to build up more test cases and use that to figure out what it would take to build from scratch.

src/aws.ts
thantos: Trying to reduce the code coverage code injected into the function closures during tests, but it has not been successful. I'll probably just remove those. The code coverage code makes up a majority of the serialized functions. Mostly just a problem for us. I also though they may be contributing to issues I was having, but I don't think they are. Just another example of where we need control.

src/compile.ts
thantos: Done

src/compile.ts
thantos: That is what that code did below, but I was having a hard time determining if something was a Contract or not.

src/compile.ts
thantos: The compiler supports returning an array of nodes, but only when the parent supports multiple nodes. I needed to inject siblings into the context. If the child returns multiple nodes and the parent cannot handle multiple nodes (ex: `VariableAssignment`), they get passed up the tree until the parent can support multiple nodes (ex: `ExpressionStatement`)

For example, if I wanted to create a client that a line needs, I wanted to return the client along with the transformed node.

```ts
bus()
```

Compiled to

```ts
const EB = new EventBridge()
await EB.putTargets()
```

Howerever... the one use case I had for this is now supported using the prewarm context, so I don't need this anymore!

src/compile.ts
thantos: See above, will remove anyways, but it was fun... is there a better way to do this?
https://github.com/sam-goodwin/functionless/pull/103#discussion_r867284322

src/compile.ts
sam-goodwin: Could we use structural typing heuristics? Or the fully qualified name of the type?

src/compile.ts
thantos:
```ts
              // integration variables can be CDK constructs which will fail serialization.
              // TODO: can we relax this be determining if a integration reference will fail?
```

src/compile.ts
thantos: Fully qualified in typescript is the path, which is different on every system, right? I wasn't able to figure out how to make good use of that. Have any examples?

thantos: Not sure why I didn't notice this before, will give it a try.

https://github.com/aws/aws-cdk/blob/master/packages/@aws-cdk/core/lib/construct-compat.ts#L21

thantos: Got it...


```ts
      /**
       * Heuristically evaluate the fqn of a symbol to be in a module and of a type name.
       *
       * /somePath/node_modules/{module}/somePath.{typeName}
       *
       * isInstanceOf(typeSymbol, "constructs", "Construct")
       * ex: /home/sussmans/functionless/node_modules/constructs/lib/index.js
       */
      function isInstanceOf(
        symbol: ts.Symbol,
        module: string,
        typeName: string
      ) {
        const find = /.*\/node_modules\/([^\/]*)\/.*\.(.*)$/g.exec(
          checker.getFullyQualifiedName(symbol)
        );

        const [_, mod, type] = find ?? [];

        return mod === module && type === typeName;
      }

      function isCDKConstruct(type: ts.Type): boolean {
        const typeSymbol = type.getSymbol();

        return (
          ((typeSymbol &&
            isInstanceOf(typeSymbol, "constructs", "Construct")) ||
            type.getBaseTypes()?.some((t) => isCDKConstruct(t))) ??
          false
        );
      }
```

src/function.ts
thantos: It does not! Been looking at that too.

Technically any of the keys could be requested at runtime event if they are not preWarmed. (and new entries can be added.)

I'll add a task to dynamically build the contents of the PreWarmContext instead of using an object.

Because the `aws_sdk` import is used, the impact of the extra methods is pretty minimal.

```js
function __f4() {
  return (function() {
    with({ aws_sdk_1: require("aws-sdk/lib/aws.js") }) {

return function /*lambda*/() {
    return this.getOrInit("LAMBDA", () => {
      var _a;

      return new aws_sdk_1.Lambda((_a = this.clientConfigRetriever) === null || _a === void 0 ? void 0 : _a.call(this, "LAMBDA"));
    });
  };

    }
  }).apply(undefined, undefined).apply(this, arguments);
}

function __f5() {
  return (function() {
    with({ aws_sdk_1: require("aws-sdk/lib/aws.js") }) {

return function /*eventBridge*/() {
    return this.getOrInit("EVENT_BRIDGE", () => {
      var _a;

      return new aws_sdk_1.EventBridge((_a = this.clientConfigRetriever) === null || _a === void 0 ? void 0 : _a.call(this, "EVENT_BRIDGE"));
    });
  };

    }
  }).apply(undefined, undefined).apply(this, arguments);
}
...
var __preWarmContext_proto = {};
Object.defineProperty(__f3, "prototype", { value: __preWarmContext_proto });
Object.defineProperty(__preWarmContext_proto, "constructor", { configurable: true, writable: true, value: __f3 });
Object.defineProperty(__preWarmContext_proto, "lambda", { configurable: true, writable: true, value: __f4 });
Object.defineProperty(__preWarmContext_proto, "eventBridge", { configurable: true, writable: true, value: __f5 });
Object.defineProperty(__preWarmContext_proto, "registerCustom", { configurable: true, writable: true, value: __f6 });
Object.defineProperty(__preWarmContext_proto, "custom", { configurable: true, writable: true, value: __f7 });
Object.defineProperty(__preWarmContext_proto, "getOrInit", { configurable: true, writable: true, value: __f8 });
Object.defineProperty(__preWarmContext_proto, "getOrFail", { configurable: true, writable: true, value: __f9 });
```

src/function.ts
thantos: nice :+1:

src/function.ts
thantos: https://github.com/sam-goodwin/functionless/issues/128

src/function.ts
thantos: reverse ~= `de-encode`, and yeah, it helps us find real tokens.

src/function.ts
thantos: We can, originally I was using variables, I guess I am no longer.

src/function.ts
thantos: I had assumed that `"${TOKEN[TOKEN_123]} and ${TOKEN[TOKEN_123]}"` would not work, but it appears like it does because the actual references are ripped out.

Before: `var __bus = {eventBusArn: "${TOKEN[TOKEN_439]}"};`
After: `var __bus = {eventBusArn: process.env.env__functionless439};`

So a templated string would look like:

`${_bus.eventBusArn} and ${_bus.eventBusArn}`

The problem is, what if a token is ever not in a standalone string?

`"${TOKEN[TOKEN_123]} stuff"`

Now the replace will not work. Need to figure out the best way to test this.

src/function.ts
thantos: Only on hashing, but adding `user: ...` fixes that. Its stupid the input is required when the bundler can reference any path.

src/function.ts
thantos: Good question, answered above that the impact is low due to the use of `require("aws_sdk")`, but I would also like to reduce the size of the outputted code in any way...

The problem I had was... any integration may need any client, and I wouldn't want a client to add it's own flavor behind a shared key like `this.getOrInit("LAMBDA", () => new SpecialLambdaClient())`.

I guess we could make constants that encode the shared logic.

```ts
prewarmClient.get(PrewarmClient.LAMBDA);

// I am not sure how well an enum would serialize either
enum PrewarmClient {
   LAMBDA: () => new Lambda();
}
```

Will play with it.

src/function.ts
thantos: I probably will now after the last comment. Original thought was that shared keys should use the known helpers and the custom keys should try to use `register` in the preWarm and `get` in the runtime.

src/function.ts
thantos: Maybe! I liked being able to assert that a key already existed, but I guess we could just add a `get: () => T | null`.

src/function.ts
thantos: Removed them all, just means we might have slightly bigger bundles while testing and that is fine.

src/function.ts
thantos: Got it

```ts
      preWarm: (preWarmContext: NativePreWarmContext) => {
        preWarmContext.getOrInit(PrewarmClients.LAMBDA);
      },
```

Serialized, only the required methods: 

```ts
var __exports_PrewarmClients = {};
var __exports_PrewarmClients_LAMBDA = {key: "LAMBDA", init: __f2};
__exports_PrewarmClients.LAMBDA = __exports_PrewarmClients_LAMBDA;
__exports.PrewarmClients = __exports_PrewarmClients;
```

test/function.localstack.test.ts
thantos: refactored

```ts
testFunctionResource("name", () => new Function(...), expected, payload?);
```

src/event-bridge/event-bus.ts
thantos: thats what it does, I've just added helpers for common clients. you can also add whatever.

```ts
preWarmConext.getOrInit({  key: "anything", init: () => "hi" });
preWarmConext.get("anything") // "hi"

preWarmConext.getOrInit(PrewarmClients.LAMBDA);
preWarmConext.get(PrewarmClients.LAMBDA.key) // lambda client
```

thantos: the helpers correctly apply any props and ensure that all consumers use the same implementation.

```ts
  LAMBDA: {
    key: "LAMBDA",
    init: (key, props) => new Lambda(props?.clientConfigRetriever?.(key)),
  },
```

src/event-bridge/event-bus.ts
thantos: `bind`?

src/function.ts
thantos: Fixed this, new API doesn't serialize unless used.

src/function.ts
thantos: I have not found a test that actually fails yet, so removed the comment.

src/event-bridge/event-bus.ts
sam-goodwin: Like it.

src/function.ts
```diff
@@ -1,43 +1,94 @@
...
- *   new aws_lambda.Function(..)
- * );
- *
- * new AppsyncResolver(() => {
- *   return getPerson("value");
- * })
- * ```
- */
-export class Function<P, O> implements Integration {
+export type FunctionClosure<P, O> = (payload: P) => Promise<O>;
```
sam-goodwin, line 29: We should pass the Lambda Context from `@types/aws-lambda` as the second argument

sam-goodwin (changes requested):

sam-goodwin:
Is the following supported?

```ts
let _allowList;
const loadAllowList = async () => {
  return (_allowList =
    _allowList ?? (await fs.promises.readFile("allow-list.json")));
};

new Function(scope, "foo", async () => {
  // reference the allowList here instead
  const list = await loadAllowList();
});
```

Will the `allow-list.json` file be bundled? Probably not right?

thantos:
Ran into issues that were difficult to solve with the current state of Pulumi's serializer. Working on a fork of pulumi:

Added features so far:

* Symbol support - Symbols cause an error when serialized - https://github.com/pulumi/pulumi/issues/6229#issuecomment-1135118712
* Ability to update the object being serialized (see below)

Updating Serialized objects:

Current Pulumi supports a `serialize` callback which allows consumers to ignore objects during serialization. The reality of this is entire objects can be ignored (turned into `undefined`), but if the object is partially needed by the runtime, then its all or nothing.

For example, a CDK Construct like a `Table` can be discovered and ignored:

```ts
if(obj instanceof aws_dynamo.Table) {
   return false;
}
```

But this fails when we need parts of the table

```ts
new Function(async () => {
    const func = (table) => table.tableName;
    console.log(func(table));
})
```

In the above case, returning false would cause `table` to be undefined and returning true would cause the entire CDK `Table` to be serialized, generally failing when a `Lazy` is found in the CDK object. Serializing the CDK construct also causes a huge bloat to the closure that is generated. I've seen simple code generate a 20MB bundle.

There are ways to work around the issue above, for example setting the table name outside the closure to avoid the entire table being serialized.

```ts
const tableName = table.tableName;
new Function(async () => {
    const func = (tableName) => tableName; // not quite the same...
    console.log(func(tableName));
})
```

While generally a nuisance that can be worked around, this created an issue when trying integrate dynamo with the `$AWS.DynamoDB.*` operations. 

```ts
const table = new Table();
new Function(async () => {
    return $AWS.DynamoDB.getItem({ TableName: table, ... }); // will also serialize the entire table.
})
```

Working on a base assumption that there are parts of the CDK object we should not need (and if we do, we should add escape hatches to unblock), we are adding a new feature to our pulumi serializer fork: The ability to override the object being serialized (instead of ignoring it).

1. Updated the `serialize` contract to accept `boolean | any`: (o: any) => boolean | any;
2. When a non-boolean truthy value is returned, use that value instead.

Now we can do

```ts
if (Resource.isResource(o as any)) {
    const { node, stack, env, ...rest } = o as unknown as Resource;
    return rest;
}
```

This block removes node, stack, and env from the object to be serialized, but leaves the rest intact.

What we are modifying so far:
* CDK Constructs/Resources - stack, node, env are not needed at least
* Cfn* L1 resources - stack, node, creationStack, _toCloudformation, and cfnProperties are not needed.
* CDK ITaggable - `tags` on constructs can easily bring in the entire CDK TagManager which appears to have minimal values at tuneimt.
* Functionless Integrations - Functionless integrations (which can also be or contain CDK constructs) can contain integrations for any of the functionless resources (step functions, app sync, etc), but we only need the `native` functions, in fact, we only need `preWarm` and `call` as bind is used during `synth`.
* Table - ?

sam-goodwin:
> Updated the serialize contract to accept boolean | any: (o: any) => boolean | any;

This has an edge case - what happens when the actual value (we want to keep) is a literal false boolean?

thantos:
> > Updated the serialize contract to accept boolean | any: (o: any) => boolean | any;
> 
> This has an edge case - what happens when the actual value (we want to keep) is a literal false boolean?

I don't like the current contract either, the current form was the fast to create to experiment.... however... you example isn't a problem as we don't overwrite then the value is a boolean

https://github.com/functionless/nodejs-closure-serializer/blob/main/src/closure/createClosure.ts#L839

The real edge case is if I wanted to update any value to a boolean (which I figured was a rare case for now...). 

* false - updates to undefined
* true - leaves as is
* anything else - updates to that value

Proposed update: https://github.com/functionless/nodejs-closure-serializer/issues/1

src/declaration.ts
thantos: Current implementation extracts the integrations during the native expr parsing. 

I question the need to output the entire `FunctionDecl` and risking any compilation failures when we don't make use of most of the output. 

Going to leave this as is for now.

thantos:
From review today: Moved the functionProps to an optional 3rd param instead of last in `Function`

src/function.ts
```diff
@@ -61,28 +125,539 @@ export class Function<P, O> implements Integration {
...
+              }
+            }
+          } else if (typeof obj === "object") {
+            /**
+             * Remove unnecessary fields from {@link CfnResource} that bloat or fail the closure serialization.
+             */
+            const transformCfnResource = (o: unknown): any => {
+              if (Resource.isResource(o as any)) {
+                const { node, stack, env, ...rest } = o as unknown as Resource;
+                return rest;
```
sam-goodwin, line 379: cute! does this capture methods from the prototype?

sam-goodwin: How big are the bundles we're getting?

src/function.ts
```diff
@@ -61,28 +125,539 @@ export class Function<P, O> implements Integration {
...
+                  _toCloudFormation,
+                  // @ts-ignore - private - adds the tag manager, which we don't need
+                  cfnProperties,
+                  ...rest
+                } = transformTable(o as CfnResource);
+                return transformTaggableResource(rest);
+              } else if (Token.isUnresolved(o)) {
+                const token = (<any>o).toString();
+                // add to tokens to be turned into env variables.
+                tokens = [...tokens, token];
```
sam-goodwin, line 395: why not push?

src/function.ts
```diff
@@ -61,28 +125,539 @@ export class Function<P, O> implements Integration {
...
+              if (
+                o.cfnResourceType ===
+                aws_dynamodb.CfnTable.CFN_RESOURCE_TYPE_NAME
+              ) {
+                const table = o as aws_dynamodb.CfnTable;
+                if (!table.streamSpecification) {
+                  const { attrStreamArn, ...rest } = table;
+
+                  return rest as unknown as CfnResource;
+                }
```
sam-goodwin, line 414: What's the logic here? Seems like we're keeping a whole bunch

src/function.ts
```diff
@@ -61,28 +125,539 @@ export class Function<P, O> implements Integration {
...
+            /**
+             * Remove unnecessary fields from {@link CfnTable} that bloat or fail the closure serialization.
+             */
+            const transformIntegration = (o: unknown): any => {
+              if (o && typeof o === "object" && "kind" in o) {
+                const integ = o as Integration;
+                const {
+                  appSyncVtl,
+                  asl,
+                  native: { call, preWarm } = {},
```
sam-goodwin, line 440: We have to remember to update this every time we add a new integration type. That's probably fine because it won't happen often, right?

src/function.ts
```diff
@@ -61,28 +125,539 @@ export class Function<P, O> implements Integration {
...
+
+    // // find any strings that contain tokens
+    // const stringsWithQuotes = /["'`].*\${Token\[.*\.[0-9]*\]}.*['`"]/g.exec(
+    //   resultText
+    // );
+
+    // stringsWithQuotes?.reduce(
+    //   (r, s) => r.replace(s, s.replace(/(^['"]|['"]$)/g, "`")),
+    //   resultText
+    // );
```
sam-goodwin, line 484: nit: remove?

src/function.ts
```diff
@@ -61,28 +125,539 @@ export class Function<P, O> implements Integration {
...
+ *   key: 'customClient',
+ *   init: () => new anyClient()
+ * })
+ * ```
+ */
+export const PrewarmClients = {
+  LAMBDA: {
+    key: "LAMBDA",
+    init: (key, props) =>
+      new (require("aws-sdk").Lambda)(props?.clientConfigRetriever?.(key)),
```
sam-goodwin, line 595: `aws-sdk` is ideal for bundling because it's included in the lambda function by default. Do we know if it has any performance implications though? Does V2 use the optimized C code that v3 does?

test-app/src/message-board.ts
```diff
@@ -425,6 +398,88 @@ customDeleteBus
...
+        id: "from the test method!!",
+      },
+    });
+    // busbus({
+    //   "detail-type": "Delete-Post-Success",
+    //   source: "MessageDeleter",
+    //   detail: {
+    //     id: "from the test method!!",
+    //   },
+    // });
```
sam-goodwin, line 438: remove

test-app/src/message-board.ts
```diff
@@ -425,6 +398,88 @@ customDeleteBus
...
+new Function(
+  stack,
+  "testFunc",
+  async () => {
+    console.log(customDeleteBus.eventBusArn);
+    console.log(busbusbus.eventBusArn);
+    console.log("huh?!?!?!??!!");
+    console.log("strange");
+    const result = func();
+    console.log(`function result: ${result}`);
```
sam-goodwin, line 424: remove logging?

test/function.localstack.test.ts
```diff
@@ -0,0 +1,533 @@
...
+            function: funcRes.resource.functionName,
+            ...outputs,
+          },
+        };
+      },
+      async (context) => {
+        const exp =
+          // @ts-ignore
+          typeof expected === "function" ? expected(context) : expected;
+        // @ts-ignore
```
sam-goodwin, line 76: why ignore?

test/function.localstack.test.ts
```diff
@@ -0,0 +1,533 @@
...
+    "Call Lambda from closure with parameter multiple 1",
+    (parent) => create(parent, "func6", "c"),
+    "c"
+  );
+
+  testFunctionResource(
+    "Call Lambda from closure with parameter multiple 2",
+    (parent) => create(parent, "func7", "d"),
+    "d"
+  );
```
sam-goodwin, line 148: multiple 1 and 2. What does that mean?

test/function.localstack.test.ts
```diff
@@ -0,0 +1,533 @@
...
+
+  testFunctionResource(
+    "Call Lambda payload",
+    (parent) =>
+      new Function(parent, "function", async (event: { val: string }) => {
+        return `value: ${event.val}`;
+      }),
+    "value: hi",
+    { val: "hi" }
```
sam-goodwin, line 181: so just so i understand, the contract is:
1. create the Function construct
2. expected output
3. input argument

sam-goodwin: For users, we're going to need this to be:
```ts
test("blah", async () => {
  const func = new Function(..);
  expect(await func(input)).toEqual(expected);
});
```

test/function.localstack.test.ts
sam-goodwin: Can we fix this now that we have control over the closure serializer?

test/function.localstack.test.ts
```diff
@@ -0,0 +1,533 @@
...
+      );
+
+      bus.bus.grantPutEventsTo(func.resource);
+
+      return func;
+    },
+    0
+  );
+
+  // See https://github.com/sam-goodwin/functionless/pull/122
```
sam-goodwin, line 299: fix link

sam-goodwin (changes requested):

.vscode/launch.json
```diff
@@ -1,6 +1,21 @@
...
+        "target": "code",
+        "projectRoot": "${workspaceFolder}/message-board-testFuncB346065B-zllkEyeNxDAI",
+        "lambdaHandler": "index.handler"
+      },
+      "lambda": {
+        "runtime": "nodejs14.x",
+        "payload": {},
+        "environmentVariables": {}
+      }
+    },
```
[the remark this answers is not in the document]
thantos: Probably not

src/aws.ts
```diff
@@ -361,4 +601,37 @@ export namespace $AWS {
...
+     * @see https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_PutEvents.html
+     */
+    export const putEvents = makeIntegration<
+      (
+        request: AWS.EventBridge.Types.PutEventsRequest
+      ) => AWS.EventBridge.Types.PutEventsResponse,
+      "EventBridge.putEvent"
+    >({
+      kind: "EventBridge.putEvent",
+      native: {
```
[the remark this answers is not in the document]
thantos: Not sure what you mean?

src/compile.ts
```diff
@@ -269,6 +287,54 @@ export function compile(
...
+      ): node is FunctionInterface {
+        return (
+          ts.isNewExpression(node) &&
+          isFunctionlessClassOfKind(
+            node.expression,
+            Function.FunctionlessType
+          ) &&
+          // only take the form with the arrow function at the end.
+          (node.arguments?.length === 3 || node.arguments?.length === 4) &&
+          ts.isArrowFunction(node.arguments[2])
```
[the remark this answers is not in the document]
thantos: Yes. WIP.

src/aws.ts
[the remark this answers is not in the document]
sam-goodwin: Can we code generate `native` integrations for the entire AWS SDK?

src/compile.ts
```diff
@@ -269,6 +287,54 @@ export function compile(
...
+        );
+      }
+
+      /**
+       * Heuristically evaluate the fqn of a symbol to be in a module and of a type name.
+       *
+       * /somePath/node_modules/{module}/somePath.{typeName}
```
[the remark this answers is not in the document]
thantos: Any other thoughts for handling this heuristically?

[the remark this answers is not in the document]
sam-goodwin: Structural duck typing. Check for the existence of properties.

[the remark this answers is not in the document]
sam-goodwin: See https://github.com/functionless/functionless/pull/103#discussion_r884030512

src/compile.ts
```diff
@@ -856,6 +1175,167 @@ export function compile(
...
+       *    const z = { y: x };
+       *    z.y()
+       * }
+       * ```
+       *
+       * getOutOfScopeValueNode(z.y) => x
+       *
+       * The call to busbus can be resolved to bus if the scope is the array function.
+       */
+      function getOutOfScopeValueNode(
```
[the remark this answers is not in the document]
thantos: Probably cannot be found. Have not exhaustively tested external reference patterns. Good test to add.

src/compile.ts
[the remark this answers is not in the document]
thantos: What property would you look for on the base CDK construct?

src/compile.ts
```diff
@@ -407,6 +473,261 @@ export function compile(
...
+          // a reference to a prewarm context that will be passed into the closure during synthesis/runtime
+          preWarmContext,
+          // the closure node used to determine if variables are inside or outside of the closure
+          closureNode: impl,
+          // pass up integrations from inside of the closure
+          registerIntegration: (integ, args) =>
+            integrations.push({ expr: integ, args }),
+        };
+
+        const body = toNativeExpr(
```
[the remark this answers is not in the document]
thantos: Yeah

src/compile.ts
```diff
@@ -407,6 +473,261 @@ export function compile(
...
+              context.factory.createArrayLiteralExpression(
+                node.arguments.map((arg) => {
+                  return toExpr(arg, nativeExprContext.closureNode);
+                })
+              )
+            );
+            // call the integration call function with the prewarm context and arguments
+            // At this point, we know native will not be undefined
+            // await integration.native.call(args, preWarmContext)
+            // FIXME: doesn't work without an array
```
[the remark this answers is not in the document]
thantos: Woops. Missed removing this.

src/event-bridge/event-bus.ts
```diff
@@ -120,9 +129,39 @@ abstract class EventBusBase<E extends EventBusRuleInput>
...
+        );
+        await eventBridge
+          .putEvents({
+            Entries: args.map((event) => ({
+              Detail: JSON.stringify(event.detail),
+              EventBusName: eventBusName,
+              DetailType: event["detail-type"],
+              Resources: event.resources,
+              Source: event.source,
+              Time: event.time ? new Date(event.time) : undefined,
```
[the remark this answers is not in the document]
thantos: Think the timestamp will ever be 0?

src/compile.ts
```diff
@@ -269,6 +287,54 @@ export function compile(
...
+      function isCDKConstruct(type: ts.Type): boolean {
+        const typeSymbol = type.getSymbol();
+
+        return (
+          ((typeSymbol &&
+            isInstanceOf(typeSymbol, "constructs", "Construct")) ||
+            type.getBaseTypes()?.some((t) => isCDKConstruct(t))) ??
+          false
+        );
+      }
```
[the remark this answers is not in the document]
sam-goodwin: I would check for the node property and then check for many of the properties on the node class.

src/compile.ts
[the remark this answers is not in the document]
sam-goodwin: I would check for the node property and then check for many of the properties on the node class.

src/compile.ts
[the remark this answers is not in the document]
sam-goodwin: Let's open an issue or track in a test?

src/event-bridge/event-bus.ts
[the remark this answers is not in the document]
sam-goodwin: Not sure, but it is a valid timestamp. Easy enough to check:

```ts
Time: typeof event.time === "number" ? new Date(event.time) : undefined
```

src/aws.ts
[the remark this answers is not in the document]
thantos: ohh, yeah, we could. https://github.com/functionless/functionless/issues/168

src/compile.ts
```diff
@@ -45,6 +46,12 @@ export function compile(
   return (ctx) => {
     const functionless = ts.factory.createUniqueName("functionless");
     return (sf) => {
+      // Do not transform any of the files matched by "exclude"
+      if (excludeMatchers.some((matcher) => matcher.test(sf.fileName))) {
+        return sf;
+      }
+
+      // TODO: this seems like a pattern, make a "public import context"
```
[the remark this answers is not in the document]
thantos: I changed compile to only include the functionless import when its used, instead of all of the time. 

Originally thought we'd have multiple dynamic imports, but ended up removing the other one. Removed the comment.

src/compile.ts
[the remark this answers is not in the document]
thantos: Actually, this pattern works for both nested and top level paths (though the comment doesn't mention that).

All it does is check to see if the npm package name and the type name are equal.

Both will return true

`/home/sussmans/functionless/node_modules/constructs/lib/construct.Construct`
and 
`/home/sussmans/functionless/node_modules/aws-cdk-lib/node_modules/constructs/lib/construct.Construct`

I think this will work for moth situations.

src/compile.ts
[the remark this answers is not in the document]
thantos: Answered here: https://github.com/functionless/functionless/pull/103#discussion_r884063485

I think this will work. The chance that a package name and type name collide in the same code base is low.

src/compile.ts
```diff
@@ -407,6 +473,261 @@ export function compile(
...
+       * ```ts
+       * const bus = new EventBus()
+       * new NativeFunctionDecl(
+       *     [new ParameterDecl("arg1")], // parameters
+       *     (prewarmContext: NativePreWarmContext) =>
+       *        (arg1: string) => {
+       *            // call can make use of the cached clients in prewarmContext to avoid duplicate inline effect
+       *            await bus.native.call(prewarmContext, { source: "src" });
+       *        },
+       *     [bus] // integrations
```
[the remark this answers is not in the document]
thantos: `table.get`

The entire expression on the call is passed up or corrected to be the out of scope form.

```ts
const table = new Table();
new Function(async () => {
   const t  = table;
   const get = t.get;
   get(...);
})
```

becomes

`[table.get]`

src/compile.ts
```diff
@@ -407,6 +473,261 @@ export function compile(
...
+          // Integration nodes have a static "kind" property.
+          if (isIntegrationNode(node.expression)) {
+            const outOfScopeIntegrationReference = getOutOfScopeValueNode(
+              node.expression,
+              nativeExprContext.closureNode
+            );
+
+            if (!outOfScopeIntegrationReference) {
+              // integration variables can be CDK constructs which will fail serialization.
+              // TODO: can we relax this be determining if a integration reference will fail?
```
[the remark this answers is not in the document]
thantos: Removed it

src/compile.ts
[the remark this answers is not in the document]
sam-goodwin: Cool, sounds good.

src/compile.ts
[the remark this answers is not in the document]
sam-goodwin: That is awesome. Great job

src/compile.ts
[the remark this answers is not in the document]
thantos: got it :-) (commit soon)

```ts
const b = {
    bus: new EventBus(...)
}
new Function(... () => {
    const { bus } = b;
    bus(...);
})
```

```ts
{ integration: b["bus"] }
```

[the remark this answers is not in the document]
thantos: summary: created a method to discover and flatten destructured assignment.

```ts
      function flattenDestructuredAssignment(
        element: ts.BindingElement
      ): ts.ElementAccessExpression {
        // if the binding renames the property, get the original
        // { a : x } -> a
        // { a } -> a
        // [a] -> 0
        const name = ts.isArrayBindingPattern(element.parent)
          ? element.pos
          : // binding renames the property or is a nested binding pattern.
          element.propertyName
          ? element.propertyName
          : // the "name" can be a binding pattern. In that case the propertyName will be set.
            (element.name as ts.Identifier);

        const getParent = () => {
          // { a } = b;
          if (ts.isVariableDeclaration(element.parent.parent)) {
            if (!element.parent.parent.initializer) {
              throw Error(
                "Expected a initializer on a destructured assignment: " +
                  element.getText()
              );
            }
            return element.parent.parent.initializer;
          } else if (ts.isBindingElement(element.parent.parent)) {
            return flattenDestructuredAssignment(element.parent.parent);
          } else {
            throw Error(
              "Cannot flatten destructured parameter: " + element.getText()
            );
          }
        };

        const parent = getParent();

        // always use element access as this will work for all possible values.
        // [parent][name]
        return ts.factory.createElementAccessExpression(
          parent,
          typeof name !== "number"
            ? ts.isComputedPropertyName(name)
              ? name.expression
              : ts.isIdentifier(name)
              ? // { a } = b -> b["a"]
                ts.factory.createStringLiteral(name.text)
              : name
            : // numeric index position for a destructured array
              name
        );
      }
```

sam-goodwin (approved):

src/function.ts
```diff
@@ -1,43 +1,107 @@
...
 
   // @ts-ignore - this makes `F` easily available at compile time
   readonly __functionBrand: ConditionalFunction<P, O>;
 
   constructor(readonly resource: aws_lambda.IFunction) {
-    // Integration object for appsync VTL
+    const functionName = this.resource.functionName;
+
+    // Native is used when this function is called from another lambda function's serialized closure.
+    // define this here to closure the functionName without using `this`
```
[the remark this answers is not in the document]
thantos: Trying this... maybe we need a lazy?

```ts
let func1: Function<string, string> | undefined;
const func1Lazy = Lazy.integ(() => func1);
const func2 = new Function(async () => {
   func1Lazy.get()("hello")
})
func1 = new Function<string, string>(async () =>{ func2() });
```

This won't work right now because we invoke the integration in the constructor of the Function its used in. 

Or maybe a ref()

```ts
const func1Ref = Function.ref<string, string>();
const funct2 = new Function(async () => {
   func1Ref("");
})
const func1 = func1Ref.set(new Function(async () => { func2() }));
```

Right now... this would just mean calling the bind function and re-writing the environment variables, I think...

https://github.com/functionless/functionless/issues/173

src/function.ts
```diff
@@ -61,28 +125,539 @@ export class Function<P, O> implements Integration {
...
+        return integ.preWarm ? [integ.preWarm] : [];
+      }
+    );
+
+    // Start serializing process, add the callback to the promises so we can later ensure completion
+    callbackLambdaCode &&
+      Function.promises.push(
+        callbackLambdaCode.generate(nativeIntegrationsPrewarm)
+      );
+  }
```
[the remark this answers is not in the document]
thantos: FINE :-P

src/function.ts
thantos: that large lambda I have in message-board is 200kb. See attached.
[index.txt](https://github.com/functionless/functionless/files/8791814/index.txt)

src/function.ts
thantos: I don't like using the mutation functions on arrays if avoidable.

src/function.ts
thantos: the table is also going through the CfnResource filtering transform, this just adds the safety for the streamArm.

thantos: Updated the docs.

src/function.ts
thantos: We don't NEED to, but new integration types may bloat the closures. I'll make this more dynamic...

thantos: Some typesafe runtime enforcement

 ```ts
/**
 * Maintain a typesafe runtime map of integration type keys to use elsewhere.
 * 
 * For example, removing all but native integration from the {@link Function} closure.
 */
const INTEGRATION_TYPES: Record<keyof IntegrationMethods<any>, any> = {
  appSyncVtl: true,
  asl: true,
  native: true
} 

export const INTEGRATION_TYPE_KEYS = Object.keys(INTEGRATION_TYPES) as (keyof IntegrationMethods<any>)[];
```

Comment (thantos) [reply src/function.ts:459]: slightly better

```ts
/**
 * Maintain a typesafe runtime map of integration type keys to use elsewhere.
 *
 * For example, removing all but native integration from the {@link Function} closure.
 */
const INTEGRATION_TYPES: { [P in keyof IntegrationMethods<any>]: P } = {
  appSyncVtl: "appSyncVtl",
  asl: "asl",
  native: "native",
};

export const INTEGRATION_TYPE_KEYS = Object.values(INTEGRATION_TYPES);
```

src/function.ts
thantos: There may be some performance improvements using v3 in a function: https://aws.amazon.com/blogs/developer/modular-packages-in-aws-sdk-for-javascript/

https://github.com/functionless/functionless/issues/174

test/function.localstack.test.ts
thantos: Laziness mostly

image

test/function.localstack.test.ts
thantos: #131 covers running the closure of integrations as typescript.

The tests here look funky because of the integration with localstack and actually running the lambda on localstack. This is how we'd test any CDK resource, not specific to our function.

Did you have an idea how testing would be done by developer? I don't think we'd want to integrate localstack in official, so its up to the developer how they'd test resources running again AWS or localstack (which look like any other CDK resource).

I guess #131 is the closest to what developers can do today, just running their lambda code, but without the step of serialization and integration being the unknown variables. 

Another option is to expose the serialized form somehow to run locally to catch any issues... not sure yet.

test/function.localstack.test.ts
thantos: Fixed all of the previous links in #175.

Fixed all of the links from this change in the next commit.

thantos:
> Can we fix this now that we have control over the closure serializer?

No idea why github isn't showing the repy box...

This case works for users now (message-board has an example).

But it doesn't work in our tests because all of $AWS is being embedded which has been including the `globals["ts-jest"]` reference which when breaks all tests. May be able to play with `serialize` function to change the behavior.

src/compile.ts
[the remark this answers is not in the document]
sam-goodwin: So the synth-time interpreters will never have to handle restructuring? Is there any chance of name conflicts after flattening?

src/function.ts
sam-goodwin: But you're setting a variable. So still mutational at the cost of o(n) time

src/compile.ts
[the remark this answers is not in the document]
thantos: We are not always flattening. I am only doing it for getting the "out of scope" version of the integration to be able to pass back into the function. In this case at least, the names won't collide.

sam-goodwin (approved):

PR #104, opened 2022-04-22
feat: implement a language service plugin to error on unsupported syntax

### LINKED ISSUE #10: Emit diagnostics for semantic validations.
Implement a TypeScript Language Service Plugin which adds domain-specific diagnostics.

https://github.com/microsoft/TypeScript/wiki/Writing-a-Language-Service-Plugin

sam-goodwin:
Fixes #10 

This change solves two problems:
- [x] Expose Diagnostics in the IDE through a Language Service Plugin
- [x] Fail the build if there are any error. Need to figure out if this is with eslint or another plugin.

We now have a `validate.ts` file that centralizes validation logic for use within https://github.com/functionless/functionless-language-service and the `tsc.ts` CLI.

All compile-time query functions from `compile.ts` are separated out into `checker.ts` to also centralize those since they are used in both the transform and the validation code.

We also now have a `functionless` CLI that performs semantic validation so that builds can fail when unsupported syntax is detected. The language service plugin only works for the IDE :(


src/validate.ts
```diff
@@ -0,0 +1,42 @@
...
+      // assignment statement
+      // TODO: check if the assignment references an expression in an illegal scope
+    }
+
+    return descend();
+
+    function descend() {
+      return ts.forEachChild(node, validateStepFunctionNode) ?? [];
+    }
+  }
```
thantos, line 41: How will we keep the compile errors and validation errors in sync? could we just run compile and report the `Error` nodes returned?

src/validate.ts
```diff
@@ -0,0 +1,42 @@
...
+    ) {
+      // assignment statement
+      // TODO: check if the assignment references an expression in an illegal scope
+    }
+
+    return descend();
+
+    function descend() {
+      return ts.forEachChild(node, validateStepFunctionNode) ?? [];
+    }
```
thantos, line 40: Now we don't need to remember to descend in each of the visits.

```ts
return (
    ts.forEachChild(sf, function visit(node): ts.Diagnostic[] {
      const visitFunctionless = () => {
         if (checker.isStepFunction(node)) {
            return validateStepFunctionNode(node);
         };
         return [];
      }
      
      const functionlessDiagnostics = visitFunctionless();

      const childDiag = ts.forEachChild(node, visit) ?? [];

      return [...functionlessDiagnostics, childDiag ];
    }) ?? []
  );
```

src/checker.ts
```diff
@@ -0,0 +1,227 @@
...
+    ...checker,
+    isAppsyncResolver,
+    isEventBusRuleMapFunction,
+    isEventBusWhenFunction,
+    isFunctionlessType,
+    isNewEventBusRule,
+    isNewEventBusTransform,
+    isReflectFunction,
+    isStepFunction,
+  };
```
thantos, line 64: Any way to avoid maintaining the signatures in 2 places?

thantos (changes requested):

src/checker.ts
sam-goodwin: How about this? And then we just leave the type as inferred. 

```
ReturnType<typeof makeFunctionlessChecker>
```

src/validate.ts
sam-goodwin: We don't have access to a TransformationContext, so we can't run the transformer code. I think we will have to separate out the validation logic into their own functions and reuse as much as possible.

sam-goodwin: We also need to surface some of the errors that will happen at synth time to compile time. E.g. a Step Function trying to set a variable outside of the scope of a Map task.

src/validate.ts
sam-goodwin: Even simpler:
```ts
return (function visit(node: ts.Node): ts.Diagnostic[] {
  const visitFunctionless = () => {
    if (checker.isStepFunction(node)) {
      return validateStepFunctionNode(node);
    }
    return [];
  };

  return [...visitFunctionless(), ...(ts.forEachChild(node, visit) ?? [])];
})(sf);
```

sam-goodwin: Too simple? lol
```ts
return (function visit(node: ts.Node): ts.Diagnostic[] {
  return [
    ...(checker.isStepFunction(node) ? validateStepFunctionNode(node) : []),
    ...(ts.forEachChild(node, visit) ?? []),
  ];
})(sf);
```

sam-goodwin:
I enabled eslint, so some changes re purely cosmetic.

thantos:
Build is failing

> error Package "" refers to a non-existing file '"/home/runner/work/functionless/functionless-language-service"'.

src/validate.ts
```diff
@@ -0,0 +1,70 @@
...
+        ].concat(children);
+      }
+    }
+    return children;
+  }
+
+  // ts.forEachChild terminates whenever a truth value is returned
+  // ts.visitEachChild requires a ts.TransformationContext, so we can't use that
+  // this wrapper uses a mutable array to collect the results
+  function visitEachChild<T>(
```
thantos, line 60: `validateEachChild`?

The behavior is different from `visitEachChild` which returns nodes back into the location of the node being processed.

thantos: Or map/reduce eachChild?

src/validate.ts
```diff
@@ -0,0 +1,70 @@
...
+  logger?: {
+    info(message: string): void;
+  }
+): typescript.Diagnostic[] {
+  logger?.info("Beginning validation of Functionless semantics");
+
+  return (function visit(node: typescript.Node): typescript.Diagnostic[] {
+    return [
+      ...(checker.isStepFunction(node) ? validateStepFunctionNode(node) : []),
+      ...(visitEachChild(node, visit) ?? []),
```
thantos, line 26: This will process the children of step function either both visit and `validateStepFunction` right? Did you want to short circuit visit once a closure is found? Also, what is the plan to validate function reference from within a closure.

```ts
const math = (x,y) => x + y;
new StepFunction(..., () => math(1,2));
```

test-app/yarn.lock
```diff
@@ -2574,6 +2579,14 @@ function-bind@^1.1.1:
   resolved "https://registry.yarnpkg.com/function-bind/-/function-bind-1.1.1.tgz#a56899d3ea3c9bab874bb9773b7c5ede92f4895d"
   integrity sha512-yIovAzMX49sF8Yl58fSCWJ5svSLuaibPxXQJFLmBObTuCr0Mf1KiPopGM9NiFjiYBCbfaa2Fh6breQ6ANVTI0A==
 
+functionless@^0.6.18:
```
thantos, line 2582: Should functionless be a peer dep for the language service or is this an uncommon consumption pattern?

test/eventpattern.test.ts
```diff
@@ -42,7 +42,7 @@ describe("event pattern", () => {
     test("index access", () => {
       ebEventPatternTestCase(
         reflect<EventPredicateFunction<TestEvent>>(
-          (event) => event["source"] === "lambda"
```
thantos, line 45: Kind of defeats the purpose of the test?

thantos: Could update to use `detail-type` (I hadn't added detail type when I created the test)

test/eventtargets.test.ts
```diff
@@ -23,7 +23,7 @@ test("event path", () => {
 
 test("event path index access", () => {
   ebEventTargetTestCase<testEvent>(
-    reflect((event) => event["source"]),
+    reflect((event) => event.source),
     aws_events.RuleTargetInput.fromEventPath("$.source")
   );
 });
```
thantos, line 29: Same as above, this was testing index access.

thantos (changes requested):

thantos:
Small comments mostly. And for future reference, lets try to keep functional and cosmetic changes as separate PRs.

thantos:
Ohh, also, add `eslint` to the `lint-staged` pre-commit command (with prettier) so es-lint runs on every commit on staged changes.

src/validate.ts
sam-goodwin: `valdiateEachChild` seems fine enough. I also struggled with the name.

test-app/yarn.lock
sam-goodwin: `functionless` is a peer dependency of `@functionless/language-service`

.projenrc.js
```diff
@@ -161,4 +164,14 @@ project.testTask.env("AWS_SECRET_ACCESS_KEY", "test");
 
 project.addPackageIgnore("/test-app");
 
+project.eslint.addRules({
+  quotes: "off",
```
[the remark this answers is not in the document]
sam-goodwin: Yes, it was affecting every file.

test/eventpattern.test.ts
sam-goodwin: Hmm I don't remember changing this.

sam-goodwin: Oh, linting changed it. Didn't realize this test was on purpose using that syntax.

sam-goodwin: Will es-ignore

test-app/package.json
```diff
@@ -19,6 +19,7 @@
     "aws-cdk": "2.20.0",
     "constructs": "10.0.0",
     "functionless": "file:../",
+    "@functionless/language-service": "file:../../functionless-language-service",
```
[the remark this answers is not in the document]
sam-goodwin: Removed for now because i can't release language-service until the validate function exists in functionless. Will first merge and release this, then i will release language-service, then come back and add it to test-app with another release.

PR #108, opened 2022-04-25
feat: formalize service integration handling for more complex cases

thantos:
Creates an interface to represent interactions with `Resources` via "integrations".

* Removes the previous "make callable" in favor of an interface with optional methods to implement.
* Strongly typed methods on the interface reduce the use of `any`
* Supports consistent (and overridable) errors for all integration support `${this.name} is not supported by context ${context.kind}.`

### Motivation

* In #103 we are adding native lambda function support via serialized typescript/javascript function code. This change will need to hook with all of the existing functionless resources including the ability to call and thus grant permissions to lambda function, event bus, step function, etc. In this case the `callExpr` is not needed as we only need to pass the `Function` itself. 
   * `native: (context: Function) => void`
* Change #103 will also require updating all or most of the existing `makeCallable` type locations to support the new `native` integration. This change will make said update easier and cleaner.
* While I have not tested it, this change should make it easier to plug arbitrary resources into functionless to be called.
* Mutating methods is error prone, its easy to miss support for things like getters and setters. This method does not mutate the resources at all, it only wraps the real object.
* Error messages for context handling were inconsistent. 
* Removed most of the `@ts-ignore`s

```ts
export class Function implements IntegrationHandler {
  readonly kind = "Function",

  // Integration Handler for ASL
  asl(call, context) {
      // return Step Function task.
  }

  default = makeIntegration<() => string>({
      kind: "Function.default",
      asl: (call, context) => {
          // return step function task
      }
  });
}

// an interface to provide the actual callable methods to users
export interface Function {
  // call me to send a string payload
  (payload: String) => string
}

// use

const func1 = new Function(...);
// uses the ASL
new StepFunction(..., () => {
  func1("some string");
  func1.default();
})
```


src/integration.ts
```diff
@@ -0,0 +1,134 @@
...
+ */
+export interface IntegrationHandler {
+  /**
+   * Integration Handler kind - for example StepFunction.describeExecution
+   */
+  readonly kind: string;
+  unhandledContext?: <T>(kind: string, context: CallContext) => T;
+  /**
+   * Integrate with VTL applications like AppAsync.
+   * @private
```
sam-goodwin, line 61: private? What's this for?

src/integration.ts
```diff
@@ -0,0 +1,134 @@
...
+type AllIntegrations = {
+  [key in keyof IntegrationHandler]: IntegrationHandler[key];
+};
+
+/**
+ * Wrapper class for Integration handlers that provides default error handling for unsupported integrations.
+ * @private
+ */
+export class Integration implements Omit<AllIntegrations, "unhandledContext"> {
+  readonly kind: string;
```
sam-goodwin, line 80: Should this be abstract to ensure that classes that extend Integration set it?

src/integration.ts
```diff
@@ -0,0 +1,134 @@
...
+  constructor(readonly integration: IntegrationHandler) {
+    this.kind = integration.kind;
+  }
+
+  private unhandledContext<T>(context: CallContext): T {
```
sam-goodwin, line 85: I'm surprised this is private since it's in the interface

src/integration.ts
```diff
@@ -0,0 +1,134 @@
...
+ *    func1.default();
+ * })
+ * ```
+ *
+ * If an integration does not support an integration type, leave the function undefined or throw an error.
+ *
+ * Implement the unhandledContext function to customize the error message for unsupported contexts.
+ * Otherwise the error will be: `${this.name} is not supported by context ${context.kind}.`
+ */
+export interface IntegrationHandler {
```
sam-goodwin, line 53: IIntegration?

src/step-function.ts
```diff
@@ -56,32 +52,27 @@ export namespace $SFN {
...
-      throw new Error(
-        `$SFN.wait is only available in the ${ASL.ContextName} context`
-      );
-    }
-    const seconds = call.args[0].expr;
-    if (seconds === undefined) {
-      throw new Error(`the 'seconds' argument is required`);
-    }
+  export const waitFor = makeIntegration<(seconds: number) => void>({
+    ...tableIntegrationBase("waitFor"),
```
sam-goodwin, line 56: table? This seems out of place or poorly named?

src/step-function.ts
```diff
@@ -153,18 +132,25 @@ export namespace $SFN {
    * @param props configure the maxConcurrency
    * @param callbackfn function to process each item
    */
-  // @ts-ignore
-  export function forEach<T>(
-    array: T[],
-    props: {
-      maxConcurrency: number;
+  export const forEach = makeIntegration<forEach>({
```
sam-goodwin, line 135: I worry about how these function signatures have been destroyed. If a developer is browsing through trying to learn what they can call, won't they be like "wtf"? Also, can we still document the functions so that it renders properly in both the IDE and generated docs?

src/step-function.ts
```diff
@@ -297,40 +249,52 @@ export namespace $SFN {
...
+        Branches: paths.items.map((func) => ({
+          StartAt: context.getStateName(func.body.step()!),
+          States: context.execute(func.body),
+        })),
+      };
+    },
+  });
+}
+
+function tableIntegrationBase(
```
sam-goodwin, line 287: Why is this called table?

src/table.ts
```diff
@@ -77,205 +77,202 @@ export class Table<
...
-        call.getArgument("input")?.expr,
-        "ObjectLiteralExpr"
-      )
-    );
-    const request = vtl.var(
-      `{"operation": "GetItem", "version": "2018-05-29"}`
-    );
-    vtl.qr(`${request}.put('key', ${input}.get('key'))`);
-    addIfDefined(vtl, input, request, "consistentRead");
+  public getItem = makeIntegration<
```
sam-goodwin, line 81: Same question around documentation and ide ergonomics. Let's make sure these convenience methods don't hurt the DX for a user.

src/integration.ts
```diff
@@ -0,0 +1,134 @@
...
+ * ```ts
+ * export class Function implements IntegrationHandler {
+ *    readonly kind = "Function",
+ *
+ *    // Integration Handler for ASL
+ *    asl(call, context) {
+ *       // return Step Function task.
+ *    }
+ *
+ *    default = makeIntegration<() => string>({
```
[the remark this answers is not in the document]
thantos: It was an example. I'll... Name it like "special payload"

src/integration.ts
thantos: It will omit this function from the docs. , We don't want StepFunctions.asl to show up on the docs as it's on the computer side, not the user side.

src/integration.ts
```diff
@@ -0,0 +1,134 @@
...
+   */
+  vtl?: (call: CallExpr, context: VTL) => string;
+  /**
+   * Integrate with ASL applications like StepFunctions.
+   * @private
+   */
+  asl?: (call: CallExpr, context: ASL) => Omit<State, "Next">;
+}
+
+type AllIntegrations = {
```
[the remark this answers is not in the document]
thantos: Ohh, shit, I meant to have this make all of the methods required. Will fix.

src/integration.ts
```diff
@@ -0,0 +1,134 @@
...
+  [key in keyof IntegrationHandler]: IntegrationHandler[key];
+};
+
+/**
+ * Wrapper class for Integration handlers that provides default error handling for unsupported integrations.
+ * @private
+ */
+export class Integration implements Omit<AllIntegrations, "unhandledContext"> {
```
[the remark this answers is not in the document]
thantos: Makes sense. Will fix the all integrations too.

src/integration.ts
thantos: I wasn't intending anything to extend integration, only the interface. 

We instantiate the class around any object of the interface at runtime.

src/step-function.ts
thantos: It has things that are `Table.operation` specific.

src/asl.ts
```diff
@@ -836,27 +839,27 @@ export class ASL {
...
+      const serviceCall = findIntegration(expr);
       if (serviceCall) {
         if (
           expr.expr.kind === "PropAccessExpr" &&
           (expr.expr.name === "waitFor" || expr.expr.name === "waitUntil")
         ) {
           delete (props as any).ResultPath;
-          return {
-            ...(serviceCall as any)(expr, this),
+          return <State>{
```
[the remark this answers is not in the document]
thantos: As long as it isn't a TSX file, haha

src/function.ts
```diff
@@ -28,46 +28,38 @@ export type AnyLambda = Function<any, any>;
...
-        throw new Error(`invalid Function call context: ${context}`);
-      }
-    });
+    const request = context.var(
+      `{"version": "2018-05-29", "operation": "Invoke", "payload": ${payload}}`
+    );
+    return context.json(request);
+  }
+
+  asl(call: CallExpr, context: ASL) {
```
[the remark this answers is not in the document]
thantos: need eslint ;)

src/integration.ts
thantos: Removed from the interface. Not for external consumption. (the signature is also different).

src/integration.ts
thantos: Sure.

src/table.ts
thantos: With `const`

image

image

With function

image

image

src/step-function.ts
sam-goodwin: Document the forEach and map parameters. Do they properly make it to the IDE? What do the generated docs look like?

thantos: You were right, after messing it, union anonymous method signatures were not correctly documented. Simple signatures like `getItem` were fine, they showed the ts-docs, but the union/overload types did not.

This seems to work

```ts
  interface ForEach {
    /**
     * Process each item in an {@link array} in parallel and run with the default maxConcurrency.
     *
     * Example:
     * ```ts
     * new ExpressStepFunction(this, "F"} (items: string[]) => {
     *   $SFN.forEach(items, { maxConcurrency: 2 }, item => task(item));
     * });
     * ```
     *
     * @param array the list of items to process
     * @param callbackfn function to process each item
     */
    <T>(
      array: T[],
      callbackfn: (item: T, index: number, array: T[]) => void
    ): void;
    /**
     * Process each item in an {@link array} in parallel and run with the default maxConcurrency.
     *
     * Example:
     * ```ts
     * new ExpressStepFunction(this, "F"} (items: string[]) => {
     *   $SFN.forEach(items, { maxConcurrency: 2 }, item => task(item));
     * });
     * ```
     *
     * @param array the list of items to process
     * @param props configure the maxConcurrency
     * @param callbackfn function to process each item
     */
    <T>(
      array: T[],
      props: {
        maxConcurrency: number;
      },
      callbackfn: (item: T, index: number, array: T[]) => void
    ): void;
  }

  export const forEach = makeIntegration<ForEach>({
    ...sfnIntegrationBase("forEach"),
    asl(call, context) {
      return mapOrForEach(call, context);
    },
  });
```

In fact it is better than before because now each of the signatures have their own params documented separately. 

I am unsure how to generate the docs.

thantos: image
image

src/function.ts
```diff
@@ -28,46 +28,38 @@ export type AnyLambda = Function<any, any>;
...
  */
-export class Function<P, O> {
+export class Function<P, O> implements IIntegration {
   readonly kind = "Function" as const;
 
   // @ts-ignore - this makes `F` easily available at compile time
   readonly __functionBrand: ConditionalFunction<P, O>;
 
-  constructor(readonly resource: aws_lambda.IFunction) {
-    return makeCallable(this, (call: CallExpr, context: VTL | ASL): any => {
```
sam-goodwin, line 38: Looking at the code, it looks like `Function` is no longer callable.

src/index.ts
```diff
@@ -4,6 +4,7 @@ export * from "./declaration";
 export * from "./error";
 export * from "./event-bridge";
 export * from "./expression";
+export { IIntegration as IntegrationHandler } from "./integration";
```
sam-goodwin, line 7: This seems like a mistake?

src/integration.ts
```diff
@@ -0,0 +1,146 @@
...
+} & {
+  /**
+   * Integration Handler kind - for example StepFunction.describeExecution
+   */
+  readonly kind: string;
+  /**
+   * Optional method that allows overriding the {@link Error} thrown when a integration is not supported by a handler.
+   */
+  readonly unhandledContext?: (kind: string, context: CallContext) => Error;
+};
```
sam-goodwin, line 83: I think this can be simplified to the following:

```ts
export interface IIntegration extends Partial<IntegrationMethods> {
  readonly kind: string;
  readonly unhandledContext?: (kind: string, contet: CallContext) => Error;
}
```

src/integration.ts
```diff
@@ -0,0 +1,146 @@
...
+/**
+ * Integration types supported by Functionless.
+ *
+ * Add an integration by creating any object that has a property named "kind" and either implements the
+ * {@link IIntegration} interface or has methods that implement it (or both).
+ *
+ * Example showing both strategies:
+ *
+ * ```ts
+ * export class Function implements IntegrationHandler {
```
sam-goodwin, line 32: `IIntegration`?

src/integration.ts
```diff
@@ -0,0 +1,146 @@
...
+ *    func1.specialPayload();
+ * })
+ * ```
+ *
+ * If an integration does not support an integration type, leave the function undefined or throw an error.
+ *
+ * Implement the unhandledContext function to customize the error message for unsupported contexts.
+ * Otherwise the error will be: `${this.name} is not supported by context ${context.kind}.`
+ */
+export type IIntegration = {
```
sam-goodwin, line 71: On second thought, should we just call this `Integration` for simplcity? The class, `Integration`, can be called `IntegrationImpl` since it'll never be seen by consumers.

src/integration.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  public asl(call: CallExpr, context: ASL): Omit<State, "Next"> {
+    if (this.integration.asl) {
+      return this.integration.asl(call, context);
+    }
+    return this.unhandledContext(context);
+  }
+}
+
+export function makeIntegration<F extends AnyFunction>(
```
sam-goodwin, line 119: I think this needs some docs. It does nothing except change the type.

src/step-function.ts
```diff
@@ -56,32 +52,27 @@ export namespace $SFN {
...
-      throw new Error(
-        `$SFN.wait is only available in the ${ASL.ContextName} context`
-      );
-    }
-    const seconds = call.args[0].expr;
-    if (seconds === undefined) {
-      throw new Error(`the 'seconds' argument is required`);
-    }
+  export const waitFor = makeIntegration<(seconds: number) => void>({
+    ...sfnIntegrationBase("waitFor"),
```
sam-goodwin, line 56: This makes me wonder why aren't we using classes and inheritance? Seems overly complex? We've gone from a Function to now spreading a base object.

sam-goodwin:
```ts
export const waitFor = new BaseStepFunctionIntegration<(seconds: number) => void>({
  asl(call) { .. }
});
```

src/step-function.ts
```diff
@@ -297,40 +278,52 @@ export namespace $SFN {
...
+): Pick<IIntegration, "kind" | "unhandledContext"> {
+  return {
+    kind: `$SFN.${methodName}`,
+    unhandledContext(kind, context) {
+      throw new Error(
+        `${kind} is only allowed within a '${VTL.ContextName}' context, but was called within a '${context.kind}' context.`
+      );
+    },
+  };
 }
```
sam-goodwin, line 327: This is a poor-mans abstract class. Why not just use a class? Let's try and keep the coding patterns simple to understand and read. Use conventional patterns.

```ts
export abstract class BaseStepFunctionIntegration extends BaseIntegration {
  constructor(readonly methodName: string) {
    super(`$SFN.${methodName}`);
  }

  public abstract asl(call: CallExpr, context: ASL): State;
}
```

src/function.ts
sam-goodwin: Ok, I now understand. The Function's interface is callable so TS won't complain. And we transform the code to the AST so the Function is never actually called - it is always called when we interpret the AST.

I assume this will change when we merge the closure serializer?

sam-goodwin (changes requested):

src/function.ts
thantos: yeah, in the closure serializer, I have a `native` integration, that is actually 3 functions right now: `bootstrap`, `prewarm`, `call`.

1. Bootstap - For wiring CDK
2. Prewarm - For anything that should run once for each function instance (clients)
3. Call - replaces the `func()` inline

May not stay this way, but the point is, each context can get whatever they need from the integrations in a well typed way.

Instead of "calling" func(), the context will do `new Integration(func).asl(call, context)` or `func.asl?.(call, context)`. The class just provides a nice wrapper for error handling for unsupported contexts.

src/index.ts
thantos: yes, lol, I hate that vscode tries to maintain old names when renaming...

src/integration.ts
thantos: duhh.. wtf

src/step-function.ts
thantos: or just a function, same result though. 

```ts
export const waitFor = makeStepFunctionIntegration<(seconds: number) => void>({
  asl(call) { .. }
});

const makeStepFunctionIntegration<F extends AnyFunction> = (integration: Omit<Integration, 'kind'>, operation: string) => makeIntegration<F>({
   kind: `SFN.${kind}`,
   ...integration
})
```

src/step-function.ts
thantos: We'll see, I am trying to avoid working with classes to keep the closure serialization simpler, also keeps the context of `this` simpler. I generally prefer simple objects with interfaces over classes. I'll try with the wrapper function detailed here https://github.com/sam-goodwin/functionless/pull/108#discussion_r859514099. This should remove the need to spread the resulting object.

src/function.ts
thantos: Re-reading your original statement, I should add the that class is never physically turned into an function. It is always an object that looks like a function. During interpretation of the AST, we wrap the object in `ref()` and then call the `.asl()`, `.vtl()`, etc properties on said object.

No more manipulating the object into a function with properties.

sam-goodwin: Are you re-writing the AST for the closure serializer? How does `func()` translate into `bootstrap` -> `prewarm` -> `call`?

sam-goodwin (approved):

sam-goodwin (approved):

src/function.ts
thantos: yes. 

1. Compile
   1. find all integrations in each closure
   2. rewrite integration call to call `integration.native.call` inline in the closure
2. Synth
  1. Run bootstrap for each integration to wire up CDK
  2.  Add pre-warm to the "factory" portion of the closure supported by pulumi for each integration (code that runs before `handle`)
3. Runtime
   1. Pre-warm methods run
   2. Handler is run, eventually, calling the `integration.native.call`

sam-goodwin: Love it!

PR #109, opened 2022-04-26
feat: localstack testing

thantos:
Depends on #108 

Run `yarn test` or `./script/localstack` to start a localstack client and deploy the CDK bootstrap. 

Tests can be written against localstack to test the real behavior of our resources.


sam-goodwin:
Why does this depend on #108

thantos:
> Why does this depend on #108

Just because #103 depended on this and #108

sam-goodwin (approved):

PR #113, opened 2022-04-28
feat: generalize all app sync integration hooks

### LINKED ISSUE #117: Support code-first AppSync
https://docs.aws.amazon.com/cdk/api/v2/docs/aws-appsync-alpha-readme.html#code-first

https://docs.aws.amazon.com/cdk/api/v2/docs/aws-appsync-alpha-readme.html#code-first-schema

```ts
api.addQuery(...)
```

thantos:
Closes #117 

Previous, much of the integration logic for AppSync were in the app sync logic. This change moves all of the per integration (ex: app sync calling lambda) into the specific integrations (ex: out of app sync file and into the `Function` file).

### Changes:

* Change the `vtl` integration into `appSyncVtl` as the integrations are specific to AppSync and not other VTL implementations (Api Gateway)
* Change the `vtl`/`appSyncVtl` integration from a single function to 3 methods.
   * request - the main request vtl template
   * result - a post request processor that can transform the results from the service call (used by step functions now)
   * dataSource - retrieve a CDK data source for the integration
*  Use CDK's `tryFindChild` to resolve hierarchical singleton constructs instead of a separate map.
* Add a static `kind` to each of the integration types so that they may be found by the compiler. Previously, only the top level objects (`Function`, `StepFunction`, `Table`) had `kind`. Now each of the integrations will have a `kind` (ex: `StepFunction.describeExecution`)
* Deleted things
* Compile: Only add `functionless` import when necessary.
* Improve the creation of localstack tests, reduce boilerplate
* fix: format function names for app sync
* feat: Added `getField` to use Functionless AppSync Resolver with CodeFirst
   * Note: a larger change is planned with #119 to refactor app sync, however this change adds some more capability without changing the interface.
   * [ ] Update Docs

### Code First App Sync Resolver

```ts
const commentResolver = new functionless.AppSyncResolver(...);
const getPostResolver = new functionless.AppSyncResolver(...);

/*
  type Query {
    getPost(postId: string!): Post
  } 

 type Post {
  postId: ID!
  title: String!
  comments(nextToken: String, limit: Int): CommentPage
 }

 type CommentPage {
  nextToken: String
  comments: [Comment]!
 }

 type Comment {
  postId: ID!
  commentId: ID!
  commentText: String!
  createdTime: String!
 }
 */

const post = api2.addType(
  new appsync.ObjectType("Post", {
    definition: {
      postId: appsync.GraphqlType.id({
        isRequired: true,
      }),
      title: appsync.GraphqlType.string({
        isRequired: true,
      }),
    },
  })
);

const comment = api2.addType(
  new appsync.ObjectType("Comment", {
    definition: {
      postId: appsync.GraphqlType.id({
        isRequired: true,
      }),
      commentId: appsync.GraphqlType.id({
        isRequired: true,
      }),
      commentText: appsync.GraphqlType.string({
        isRequired: true,
      }),
      createdTime: appsync.GraphqlType.string({
        isRequired: true,
      }),
    },
  })
);

const commentPage = api2.addType(
  new appsync.ObjectType("CommentPage", {
    definition: {
      nextToken: appsync.GraphqlType.string(),
      comments: appsync.GraphqlType.intermediate({
        intermediateType: comment,
        isRequiredList: true,
      }),
    },
  })
);

post.addField({
  fieldName: "comments",
  field: commentResolver.getField(api2, commentPage.attribute(), {
    args: {
      nextToken: appsync.GraphqlType.string(),
      limit: appsync.GraphqlType.int(),
    },
  }),
});

api2.addQuery(
  "getPost",
  getPostResolver.getField(api2, post.attribute(), {
    args: {
      postId: appsync.GraphqlType.string({ isRequired: true }),
    },
  })
);
```


sam-goodwin:
Why is this merging into local stack instead of main?

sam-goodwin:
Is this a `feat` or `chore`?

src/appsync.ts
```diff
@@ -816,33 +737,11 @@ function getUniqueName(api: appsync.GraphqlApi, name: string): string {
...
-  target: CanReference | null,
-  compute: () => appsync.BaseDataSource
-): appsync.BaseDataSource {
-  const ds = getDataSources(api);
-  const key = target ?? None;
-  if (!ds.has(key)) {
-    ds.set(key, compute());
-  }
-  return ds.get(key)!;
+export interface AppSyncVtlIntegration {
```
sam-goodwin, line 740: Documentation?

src/util.ts
```diff
@@ -134,3 +93,9 @@ export const isPrimitive = (val: any): val is PrimitiveValue => {
     val === null
   );
 };
+
+
+export const singletonConstruct = <T extends Construct>(scope: Construct, id: string, create: (scope: Construct, id: string) => T): T => {
```
sam-goodwin, line 98: Be sure to run prettier automatically in the ide or via cli before committing changes. I'm noticing a bunch of formatting changes and am wondering if prettier is running for you?

sam-goodwin (changes requested):

src/aws.ts
```diff
@@ -65,192 +64,194 @@ export namespace $AWS {
     /**
      * @see https://docs.aws.amazon.com/step-functions/latest/dg/connect-ddb.html
      */
-    export const DeleteItem =
-      makeDynamoIntegration<
-        <
-          T extends Table<any, any, any>,
-          Key extends TableKey<
+    export const DeleteItem = makeDynamoIntegration<
```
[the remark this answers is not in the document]
thantos: Adding a second type parameter changed prettier's opinion.

src/aws.ts
```diff
@@ -354,7 +356,7 @@ export namespace $AWS {
       },
       unhandledContext(kind, context) {
         throw new Error(
-          `$AWS.${kind} is only available within an '${ASL.ContextName}' context, but was called from within a '${context.kind}' context.`
+          `$AWS.${kind} is only available within an '${ASL.ContextName}' context, but was called from within a '${context}' context.`
```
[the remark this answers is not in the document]
thantos: It is now. In the case of an error, I think we only care about the kind. This change allowed me to make the integration types into objects instead of functions

 `integrationImpl.appSyncVtl.request(call, context)`

thantos:
> Why is this merging into local stack instead of main?

@sam-goodwin. This change is sitting on the localstack testing change (#109), it may not need to be, #103 already depends on #109 and it can only have one parent.

Need to look at why the localstack testing branch is failing build.

thantos:
> Is this a `feat` or `chore`?

I considered it a `feat` because now we can extend app sync integrations without changing app sync. Will update the name to reflect.

src/integration.ts
```diff
@@ -161,4 +164,24 @@ export function findIntegration(call: CallExpr): IntegrationImpl | undefined {
   }
 }
 
+/**
+ * Dive until we find a integration object.
+ */
+export function findDeepIntegration(
```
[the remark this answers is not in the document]
thantos: I moved it from utils to integration and renamed it. Starting to move away from "services" to just be solve contained `integrations` that may reference their internal services or take them as arguments.

src/util.ts
thantos: Looks like it didn't format. Been having issues with my vscode where the tests and ts server are taking too much memory... likely prettier didn't get to run when I saved.

Maybe we need to add git hooks?

thantos: https://github.com/sam-goodwin/functionless/pull/116

sam-goodwin (approved):

sam-goodwin:
Is this a breaking change?

thantos:
> Is this a breaking change?

Nope, all backwards compatible. Add docs for the new feature, will have them out in 5.

sam-goodwin:
> > Is this a breaking change?
> 
> 
> 
> Nope, all backwards compatible. Adding docs for the new feature, will have them out in 5. 


> - Change the vtl integration into appSyncVtl as the integrations are specific to AppSync and not other VTL implementations (Api Gateway)
> - Change the vtl/appSyncVtl integration from a single function to 3 methods.

thantos:
> > > Is this a breaking change?

Only breaking for internal Functionless implementation, consumers would not be impacted. Someone may have a nasty merge if they were deep into a larger feature (like I am doing right now with #103 )

sam-goodwin:
Is the interface for integrations public or private? I think private. In practice, I doubt this breaks anyone. Just trying to be principled.

thantos:
> Is the interface for integrations public or private? I think private. In practice, I doubt this breaks anyone. Just trying to be principled.

Totally. Right now I was considering integration to be internal. I was going to make an issue where we'd make integrations public, documented, and with examples. At this point, I'd agree, this is breaking for integrators.

sam-goodwin:
Oh I meant to say public. They are public lol. Which is why I think this is a breaking change.

thantos:
Well it wasn't really possible to implement one before this change since AppSync had a bunch of internal logic. Do you think anyone is using it? I guess I am OK with marking this as break if you want.

Either way, this it the effort I think is left to be officially "publicly supported": #121

sam-goodwin:
Happy with that. Let's move forward.

sam-goodwin:
We can consider it private for now and therefore not breaking.

PR #116, opened 2022-04-29
chore: add githooks to projen and prettier precommit hook to project

thantos:
* Adds a githooks component to projen
* Add a preCommit hook to run prettier on changed files before commit.

https://github.com/sam-goodwin/functionless/pull/113#discussion_r861292702


.projenrc.js
```diff
@@ -1,6 +1,41 @@
...
+/**
+ * Adds githooks into the /git/hooks folder during projen synth.
+ *
+ * @see https://git-scm.com/docs/githooks
+ *
+ * githooks: {
+ *    preCommit: ["lint-staged"] // array of bash commands
+ * }
+ */
+class GitHooksComponent extends Component {
```
sam-goodwin, line 14: Can't you use the TextFile component instead of this?

sam-goodwin (changes requested):

.projenrc.js
thantos: Could, yeah, didn't know it existed. Wanted to play with supporting githooks as component instead of just making the file.

TextFile doesn't have chmod option though.

sam-goodwin (approved):

sam-goodwin:
Should this be chore?

PR #124, opened 2022-05-05
feat: generalize all app sync integration hooks

thantos:
#113 was not merged into the right branch. Recreating to merge again.


thantos:
Was approved in #113 , only change here was a rebase on main (dep upgrades)

PR #146, opened 2022-05-21
fix(appsync): set a temporary variable for BinaryExpr inside an object put

brendo-m:
Fixes a weird but I ran into where `$obj.put('x', 1 + 1)` would cause a 500

### DIFF
--- a/src/vtl.ts
+++ b/src/vtl.ts
@@ -1,4 +1,4 @@
-import { CallExpr, Expr, FunctionExpr } from "./expression";
+import { CallExpr, Expr, FunctionExpr, isBinaryExpr } from "./expression";
 import { isInTopLevelScope } from "./util";
 import { assertNever, assertNodeKind } from "./assert";
 import { FunctionlessNode } from "./node";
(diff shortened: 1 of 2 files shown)


src/vtl.ts
```diff
@@ -99,11 +99,19 @@ export class VTL {
...
   public put(objVar: string, name: string, expr: string | Expr) {
-    this.qr(
-      `${objVar}.put(${name}, ${
-        typeof expr === "string" ? expr : this.eval(expr)
-      })`
-    );
+    let result;
+    if (typeof expr === "string") {
+      result = expr;
+    } else if (isBinaryExpr(expr)) {
```
sam-goodwin, line 105: By the way, I think it's possible for this BinaryExpr to be an assignment where the operator is `=`

Are we handling this right?

brendo-m: Wouldn't that be a `VariableStmt`?

sam-goodwin:
Should this be fix(appsync) ?

src/vtl.ts
sam-goodwin: No, it would be an ExprtStmt(BinaryExpr) if it's assigning without const/let/var

```
a = b
```

sam-goodwin: It's an expression that returns the value that is assigned. 

```ts
this.a = (a = b)
```

brendo-m: Gross. You have to go out of your way to make it happen, but we are not handling it correctly.

```
// @ts-ignore
let y = 1;
return { x: (y = 2) };
```

Will fix in another PR

sam-goodwin (approved):

thantos:
Is this true for other expressions too? 
```ts
{
   x : -1, // unary
   y: `template ${1}`,
   z: func() // some kind of call
}
```

sam-goodwin:
Is this ready to be merged?

brendo-m:
> Is this true for other expressions too?

Unary is broken right now because VTL doesn't like the space after the sign. `- 1` fails but `-1` is fine. Templates don't work either, not sure why. 

I'll merge this and create issues to follow up

PR #154, opened 2022-05-23
fix(appsync): use temp variables for all BinaryExpr and UnaryExpr nodes

### LINKED ISSUE #149: Correctly handle unary expressions inside object puts in VTL
The compiler converts
```
{
   x : -1,
}
```

to 

```
#set($v1 = {})
$set($discard = $v1.put('x', - 1))
```

Which fails because VTL engine doesn't handle a space after the minus sign for whatever reason. `$set($discard = $v1.put('x', -1))` works

brendo-m:
Fixes #149

### DIFF
--- a/test/vtl.test.ts
+++ b/test/vtl.test.ts
@@ -71,11 +71,12 @@ test("computed property names", () => {
     ),
     payload,
     `#set($context.stash.name = $context.arguments.arg)
-#set($context.stash.value = $context.stash.name + '_test')
-#set($v1 = {})
-$util.qr($v1.put($context.stash.name, $context.arguments.arg))
-$util.qr($v1.put($context.stash.value, $context.arguments.arg))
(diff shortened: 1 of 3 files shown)


thantos (approved):
Nice! Simple fix

test/vtl.test.ts
```diff
@@ -207,7 +208,8 @@ test("if statement", () => {
       }
     }),
     payload,
-    `#if($context.arguments.list.length > 0)
+    `#set($v1 = $context.arguments.list.length > 0)
+#if($v1)
```
brendo-m, line 212: these changes are a little annoying, we don't need to set a variable if the parent node is an `IfStmt`, but I'll leave it as is for consistency unless someone feels strongly about it

sam-goodwin (approved):

PR #155, opened 2022-05-24
feat: introduce AppsyncVTL subclass of VTL

brendo-m:
This refactor is in preparation for APIGW mapping template support. Appsync and APIGW have different VTL engines and I have moved the parts that differ into abstract methods on `VTL`, which is now an abstract class. `AppsyncVTL` behaves the same as `VTL` did, `ApiGatewayVTL` will be introduced in another PR

### DIFF
--- a/src/vtl.ts
+++ b/src/vtl.ts
@@ -1,9 +1,15 @@
-import { CallExpr, Expr, FunctionExpr } from "./expression";
-import { isInTopLevelScope } from "./util";
 import { assertNever, assertNodeKind } from "./assert";
+import {
+  CallExpr,
+  Expr,
+  FunctionExpr,
+  isIdentifier,
+  PropAccessExpr,
+} from "./expression";
+import { findIntegration } from "./integration";
 import { FunctionlessNode } from "./node";
 import { Stmt } from "./statement";
-import { findIntegration } from "./integration";
+import { isInTopLevelScope } from "./util";
 
 // https://velocity.apache.org/engine/devel/user-guide.html#conditionals
 // https://cwiki.apache.org/confluence/display/VELOCITY/CheckingForNull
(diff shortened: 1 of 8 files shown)


src/vtl.ts
```diff
@@ -29,6 +47,35 @@ export class VTL {
...
+  /**
+   * Return the value of the expression.
+   * @param expr Expr to evaluate and return.
+   */
+  public abstract return(expr: string | Expr): void;
+  /**
+   * Output a variable as JSON.
+   * @param reference Name of the variable to JSONify.
+   */
+  public abstract json(reference: string): string;
```
thantos, line 69: These are Appsync only, right?

thantos: Or is the plan to support them with different implementations? (assuming it can be done)

thantos:
should the appsyncVtl integration be updated to pass the AppsyncVTL?

test/vtl.test.ts
```diff
@@ -1,8 +1,7 @@
 import "jest";
-import { $util } from "../src";
-import { AppsyncContext } from "../src";
+import { $util, AppsyncContext } from "../src";
 import { reflect } from "../src/reflect";
-import { returnExpr, appsyncTestCase } from "./util";
+import { appsyncTestCase, returnExpr } from "./util";
```
thantos, line 5: What did you run that updated these?

src/vtl.ts
brendo-m: There are equivalents. For `quiet` you can do `#set($discard = ...)`. For `return` just output the variable. This one probably won't be used/allowed for API GW, still figuring it all out. For `json` you use [The Beast](https://github.com/sam-goodwin/functionless/issues/23#issuecomment-1088336878)

test/vtl.test.ts
brendo-m: I think it was the ESLint VS Code plugin

brendo-m:
> should the appsyncVtl integration be updated to pass the AppsyncVTL?

You mean here? https://github.com/sam-goodwin/functionless/blob/f1ea86035d9bbca2b7ae7df10a0fa6b5a4fbb2f0/src/appsync.ts#L850

test/vtl.test.ts
thantos: Things I don't run because they eat all of my RAM

thantos:
> > should the appsyncVtl integration be updated to pass the AppsyncVTL?
> 
> You mean here? https://github.com/sam-goodwin/functionless/blob/f1ea86035d9bbca2b7ae7df10a0fa6b5a4fbb2f0/src/appsync.ts#L850

yes

test/vtl.test.ts
sam-goodwin: Download more ram

sam-goodwin (approved):

sam-goodwin:
Is this ready to be merged?

brendo-m:
Not yet, it may not be necessary anymore

Sent via Superhuman iOS ***@***.***>

PR #159, opened 2022-05-26
feat: update event bus target to use integrations

### LINKED ISSUE #160: Event Bus: Use integrations for targets
Currently `EventBus` targets use a single `pipe` method which accept a resource and returns a `RuleTarget`. This means that new integrations require modifying the `pipe` method. 

Other resources like StepFunctions do all integration via the `Integration` interface which should allow for #90 via external libraries.

Update EventBus with a new Integration object that returns a RuleTarget.

thantos:
Closes #160

Breaking change that updates EventBus to use an `Integration` instead of specific resource integrations.

## Before

```ts
const bus = new EventBus();
const func = new Function();
bus.when(...).pipe(func);
bus.when(...).pipe({ func, deadLetterQueue: new Queue() });
```

## After

```ts
const bus = new EventBus();
const func = new Function();
bus.when(...).pipe(func);
bus.when(...).pipe(func, { deadLetterQueue: new Queue() });
```

## Motivation

* Allows the extension of `EventBus` `*.pipe` targets without modifying the Functionless core code base or without modifying the `EventBus` code base.

## TODO

* [x] Make `.pipe` typesafe. @sam-goodwin can you help here?
   * IDE/Compiler should fail when the integration accepts a different parameter from the type that the `EventBusRule` or `Transform` will output.
 ```ts
  interface tt extends EventBusRuleInput<t1 | t2> {}
  const lambda = new Function<string, void>(
    aws_lambda.Function.fromFunctionArn(stack, "func", "")
  );
  const bus = EventBus.default<tt>(stack);

  bus
    .when(
      stack,
      "rule",
      (event): event is EventBusRuleInput<t1> => event.detail.type === "one"
    )
    .map((event) => event.detail) // returns t1 or t2 
    // should fail compilation if the types don't match
    .pipe(lambda); // should fail, expects a string
 ```

BREAKING_CHANGE: Previously, `.pipe` accepted a `Function`, `EventBus`, or `StepFunction` or an object that contained the resource and other options. `.pipe` will now accept any `Integration` that implements `eventBus.target` and a section optional `Props` argument determined by the `Integration`.


thantos:
Thanks for the save @sam-goodwin

src/event-bridge/event-bus.ts
```diff
@@ -213,6 +225,22 @@ abstract class EventBusBase<E extends EventBusRuleInput>
...
   }
 
+  eventBus: EventBusTargetIntegration<
+    E,
+    aws_events_targets.EventBusProps | undefined
+  > = {
+    target: (props, targetInput?) => {
+      if (targetInput) {
+        throw new Error("Event bus rule target does not support target input.");
+      }
```
sam-goodwin, line 235: Not sure what this means? Why is it available as an argument then?

src/event-bridge/event-bus.ts
```diff
@@ -327,3 +355,44 @@ class ImportedEventBus<E extends EventBusRuleInput> extends EventBusBase<E> {
     super(bus);
   }
 }
+
+export interface EventBusTargetIntegration<
```
sam-goodwin, line 359: Should we document these interfaces? Are they public interfaces?

src/event-bridge/rule.ts
```diff
@@ -138,13 +135,10 @@ export interface IEventBusRule<T extends EventBusRuleInput> {
    * bus.when(...).pipe(new EventBus(targetBus))
    * ```
    */
-  pipe(props: LambdaTargetProps<T>): void;
-  pipe(func: Function<T, any>): void;
-  pipe(bus: EventBus<T>): void;
-  pipe(props: EventBusTargetProps<T>): void;
-  pipe(props: StateMachineTargetProps<T>): void;
-  pipe(props: StepFunction<T, any>): void;
-  pipe(props: ExpressStepFunction<T, any>): void;
```
sam-goodwin, line 147: Nice!

src/function.ts
```diff
@@ -75,6 +86,19 @@ export class Function<P, O> implements Integration {
       ResultSelector: "$.Payload",
     };
   }
+
+  public eventBus: EventBusTargetIntegration<
```
sam-goodwin, line 90: should this be readonly?

sam-goodwin: Does this show up as a function in the IDE or a method?

sam-goodwin (approved):

src/event-bridge/event-bus.ts
thantos: `PIpe` takes any integration and event bus is an integration. however, event bus cannot accept transformed input (`.map`).

I updated the types to fail when an event bus is given. Keeping the error because someone can ignore the type error.

src/event-bridge/event-bus.ts
thantos: Errr. they are to integrators. I'll add some docs.

src/function.ts
thantos: It does show up, yeah. I'd like for it not to, however, we now use the type to constrain the `pipe` function.

src/event-bridge/event-bus.ts
```diff
@@ -327,3 +374,60 @@ class ImportedEventBus<E extends EventBusRuleInput> extends EventBusBase<E> {
...
+ * const myEbIntegration = { // an Integration
+ *    kind: 'myEbIntegration',
+ *    eventBus: {
+ *       target: (props, targetInput) => {
+ *           // return a RuleTarget
+ *       }
+ *    }
+ * }
+ *
+ * new EventBus().when(() => true).pipe(myEbIntegration);
```
sam-goodwin, line 391: This syntax is not correct, missing scope and id

sam-goodwin (changes requested):

src/aws.ts
```diff
@@ -440,7 +440,7 @@ export namespace $AWS {
       operationName: Op,
       integration: Omit<
         Integration<F, `$AWS.DynamoDB.${Op}`>,
-        "kind" | "native"
+        "kind" | "native" | "__functionBrand"
```
sam-goodwin, line 443: DO we need a type alias for this behavior? Why do these fields need to be omitted?

src/event-bridge/event-bus.ts
```diff
@@ -99,11 +99,28 @@ export interface IEventBusFilterable<E extends EventBusRuleInput> {
...
 
 export interface IEventBus<E extends EventBusRuleInput = EventBusRuleInput>
-  extends IEventBusFilterable<E> {
+  extends IEventBusFilterable<E>,
+    Integration<
+      (
+        event: EventBusPutEventInput<E>,
+        ...events: EventBusPutEventInput<E>[]
+      ) => void,
+      "EventBus",
```
sam-goodwin, line 108: Should this be the first argument?

src/event-bridge/event-bus.ts
```diff
@@ -374,3 +419,75 @@ class ImportedEventBus<E extends EventBusRuleInput> extends EventBusBase<E> {
...
+ *
+ * new EventBus().when(() => true).pipe(myEbIntegration);
+ * ```
+ */
+export interface EventBusTargetIntegration<
+  // the payload type we expect to be transformed into before making this call.
+  P,
+  Props extends object | undefined = undefined
+> {
+  __payloadBrand: P;
```
sam-goodwin, line 444: Document why you need this. `target` has no relation to `P`, so you store this here to carry the `P` type forward and ensure data integrity? Where are you using this to type check? Can you link to that code with `{@link}`?

When we use brands like this, where there is actually no value present even though the types suggests otherwise, we should always provide a comment.

src/event-bridge/transform.ts
```diff
@@ -59,19 +58,21 @@ export class EventBusTransform<T extends EventBusRuleInput, P> {
...
+  pipe<
+    I extends Integration<any, string, EventBusTargetIntegration<P, Props>> & {
+      eventBus: EventBusTargetIntegration<P, Props>;
+    },
+    Props extends object | undefined
+  >(
+    integration: NonEventBusIntegration<I>,
+    ...props: Parameters<DynamicProps<Props>>
+  ) {
+    pipe(this.rule, integration, props[0] as Props, this.targetInput);
```
sam-goodwin, line 70: Why is this not returned?

src/function.ts
```diff
@@ -1,10 +1,13 @@
 import fs from "fs";
 import path from "path";
 import * as appsync from "@aws-cdk/aws-appsync-alpha";
+
+// @ts-ignore - imported for typedoc
 import { serializeFunction } from "@functionless/nodejs-closure-serializer";
```
sam-goodwin, line 6: use `import type` then to ensure it is not included in the physical module.

test/eventbus.test.ts
```diff
@@ -271,6 +271,71 @@ test("map narrows type and pipe enforces", () => {
...
+      .when(
+        stack,
+        "rule",
+        (event): event is EventBusRuleInput<t1> => event.detail.type === "one"
+      )
+      .map((event) => event)
+      // @ts-expect-error
+      .pipe(bus)
+  ).toThrow();
+});
```
sam-goodwin, line 341: This is a beautiful test. I love how it has `@ts-expect-error` and `expect().toThrow()`. Total parity between runtime and compile time logic - the types properly describe valid and invalid states.

sam-goodwin (approved):
Approve - please merge after responding to latest comments.

src/event-bridge/transform.ts
thantos: It can be, though it doesn't return a wrapped type.

sam-goodwin: Curious for we will configure alarms?

src/function.ts
sam-goodwin: I add them so that links in docs can still be used

src/event-bridge/transform.ts
thantos: soo... there are no monitoring helpers on the event bus CDK constructs.

The metrics on event bus are not great, mostly limited to the bus or rule level.

https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-monitoring.html

So right now monitoring would be on:
1. The source bus or rule
2. The target
3. the dead letter queue

src/function.ts
thantos: Was this meant to be on a different line? Looks like eslint may have moved it.

src/function.ts
```diff
@@ -39,26 +42,52 @@ export function isFunction<P = any, O = any>(a: any): a is IFunction<P, O> {
...
+  extends Omit<aws_events_targets.LambdaFunctionProps, "event"> {}
+
 export type FunctionClosure<P, O> = (
   payload: P,
   context: Context
 ) => Promise<O>;
 
-export interface IFunction<P, O> {
+/**
+ * Wraps an {@link aws_lambda.Function} with a type-safe interface that can be
```
sam-goodwin, line 54: Is it this interface that is doing the wrapping? Don't think that makes sense?

src/function.ts
```diff
@@ -140,6 +169,19 @@ abstract class FunctionBase<P, O>
...
+
+  public readonly eventBus = makeEventBusIntegration<
+    P,
+    FunctionEventBusTargetProps | undefined
+  >({
+    target: (props, targetInput) =>
+      new aws_events_targets.LambdaFunction(this.resource, {
+        deadLetterQueue: props?.deadLetterQueue,
+        maxEventAge: props?.maxEventAge,
+        retryAttempts: props?.retryAttempts,
```
sam-goodwin, line 181: Nit: 
```ts
...props
```

More concise and automatically accounts for the case when new fields are added.

src/integration.ts
```diff
@@ -144,6 +173,14 @@ export class IntegrationImpl<F extends AnyFunction = AnyFunction>
     return this.unhandledContext(context.kind);
   }
 
+  public get eventBus(): EventBusTargetIntegration<any, any> {
+    if (this.integration.eventBus) {
+      return this.integration.eventBus;
+    }
+
+    return this.unhandledContext("EventBus");
```
sam-goodwin, line 181: This boilerplate pattern keeps occurring

sam-goodwin (approved):

src/event-bridge/transform.ts
sam-goodwin: Are these exposed on the return type of `.pipe`?

src/function.ts
sam-goodwin: Ah, eslint probably sorted and left it where it was.

src/function.ts
thantos: Wonder if this was a merge issue. I don't remember this being there.

src/event-bridge/transform.ts
thantos: `pipe` doesn't return anything right now because I didn't have a use case for it yet.

We could return:
1. the transform (`.pipe().pipe().rule`)
2. the rule (`.pipe().map().pipe()`)
3. the CDK IRuleTarget
4. a new Target object (what would this have?)
5. nothing

I guess the transform makes the most sense. Pipe itself doesn't really generate a resource, it adds a target to a rule and there is nothing you could really do with that target except for add it to another rule (which wouldn't be type safe and targets are cheap; aka not much value in reuse).  

I think I'll leave this for now as I don't see much utility in any of the use cases.

sam-goodwin (approved):

PR #161, opened 2022-05-27
feat(event-bridge): schedule, whenAny, rename, when without scope, docs

thantos:
Closes #142, closes #89, closes #140, closes #147 

* [x] Add `schedule`'d rules #140 
  * Can only create schedules rules on the default bus
  * `EventBus.default().schedule(scope, id, schedule, props?)` - creates a scheduled rule on the default bus
  * `EventBus.schedule(scope, id, schedule, props?)` - creates a scheduled rule on the default bus
* [x] Add `all` catch all rule on the `EventBus` #89 
  * `bus.all()` - Uses a singleton rule on the event bus, creating a rule call `all`.
  * `bus.all(scope, id)` - creates a new rule with the scope and id 
* [x] Support `.pipe` escape hatch for unsupported integrations. `bus.when().map().pipe((targetInput) => aws_event_targets.LambdaFunction(func, { event: targetInput }))`
* [x] Rename 
    * `EventtBusRule` ->`Rule`
    * `EventBusTransform` -> `EventTransform`
    * `EventBusRuleInput` -> `Event`
* [x] Support `.when` without `scope` parameter (`when(id, () => {}`)
   * *Why not support removing `id` (second param)?* While possible (I started to do this), we don't want to auto generate CDK resource IDs as small changes would force recreation of the rules, which may disrupt service. This isn't a problem for scope because we always use the bus the rule was created on.
* [x] Update the integration for putting events to a bus from `bus(event)` to `bus.putEvents(event)`.
* [x] Update docs #147 
   * [x] Add `schedule`
   * [x] Add `whenAny`
   * [x] Add new `when` pattern
   * [x] Reflect renaming
   * [x] Review all docs and add
   * [x] Integrations

```ts
interface MyEventDetails {
   name: string;
}
interface MyEvent extends Event<MyEventDetails, 'myDetailType', 'mySource'>{}
interface LambdaEvent extends Event<{}, string, 'lambda'>{}

const bus = new EventBus<LambdaEvent  | MyEvent>(stack, 'bus');

const allEvents = bus.all();
// transform and then send to lambda
allEvents.map(event => event.id).pipe(new Function(stack, 'func', async (id) => console.log(id)));

const lambdaEvents = bus
   // when the source is lambda
   .when(event => event.source === "lambda")
   // bus to bus
   .pipe(new EventBus());

const logGroup = new aws_logs.LogGroup(this, 'MyLogGroup', {
  logGroupName: 'MyLogGroup',
});

[...]

PR #179, opened 2022-05-31
feat(lambda,event-bridge): lambda event sources

### LINKED ISSUE #44: Event Sources
Add auto event bus rule generation from integrations with the ability.

Some AWS services like Lambda, Step Functions, and more automatically generate Rules on the default bus.

Add the ability to automatically create these rules and then send to all Functionless Event Bus target. They should support type safe event types based on the event.

```ts
myFunction.onSucccess().send(myMachine, event => event.detail.result);
myMachine.onStarted().send(...)
```

Where `send` is the outcome of #37 's final target design.

thantos:
Closes #44 

Depends on #161 - This change made major changes to EventBus, stacking these changes to reduce churn.

* [x] Lambda
* [x] Step Functions - Already done
❌ Api Gateway
❌ App Sync
❌ Dynamo

 ## Lambda

Unlike Step Functions, Lambda does not emit events by default. Lambda has a concept of a `destination` or `async invocation`. Each lambda, version, and alias can only have a single onSuccess and onFailure destination. Destinations can be EventBuses, Sns topics, Sqs queues, or other lambda functions.

```ts
const func = new Function(stack, 'func', {
   onSuccess: bus, // IEventBus | IFunction | aws_lambda.IDestination
   onFailure: bus // IEventBus | IFunction | aws_lambda.IDestination
});

// fail when onSuccess is not set and not a bus?
func.onSuccess(bus).map().pipe();
func.onFailure(bus).map().pipe();
```


src/function.ts
```diff
@@ -140,6 +208,56 @@ abstract class FunctionBase<P, O>
...
+  ): Rule<AsyncResponseFailureEvent<P>> {
+    return new PredicateRuleBase<AsyncResponseFailureEvent<P>>(
+      bus.bus,
+      id,
+      bus,
+      /**
+       * when(event => event.source === "lambda"
+       *    && event["detail-type"] === "Lambda Function Invocation Result - Failure"
+       *      && event.resources.includes(this.resource.functionArn))
+       */
```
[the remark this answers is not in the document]
thantos: yes :(

src/event-bridge/rule.ts
```diff
@@ -138,8 +138,8 @@ export interface IRule<T extends Event> {
    */
   pipe(props: LambdaTargetProps<T>): void;
   pipe(func: IFunction<T, any>): void;
-  pipe(bus: IEventBus<T>): void;
-  pipe(props: EventBusTargetProps<T>): void;
+  pipe(bus: IEventBus<any>): void;
+  pipe(props: EventBusTargetProps<any>): void;
```
sam-goodwin, line 142: Why `any`? Wouldn't that break type safety?

src/function.ts
```diff
@@ -52,8 +83,112 @@ export interface IFunction<P, O> {
...
+   * and that destination must match the bus provided here.
+   *
+   * @see https://docs.aws.amazon.com/lambda/latest/dg/invocation-async.html#invocation-async-destinations
+   *
+   * The rule returned will contain the logic:
+   *
+   * when(id, event => event.source === "lambda"
+   *    && event["detail-type"] === "Lambda Function Invocation Result - Success"
+   *      && event.resources.includes(this.resource.functionArn))
+   */
```
sam-goodwin, line 101: Surround with triple backticks block. Tsdoc produces markdown

src/function.ts
```diff
@@ -140,6 +275,78 @@ abstract class FunctionBase<P, O>
...
+
+  configureAsyncInvoke(config: EventInvokeConfigOptions<P, O>): void {
+    this.resource.configureAsyncInvoke({
+      ...config,
+      onSuccess: FunctionBase.normalizeAsyncDestination<P, O>(config.onSuccess),
+      onFailure: FunctionBase.normalizeAsyncDestination<P, O>(config.onFailure),
+    });
+  }
+
+  onSuccess(
```
sam-goodwin, line 301: Does this inherit docs?

src/function.ts
```diff
@@ -164,6 +374,9 @@ export interface FunctionProps
   clientConfigRetriever?: (
     clientName: ClientName | string
   ) => Omit<AWS.Lambda.ClientConfiguration, keyof AWS.Lambda.ClientApiVersions>;
+
+  onSuccess?: FunctionAsyncOnSuccessDestination<P, O>;
+  onFailure?: FunctionAsyncOnFailureDestination<P>;
```
sam-goodwin, line 379: Missing docs

src/function.ts
```diff
@@ -140,6 +275,78 @@ abstract class FunctionBase<P, O>
       ResultSelector: "$.Payload",
     };
   }
+
+  protected static normalizeAsyncDestination<P, O>(
```
sam-goodwin, line 279: Why protected?

src/function.ts
```diff
@@ -140,6 +275,78 @@ abstract class FunctionBase<P, O>
...
+    return destination === undefined
+      ? undefined
+      : isEventBus<any>(destination)
+      ? new EventBridgeDestination(destination.bus)
+      : isFunction(destination)
+      ? new LambdaDestination(destination.resource)
+      : destination;
+  }
+
+  configureAsyncInvoke(config: EventInvokeConfigOptions<P, O>): void {
```
sam-goodwin, line 293: Should this be public or private? I didn't see it in the interface so private?

src/function.ts
sam-goodwin: No was just curious if it was happpening implicitly?

thantos: Yeah, it inherits from the interface.

src/function.ts
thantos: This is an abstract base class, this method is used by the base and child class. And we don't need to expose this.

src/function.ts
thantos: Its in the interface, made it public

image

thantos: Added an es-lint rule to error on these cases to make your life easier.

thantos:
Note: prettier is broken due to the use of typescript 4.7

Waiting on: https://github.com/prettier/prettier/issues/12640

src/function.ts
```diff
@@ -44,17 +57,154 @@ export type FunctionClosure<P, O> = (
...
 
-export interface IFunction<P, O> {
+/**
+ * Returns the payload type on the {@link IFunction}.
+ */
+export type FunctionPayload<F extends IFunction<any, any>> = [F] extends [IFunction<infer P, any>] ? P : never;
+/**
+ * Returns the output type on the {@link IFunction}.
+ */
+ export type FunctionOutput<F extends IFunction<any, any>> = [F] extends [IFunction<any, infer O>] ? O : never;
```
sam-goodwin, line 67: nit: indentation

sam-goodwin: why the `[F] extends `? What's the purpose of packing it a 1-tuple?

src/function.ts
```diff
@@ -44,17 +57,154 @@ export type FunctionClosure<P, O> = (
...
+   *
+   * The rule returned will contain the logic:
+   *
+   * ```ts
+   * when(id, event => event.source === "lambda"
+   *    && event["detail-type"] === "Lambda Function Invocation Result - Success"
+   *      && event.resources.includes(this.resource.functionArn))
+   * ```
+   */
+  onSuccess(
```
sam-goodwin, line 115: nit: looks like eslint missed this missing `public`

src/function.ts
```diff
@@ -44,17 +57,154 @@ export type FunctionClosure<P, O> = (
...
   (...args: Parameters<ConditionalFunction<P, O>>): ReturnType<
     ConditionalFunction<P, O>
   >;
+
+  /**
+   * Event Source for the {@link Function} onSuccess async invocation destination.
+   *
+   * For Lambda, the onSuccess destination must first be configured via either
+   * the {@link Function} constructor or by using {@link IFunction.configureAsyncInvoke}
+   * and that destination must match the bus provided here.
```
sam-goodwin, line 103: I'm not sure what that means and how to take action on it.

src/function.ts
thantos: `[F] extends []` changes the way that union types are interpreted.

https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types

With `[F]`

image

Without

image

Which in the case its being used in... if the types are not blown out (`Function<A | B>`) then the `isFunction` does not narrow the types correctly. 

```ts
if(isFunction(destination)) {
    destination // is IFunction<AsyncResponseSuccess<P, O> | AsyncResponseFailure<P>, any>
} else {
     destination // is IDestination | IFunction<AsyncResponseSuccess<P, O> | AsyncResponseFailure<P>, any>
}
```

image

But with... we get nice type narrowing.

```ts
if(isFunction(destination)) {
    destination // is IFunction<AsyncResponseSuccess<P, O>, any> | IFunction<AsyncResponseFailure<P>, any>
} else {
     destination // is IDestination
}
```

image

src/function.ts
thantos: Thats the interface :-)

src/function.ts
sam-goodwin: Awesome analysis. Thanks!

src/function.ts
sam-goodwin: Ahhhh my bad. Thumbs down feels bad lol

src/event-bridge/event-bus.ts
```diff
@@ -40,7 +40,12 @@ export const isEventBus = <E extends Event>(v: any): v is IEventBus<E> => {
   );
 };
 
-export interface IEventBusFilterable<E extends Event> {
+/**
+ * Returns the {@link Event} type on the {@link EventBus}.
+ */
+export type EventBusEvent<B extends IEventBus<any>> = [B] extends [IEventBus<infer E>] ? E : never;
```
sam-goodwin, line 46: Should this be called GetEventBusType? Or else it could be confused with the Event type. The verb helps with understanding

src/event-bridge/event-bus.ts
```diff
@@ -115,16 +120,16 @@ export interface IEventBusFilterable<E extends Event> {
   ): Rule<E, O>;
 }
 
-export interface IEventBus<E extends Event = Event>
+export interface IEventBus<in E extends Event = Event, O extends E = E>
```
sam-goodwin, line 123: Might want to document the type parameters. Not immediately obvious what E and O mean

src/event-bridge/event-bus.ts
```diff
@@ -339,7 +339,7 @@ abstract class EventBusBase<E extends Event> implements IEventBus<E> {
   }
 
   public readonly eventBus = makeEventBusIntegration<
-    E,
+    EventBusPutEventInput<O>,
```
sam-goodwin, line 342: A bit verbose.

src/function.ts
```diff
@@ -64,13 +116,153 @@ export interface IFunction<P, O>
...
+  /**
+   * Event Source for the {@link Function} onSuccess async invocation destination.
+   *
+   * For Lambda, the onSuccess destination is not enabled by default.
+   * It must first be configured via either the {@link Function} constructor
+   * or by using {@link IFunction.configureAsyncInvoke} and that destination must match the bus provided here.
+   *
+   * ```ts
+   * const bus = new EventBus(stack, 'bus');
+   * new Function(stack, 'func', { onSuccess: bus }, async () => {});
```
sam-goodwin, line 128: Cool!

src/function.ts
```diff
@@ -64,13 +116,153 @@ export interface IFunction<P, O>
...
+   * ```
+   *
+   * or
+   *
+   * ```ts
+   * const bus = new EventBus(stack, 'bus');
+   * const func = new Function(stack, 'func', async () => {});
+   * // if onSuccess or onFailure is already set, this will fail.
+   * func.configureAsyncInvoke({ onSuccess: bus });
```
sam-goodwin, line 137: Hmm configureAsyncInvoke isn't telling me exactly what it means. This is enabling async invoke and routing events to the bus? Maybe `enableAsyncInvoke`?

sam-goodwin (approved):

src/event-bridge/event-bus.ts
sam-goodwin: In standard libraries, I usually prefer to use words for type parameters instead of symbols.

src/function.ts
thantos: This is the name from the CDK, which I agree, is also confusing..will take another pass.

sam-goodwin:
Should the title of the PR be more granular? Would someone reading the CHANGELOG know what this change is about based purely on the title?

thantos:
> Should the title of the PR be more granular? Would someone reading the CHANGELOG know what this change is about based purely on the title?

Ha... the change started with higher ambitions and then I realized that no one else but lambda and step function publish events!

Will update.

src/event-bridge/event-bus.ts
thantos: Thanks, that was a helpful exercise. 

```ts
/**
 * @typeParam Evnt - the union type of events that this EventBus can accept.
 *                   `Evnt` is the covariant version of `OutEnvt` in that
 *                   the bus will accept any of `Evnt` while the EventBus can
 *                   emit any of `OutEvnt`.
 * @typeParam OutEvnt - the union type of events that this EventBus will emit through rules.
 *                      `OutEvnt` is the contravariant version of `Envt` in that
 *                      the bus will emit any of `OutEvnt` while the EventBus can
 *                      can accept any of `Evnt`. This type parameter should be left
 *                      empty to be inferred.
 */
```

src/event-bridge/rule.ts
```diff
@@ -21,11 +21,15 @@ import type { Event } from "./types";
...
+export type RulePredicateFunction<Evnt, OutEvnt extends Evnt = Evnt> =
+  | ((event: Evnt) => event is OutEvnt)
+  | ((event: Evnt) => boolean);
 
-export interface IRule<T extends Event> {
+/**
+ * @typeParam - Evnt - The original event type from the {@link EventBus}.
+ * @typeParam - OutEvnt - The narrowed event type after the predicate is applied.
+ */
+export interface IRule<in Evnt extends Event, out OutEvnt extends Evnt> {
```
sam-goodwin, line 32: Why do we need two type parameters?

src/event-bridge/event-bus.ts
```diff
@@ -103,28 +108,42 @@ export interface IEventBusFilterable<E extends Event> {
...
    */
-  when<O extends E>(
+  when<InEvnt extends OutEvnt, NewEvnt extends InEvnt>(
     id: string,
-    predicate: RulePredicateFunction<E, O>
-  ): Rule<E, O>;
-  when<O extends E>(
+    predicate: RulePredicateFunction<InEvnt, NewEvnt>
+  ): Rule<InEvnt, NewEvnt>;
+  when<InEvnt extends OutEvnt, NewEvnt extends InEvnt>(
```
sam-goodwin, line 119: These two parameters seem overly complex and redundant. Why did you add them?

src/event-bridge/event-bus.ts
```diff
@@ -103,28 +108,42 @@ export interface IEventBusFilterable<E extends Event> {
...
-  ): Rule<E, O>;
+    predicate: RulePredicateFunction<InEvnt, NewEvnt>
+  ): Rule<InEvnt, NewEvnt>;
 }
 
-export interface IEventBus<E extends Event = Event>
-  extends IEventBusFilterable<E>,
+/**
+ * @typeParam Evnt - the union type of events that this EventBus can accept.
+ *                   `Evnt` is the covariant version of `OutEvnt` in that
```
sam-goodwin, line 128: `Evnt` is contra-variant, not co-variant. `out` is for covariance

src/event-bridge/event-bus.ts
```diff
@@ -103,28 +108,42 @@ export interface IEventBusFilterable<E extends Event> {
...
+ *                   the bus will accept any of `Evnt` while the EventBus can
+ *                   emit any of `OutEvnt`.
+ * @typeParam OutEvnt - the union type of events that this EventBus will emit through rules.
+ *                      `OutEvnt` is the contravariant version of `Evnt` in that
+ *                      the bus will emit any of `OutEvnt` while the EventBus can
+ *                      can accept any of `Evnt`. This type parameter should be left
+ *                      empty to be inferred. ex: `EventBus<Event<Detail1> | Event<Detail2>>`.
+ */
+export interface IEventBus<in Evnt extends Event = Event, OutEvnt extends Evnt = Evnt>
```
sam-goodwin, line 137: This doesn't seem right. I think we can remove the `OutEvnt`

src/event-bridge/rule.ts
thantos: `Evnt` is the event in. The input to the predicate on the rule.

`OutEvnt` is the narrowed type as the input to `pipe` or `map` or `when`.

Essentially splitting and showing the two sides of the predicate.

sam-goodwin: Why do you need `Evnt`?

src/event-bridge/event-bus.ts
thantos: Need to invert the `in`Evnt` to `out NewEvnt` and `out OutEvnt` to `in InEvnt` for the when method.

src/event-bridge/rule.ts
thantos: Its being used on `integration: IntegrationWithEventBus<Evnt, Props>,`

But that doesn't seem right... will look.

sam-goodwin (approved):
We can modify the types in a subsequent change. Ship it

thantos:
> We can modify the types in a subsequent change. Ship it

Removed the unessecary `in Evnt` from `IRule` and `RuleBase` which cleaned up a bit.

Also fixed the covariance/contravariant naming issues and think I understand them better now :-)

https://dmitripavlutin.com/typescript-covariance-contravariance/

thantos:
> Error: Cannot find module '/opt/build/repo/scripts/compile-error-code-page'

I don't understand what changed

PR #180, opened 2022-06-01
feat: provide link to documentation for compile time error codes

### LINKED ISSUE #176: Generate documentation page for error code
We are introducing error codes for the Functionless compiler. These errors are displayed in the IDE via the language service plugin and `functionless` CLI, and thrown during synth time (if they are not caught earlier).

We need to standardize how we describe errors, including a numeric error code, description and link to functionless.org.

> Cannot perform arithmetic on variables in Step FunctionFunctionless(100). See https://functionless.org/error-codes#100

- [x] Each error code should be easily searchable on google and a link to documentation on functionless.org should be included in the error that can be clicked within VS code or the CLI.

To achieve this, I suggest we define all errors in a constants file, e.g. `error-code.ts` from which we will generate a markdown file under `website/docs/error-codes.md` containing a searchable and linkable description of each error.

sam-goodwin:
Fixes #176 

Adding an `error-code.ts` namespace where we will store all error code constants. Each error code has an ID, `code`, `messageText` and documentation in TypeScript code. 

```ts
export namespace ErrorCodes {
  /**
   * The computations that [Amazon States Language](https://docs.aws.amazon.com/step-functions/latest/dg/concepts-amazon-states-language.html)
   * can do is restricted by JSON Path and the limited [Intrinsic Functions](https://docs.aws.amazon.com/step-functions/latest/dg/amazon-states-language-intrinsic-functions.html). Currently, arithmetic expressions are not supported.
   * ```ts
   * // ok
   * new StepFunction(scope, id, () => 1 + 2);
   *
   * // illegal!
   * new StepFunction(scope, id, (input: {num: number}) => input.number + 1);
   * ```
   */
  export const Cannot_perform_arithmetic_on_variables_in_Step_Function: ErrorCode =
    {
      code: 100,
      messageText: "Cannot perform arithmetic on variables in Step Function",
    };
}
```

TS doc generates a markdown file under `website/docs/api/namespaces/ErrorCodes.md`. This file is not easy to read and hidden away in the file tree, so a script is used to pre-process the file and create `website/docs/error-codes.md` at the top-level of the file tree and in a linkable and easy to read format.

image

All errors reported by the language service or CLI now provide a link directly to this documentation explaining the error.

image

- [ ] Make the link clickable within VS code (current not sure if this is possible)

### DIFF
--- a/scripts/compile-error-code-page.js
+++ b/scripts/compile-error-code-page.js
@@ -3,6 +3,10 @@ const { ErrorCodes } = require("../lib/error-code");
 const fs = require("fs");
 const path = require("path");
 
+/**
+ * Generate `functionless.org/docs/error-codes file.
+ */
+
 const errorCodesAPIReference = path.join(
   __dirname,
   "..",
@@ -21,35 +25,31 @@ const errorCodeDocumentationPath = path.join(
   "error-codes.md"
 );
 
-(async function () {
-  let errorCodeMarkdown = (await fs.promises.readFile(errorCodesAPIReference))
-    .toString()
-    .replace("## Variables\n", "# Error Codes\n")
-    .replace(
-      `id: "ErrorCodes"
+let errorCodeMarkdown = fs
+  .readFileSync(errorCodesAPIReference)
+  .toString()
+  .replace("## Variables\n", "# Error Codes\n")
+  .replace(
+    `id: "ErrorCodes"
 title: "Namespace: ErrorCodes"
 sidebar_label: "ErrorCodes"
 sidebar_position: 0
 custom_edit_url: null`,
-      `title: "Error Codes"
+    `title: "Error Codes"
 sidebar_position: 3`
-    )
-    .replace(/• `Const`.*\n/g, "")
-    .replace(/\n\n\n/g, "\n\n")
-    .replace(/\n#### Defined in.*\n\n.*error.*\n/g, "");
+  )
+  .replace(/• `Const`.*\n/g, "")
+  .replace(/\n\n\n/g, "\n\n")
+  .replace(/\n#### Defined in.*\n\n.*error.*\n/g, "");
 
-  for (const [errorId, errorCode] of Object.entries(ErrorCodes)) {
-    //### Cannot\_perform\_arithmetic\_on\_variables\_in\_Step\_Function
-    errorCodeMarkdown = errorCodeMarkdown.replace(
-      `### ${errorId.replace(/_/g, "\\_")}`,
-      `### ${errorCode.messageText}
+for (const [errorId, errorCode] of Object.entries(ErrorCodes)) {
+  //### Cannot\_perform\_arithmetic\_on\_variables\_in\_Step\_Function
+  errorCodeMarkdown = errorCodeMarkdown.replace(
+    `### ${errorId.replace(/_/g, "\\_")}`,
+    `### ${errorCode.messageText}
 
 __Error Code__: Functionless(${errorCode.code})`
-    );
-  }
+  );
+}
 
-  await fs.promises.writeFile(errorCodeDocumentationPath, errorCodeMarkdown);
-})().catch((err) => {
-  console.error(err);
-  process.exit(1);
-});
+fs.writeFileSync(errorCodeDocumentationPath, errorCodeMarkdown);
--- a/scripts/compile-error-code-page.js
+++ b/scripts/compile-error-code-page.js
@@ -1,0 +1,55 @@
+const { ErrorCodes } = require("../lib/error-code");
+
+const fs = require("fs");
+const path = require("path");
+
+const errorCodesAPIReference = path.join(
+  __dirname,
+  "..",
+  "website",
+  "docs",
+  "api",
+  "namespaces",
+  "ErrorCodes.md"
+);
+
+const errorCodeDocumentationPath = path.join(
+  __dirname,
+  "..",
+  "website",
+  "docs",
+  "error-codes.md"
+);
+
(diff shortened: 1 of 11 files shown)


sam-goodwin:
@thantos @brendo-m  after this change is complete, please adopt this standard and contribute to the error codes. We should begin sharing al error codes throughout both the TS code phase and CDK synth phase

website/docs/error-codes.md
```diff
@@ -0,0 +1,21 @@
+---
```
sam-goodwin, line 1: This file is generated. Perhaps I should not include it in Git?

website/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "docusaurus": "docusaurus",
-    "start": "docusaurus start",
+    "start": "LOCAL=true; docusaurus start",
```
sam-goodwin, line 7: Not using this right now but I assume we will need some way to specify environment so development points to localhost

scripts/compile-error-code-page.js
```diff
@@ -0,0 +1,55 @@
...
+title: "Namespace: ErrorCodes"
+sidebar_label: "ErrorCodes"
+sidebar_position: 0
+custom_edit_url: null`,
+      `title: "Error Codes"
+sidebar_position: 3`
+    )
+    .replace(/• `Const`.*\n/g, "")
+    .replace(/\n\n\n/g, "\n\n")
+    .replace(/\n#### Defined in.*\n\n.*error.*\n/g, "");
```
sam-goodwin, line 39: I went with regex replacements for a quick and dirty solution. Works well. Would be interesting to work programmatically with the tsdoc API though, if that's possible.

src/error-code.ts
```diff
@@ -0,0 +1,23 @@
+export interface ErrorCode {
+  code: number;
+  messageText: string;
+}
+
+export namespace ErrorCodes {
```
thantos, line 6: Why a namespace and not an object or enum?

```ts
export const ErrorCodes <Record<string, ErrorCode>>{
    /**
     * ...
     */
    Cannot_perform_arithmetic_on_variables_in_Step_Function: {
      code: 100,
      messageText: "Cannot perform arithmetic on variables in Step Function",
    }
}
``` 

Comments and intellisense seem to work

image

I ask because a namespace creates a functional closure and will bloat the lambda functions and other code. Unless we need to make use of the functional closure and will have more than just error codes. Also less verbose.

src/error-code.ts
```diff
@@ -0,0 +1,23 @@
...
+  /**
+   * The computations that [Amazon States Language](https://docs.aws.amazon.com/step-functions/latest/dg/concepts-amazon-states-language.html)
+   * can do is restricted by JSON Path and the limited [Intrinsic Functions](https://docs.aws.amazon.com/step-functions/latest/dg/amazon-states-language-intrinsic-functions.html). Currently, arithmetic expressions are not supported.
+   * ```ts
```
thantos, line 10: I have been adding workarounds to the "Limitations" pages for event bus. Should errors/limitations include how (easy it is) to work around the limitation?

src/validate.ts
```diff
@@ -76,13 +76,19 @@ export function validate(
...
   function newError(
     invalidNode: typescript.Node,
-    code: number,
-    messageText: string
+    error: ErrorCode,
+    messageText?: string
   ): ts.Diagnostic {
+    const baseUrl = process.env.FUNCTIONLESS_LOCAL
+      ? `http://localhost:3000`
+      : `https://functionless.org`;
```
thantos, line 84: Inject the base url instead of the flag?

website/package.json
thantos: you may be able to get flags from webpack.

https://webpack.js.org/guides/environment-variables/

try `!process.env.production`

brendo-m: shouldn't this be `FUNCTIONLESS_LOCAL` to match above?

src/error-code.ts
sam-goodwin: Good point, we should document workarounds for each of the error codes 👍

src/validate.ts
sam-goodwin: Then every caller would have to pass it?

src/error-code.ts
```diff
@@ -0,0 +1,23 @@
...
+   * // ok
+   * new StepFunction(scope, id, () => 1 + 2);
+   *
+   * // illegal!
+   * new StepFunction(scope, id, (input: {num: number}) => input.number + 1);
+   * ```
+   */
+  export const Cannot_perform_arithmetic_on_variables_in_Step_Function: ErrorCode =
+    {
+      code: 100,
```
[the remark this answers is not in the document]
sam-goodwin: Opened an issue to track this discussion. https://github.com/functionless/functionless/issues/182

sam-goodwin:
> Have an example of how to use this in the compile and synth code?

Here is an example within `validate.ts`

```ts
function validateStepFunctionNode(
  node: typescript.Node
): typescript.Diagnostic[] {
  if (
    (ts.isBinaryExpression(node) &&
      isArithmeticToken(node.operatorToken.kind) &&
      !checker.isConstant(node)) ||
    (ts.isPrefixUnaryExpression(node) && !checker.isConstant(node))
  ) {
    return [
      newError(
        node,
        ErrorCodes.Cannot_perform_arithmetic_on_variables_in_Step_Function
      ),
    ];
  }
  return [];
}
```

Have not thrown any of these errors during synth yet. Need to do that. See https://github.com/functionless/functionless/issues/183

PR #185, opened 2022-06-01
fix: explicitly make EventBus event parameter as contra-variant

sam-goodwin:


src/event-bridge/rule.ts
```diff
@@ -104,7 +104,7 @@ export interface IRule<T extends Event> {
    * Unsupported by Functionless:
    * * Variables from outside of the function scope
    */
-  map<P>(transform: EventTransformFunction<T, P>): EventTransform<T, P>;
+  map<E extends T, P>(transform: EventTransformFunction<E, P>): EventTransform<E, P>;
```
sam-goodwin, line 107: This was the hard part to figure out. If you just use `T`, you get contra vs co-variant errors. This doesn't seem 100% right but it makes it all compile and infer.

thantos (approved):

test/function.localstack.test.ts
```diff
@@ -242,6 +242,8 @@ localstackTestSuite("functionStack", (testResource, _stack, _app) => {
 
   testFunctionResource(
     "Call Lambda throw error",
+    // @ts-ignore - contra-variance on IEventBus<in T> makes Function<_, never> fail to type check because never can't be a super type
+    //              IEventBus is relevant because of Function.onSuccess takes in an IEventBus
```
sam-goodwin, line 246: Oh do we still need this?

src/event-bridge/rule.ts
```diff
@@ -28,7 +28,7 @@ export type RulePredicateFunction<E, O extends E = E> =
   | ((event: E) => event is O)
   | ((event: E) => boolean);
 
-export interface IRule<T extends Event> {
+export interface IRule<in T extends Event> {
```
sam-goodwin, line 31: I feel like this is wrong. Rule produces a value, not accepts one

Issue #119, opened 2022-05-02
AppSync: Schema and code-first reflection from typescript types.


thantos:
Functionless should support 100% native typescript AppSync/GraphQL generation

Inspired by 
* https://typegraphql.com/'s API.
* CDK Code First AppSync - https://docs.aws.amazon.com/cdk/api/v1/docs/aws-appsync-readme.html#code-first-schema

```ts
interface Comment {
   postId: string,
   commentId: string,
   message: string
}

interface CommentPage {
   nextToken?: string,
   comments: Comments[]
}

class Post {
  constructor(readonly postId: string, title: string) {}

  comments(): CommentPage {
     const page = database.query(... dynamo stuff ...)

     return {
        nextToken: ...,
        comment: page.Items // transform
     }
  }
}

const api = new functionless.AppSyncApi();
const postType = api.type<Post>();
// Add a query function
api.query(function getPost(id: string) { database.getItem(...) });
// or 
api.query("getPost", (id: string) => { database.getItem(...) });

// support mutating already registered types. 
postType.query("addComment", (comment: Omit<Comment, "commentId">) => {
   database.putItem(...);
})

// also support wrapping schemas without trying to an api
const schema = new functionless.Schema().type<Post>; // new aws_appsync.GraphQlApi(..., {  schema });
// or static
functionless.Schema.type<Post>();
```

Improvements over TypeGraphQL

1. No annotation duplication: a field can be be undefined or null is options, else required.
2. `Readonly` fields cannot be  mutated, all else can
3. Support from interfaces?

Problems:

1. these classes are not usable outside of graphql, maybe they should be wrapped in a closure?

sam-goodwin:
Consider integrating with https://github.com/hayes/pothos.

thantos:
https://deepkit.io/documentation/type/reflection

PR #202, opened 2022-06-06
feat: update resource callbacks to support async and most integrations to return async

### LINKED ISSUE #105: Inconsistent use of Promises
Most of the existing integrations assume that integrations are sync, but this assumption doesn't work when the integration has the ability to be sync or async and we should not be opinionated. 

### Step Function

```ts
const bus = new EventBus();
// Makes a sync call to event bus within step functions.
new StepFunction(this, '', () => bus());
```

### Lambda

Part of #103

```ts
const bus = new EventBus();
// Is the call to bus sync or async?
new Function(this, '', async () => bus());

// Is the call to bus sync or async?
new Function(this, '', async () => await bus());
```

In lambda, the output would be a client call to event bus which would be a promise or callback in the actual lambda code

```ts
const EB = new EventBus() // client
const handler = async () => {
  // puts an event, but does not wait for success, cannot throw an error
   EB.putEvent(...);
  // put an event, waits for success, can throw an error
   await EB.putEvent(...);
}
```

## Ambiguous Calls

Another issue is that the integration calls on resources are generally not returning promises, so even if we wanted to support `await bus()` it wouldn't make sense as `bus = EventBus<Input, Output>` returns `Output` and not `Promise<Output>`.

|                  | Integration returns promise (in) | callback supports promise (out) | service supports async (out) | callback supports sync (out) |
| ---            |   --- | ---  | --| --|
| Event Bus | No | Yes (`pipe` is always async) | NA | NA |
| Step Functions | No | No, callback doesn't support async | Yes, with Request/Response integration type | Yes, with Run Job integration type |
| Function | No | Yes (callback as of #103 must be async) | Yes | Yes |
| App Sync | No | No | No? | Yes |

### Option 1 - Make all callbacks (out) and integrations (in) support async and promises. Fail when a combination is not supported.

```ts
// update event bus signature from
(event: Partial<E>, ...events: Partial<E>[]): void;
// to
(event: Partial<E>, ...events: Partial<E>[]): Promise<void>;
```

```ts
// ?? - Could be supported if we treat this the same as the await case, a thrown error would break the contract
new AppSyncResolver(async () => bus())
// Success - put is used synchronously 
new AppSyncResolver(async () => await bus())

// Success - machine is started, but no response is waited for
// Caveat - contract is not the same as the user may want to get the http response from startExecution (a util could do this)
new AppSyncResolver(async () => anotherStepFunction())
// Success - machine is started, we wait for the execution to complete
new AppSyncResolver(async () => await anotherStepFunction())
```

```ts
// the behavior of both of these would be the same, step functions waits for the http response and then continues on.
// Caveat - the contract is not exactly the same as typescript, where no response is waited for
new StepFunction(async () => bus())
new StepFunction(async () => await bus())

// Success - machine is started, but no response is waited for, using the request/response integration
// Caveat - contract is not the same as the user may want to get the http response from startExecution (a util could do this)
new StepFunction(async () => anotherStepFunction())

// ?? - This would take additional work, including supporting storing and polling for sfn executions.
new StepFunction(async () => { 
   const ex = expressStepFunction();
   const ex2 = standardStepFunction();
   // result
   const exResult = await ex;
   // http 200
   const ex2Result = await ex2:
} )

// Success - machine is started, we wait for the execution to complete
new StepFunction(async () => await expressStepFunction())
// Success - machine is started, we wait for the http 200
new StepFunction(async () => await standardStepFunction())
```

```ts
// new EventBus.putEvent()
new Function(..., async () => bus())
// await new EventBus.putEvent()
new Function(..., async () => await bus())
```

### Option 2 - Only support sync

```ts
// Puts event and waits for http 200
new AppSyncResolver(() => bus())

// starts machine and waits for completion
new AppSyncResolver(() => expressStepFunction())
// starts machine and waits for http 200
new AppSyncResolver(() => standardStepFunction())
```

```ts
// waits for http 200
new StepFunction(() => bus())

// starts machine and waits for completion
new StepFunction(() => expressionStepFunction())
// starts machine and waits for http 200
new StepFunction(() => standardStepFunction())
```

```ts
// assume we want all integrations to behave in a sync way
// await new EventBus.putEvent()
new Function(..., async () => bus())

// not a promise
new Function(..., async () => await bus())
```

### Option 3 - Special function for async behavior

```ts
interface IEventBus {
  async(event: Partial<E>, ...events: Partial<E>[]): Promise<void>;
  (event: Partial<E>, ...events: Partial<E>[]): void;
}
```

```ts
// assume we want all integrations to behave in a sync way
// await new EventBus.putEvent()
new Function(..., async () => bus())

// assume we want all integrations to behave in a sync way
// new EventBus.putEvent()
new Function(..., async () => bus.promise())
```

thantos:
Closes #105 

Updates Resource Callbacks to support promises. Updates most integrations to return callbacks. 

Why? - Lambda supports the full promise API running in a native function, but was written to mock synchronous calls as most of the integration were synchronous. Step Functions and App Sync didn't support promises and assume all integrations are synchronous. This change allows for the addition of async behavior and allows a user in a lambda function to choose how to use the promise API.

Before

```ts
const bus = new EventBus();

const sfn = new StepFunction(this, 'sfn', (event) => {
    bus.putEvents(event);
})

const sfn = new Function(this, 'sfn', async (event) => {
    bus.putEvents(event);
})
```

After

```ts
const bus = new EventBus();

const sfn = new StepFunction(this, 'sfn', async (event) => {
    await bus.putEvents(event);
})

const sfn = new Function(this, 'sfn', async (event) => {
    await bus.putEvents(event);
    const puts = [1,2,3].map(i => bus.putEvents(event));
    await Promise.All(puts);
})
```

New Validation being added: https://github.com/functionless/functionless/issues/105#issuecomment-1147678100

* [x] Refactor interfaces
* [x] Refactor tests
* [x] Refactor test-app
* [ ] Compiler updates
   * [ ] Errors
* [ ] Step Functions
* [ ] App Sync
* [ ] Lambda
* [ ] Language Server
* [ ] Refactor Docs
* [ ] Refactor inline examples

BREAKING_CHANGE: Most integrations return `Promise`s and will need to be awaited in order to work.
BREAKING_CHANGE: Using `.map` in sfn and appsync (?) which calls integrations will need to use `Promise.all()` to join the concurrent requests.
BREAKING_CHANGE: Using `.map` + `Promise.all` in sfn will use max concurrency while the a map that does not return a promise will continue to use 1 concurrency. Note: Use `$SFN.map` to configure the concurency and other aspects of the `Map` State.

### DIFF
--- a/src/function.ts
+++ b/src/function.ts
@@ -103,9 +103,12 @@ abstract class FunctionBase<P, O> implements IFunction<P, O> {
       /**
        * This method is called from the calling runtime lambda code (context) to invoke this lambda function.
        */
+      // @ts-ignore - Typescript fails when comparing Promise<O> with ReturnType<ConditionalFunction<P, O>> though they should be the same.
       call: async (args, prewarmContext) => {
         const [payload] = args;
-        const lambdaClient = prewarmContext.getOrInit(PrewarmClients.LAMBDA);
+        const lambdaClient = prewarmContext.getOrInit<AWS.Lambda>(
+          PrewarmClients.LAMBDA
+        );
         const response = (
           await lambdaClient
             .invoke({
@@ -114,7 +117,7 @@ abstract class FunctionBase<P, O> implements IFunction<P, O> {
             })
             .promise()
         ).Payload?.toString();
-        return response ? JSON.parse(response) : undefined;
+        return (response ? JSON.parse(response) : undefined) as O;
       },
     };
 
@@ -353,8 +356,8 @@ export class ImportedFunction<P, O> extends FunctionBase<P, O> {
 }
 
 type ConditionalFunction<P, O> = P extends undefined
-  ? (payload?: P) => O
-  : (payload: P) => O;
+  ? (payload?: P) => Promise<O>
+  : (payload: P) => Promise<O>;
 
 interface CallbackLambdaCodeProps extends PrewarmProps {}
 
@@ -578,6 +581,10 @@ export class CallbackLambdaCode extends aws_lambda.Code {
   }
 }
 
+type _ReturnType<F extends AnyFunction> = F extends (p: any) => infer O
+  ? O
+  : never;
+
 /**
  * Interface to consume to add an Integration to Native Lambda Functions.
  *
@@ -608,7 +615,7 @@ export interface NativeIntegration<F extends AnyFunction> {
   call: (
     args: Parameters<F>,
     preWarmContext: NativePreWarmContext
-  ) => Promise<ReturnType<F>>;
+  ) => _ReturnType<F>;
   /**
    * Method called outside of the handler to initialize things like the PreWarmContext
    */
--- a/src/step-function.ts
+++ b/src/step-function.ts
@@ -50,6 +50,11 @@ export type AnyStepFunction =
   | ExpressStepFunction<any, any>
   | StepFunction<any, any>;
 
+export type StepFunctionClosure<
+  P extends Record<string, any> | undefined,
+  O
+> = (arg: P) => Promise<O> | O;
+
 export namespace $SFN {
   export const kind = "SFN";
   /**
@@ -134,8 +139,8 @@ export namespace $SFN {
      */
     <T>(
       array: T[],
-      callbackfn: (item: T, index: number, array: T[]) => void
-    ): void;
+      callbackfn: (item: T, index: number, array: T[]) => Promise<void>
+    ): Promise<void>;
     /**
      * Process each item in an {@link array} in parallel and run with the default maxConcurrency.
      *
@@ -155,8 +160,8 @@ export namespace $SFN {
       props: {
         maxConcurrency: number;
       },
-      callbackfn: (item: T, index: number, array: T[]) => void
-    ): void;
+      callbackfn: (item: T, index: number, array: T[]) => Promise<void> | void
+    ): Promise<void>;
   }
 
   export const forEach = makeStepFunctionIntegration<"forEach", ForEach>(
@@ -185,8 +190,8 @@ export namespace $SFN {
      */
     <T, U>(
       array: T[],
-      callbackfn: (item: T, index: number, array: T[]) => U
-    ): U[];
+      callbackfn: (item: T, index: number, array: T[]) => U | Promise<U>
+    ): Promise<U[]>;
     /**
      * Map over each item in an {@link array} in parallel and run with the default maxConcurrency.
      *
@@ -207,8 +212,8 @@ export namespace $SFN {
       props: {
         maxConcurrency: number;
       },
-      callbackfn: (item: T, index: number, array: T[]) => U
-    ): U[];
+      callbackfn: (item: T, index: number, array: T[]) => U | Promise<U>
+    ): Promise<U[]>;
   }
 
   export const map = makeStepFunctionIntegration<"map", Map>("map", {
@@ -389,7 +394,7 @@ abstract class BaseStepFunction<
     aws_stepfunctions.IStateMachine,
     Integration<
       "StepFunction",
-      (input: CallIn) => CallOut,
+      (input: CallIn) => Promise<CallOut>,
       EventBusTargetIntegration<P, StepFunctionEventBusTargetProps | undefined>
     >
 {
@@ -402,7 +407,7 @@ abstract class BaseStepFunction<
   readonly appSyncVtl: AppSyncVtlIntegration;
 
   // @ts-ignore
-  readonly __functionBrand: (arg: CallIn) => CallOut;
+  readonly __functionBrand: (input: CallIn) => Promise<CallOut>;
 
   readonly stateMachineName: string;
   readonly stateMachineArn: string;
@@ -418,17 +423,17 @@ abstract class BaseStepFunction<
     scope: Construct,
     id: string,
     props: StepFunctionProps,
-    func: (arg: P) => O
+    func: StepFunctionClosure<P, O>
   );
 
-  constructor(scope: Construct, id: string, func: (arg: P) => O);
+  constructor(scope: Construct, id: string, func: StepFunctionClosure<P, O>);
 
   constructor(
     scope: Construct,
     id: string,
     ...args:
-      | [props: StepFunctionProps, func: (arg: P) => O]
-      | [func: (arg: P) => O]
+      | [props: StepFunctionProps, func: StepFunctionClosure<P, O>]
+      | [func: StepFunctionClosure<P, O>]
   ) {
     const props =
       isFunctionDecl(args[0]) || typeof args[0] === "function"
@@ -1187,24 +1192,24 @@ export class ExpressStepFunction<
   }
 
   readonly native: NativeIntegration<
-    (input: StepFunctionRequest<P>) => SyncExecutionResult<O>
+    (input: StepFunctionRequest<P>) => Promise<SyncExecutionResult<O>>
   >;
 
   constructor(
     scope: Construct,
     id: string,
     props: StepFunctionProps,
-    func: (arg: P) => O
+    func: StepFunctionClosure<P, O>
   );
 
-  constructor(scope: Construct, id: string, func: (arg: P) => O);
+  constructor(scope: Construct, id: string, func: StepFunctionClosure<P, O>);
 
   constructor(
     scope: Construct,
     id: string,
     ...args:
-      | [props: StepFunctionProps, func: (arg: P) => O]
-      | [func: (arg: P) => O]
+      | [props: StepFunctionProps, func: StepFunctionClosure<P, O>]
+      | [func: StepFunctionClosure<P, O>]
   ) {
     super(
       scope,
@@ -1309,7 +1314,7 @@ export interface ExpressStepFunction<
   P extends Record<string, any> | undefined,
   O
 > {
-  (input: StepFunctionRequest<P>): SyncExecutionResult<O>;
+  (input: StepFunctionRequest<P>): Promise<SyncExecutionResult<O>>;
 }
 
 export class StepFunction<
@@ -1327,24 +1332,26 @@ export class StepFunction<
   public static readonly FunctionlessType = "StepFunction";
 
   readonly native: NativeIntegration<
-    (input: StepFunctionRequest<P>) => AWS.StepFunctions.StartExecutionOutput
+    (
+      input: StepFunctionRequest<P>
+    ) => Promise<AWS.StepFunctions.StartExecutionOutput>
   >;
 
   constructor(
     scope: Construct,
     id: string,
     props: StepFunctionProps,
-    func: (arg: P) => O
+    func: StepFunctionClosure<P, O>
   );
 
-  constructor(scope: Construct, id: string, func: (arg: P) => O);
+  constructor(scope: Construct, id: string, func: StepFunctionClosure<P, O>);
 
   constructor(
     scope: Construct,
     id: string,
     ...args:
-      | [props: StepFunctionProps, func: (arg: P) => O]
-      | [func: (arg: P) => O]
+      | [props: StepFunctionProps, func: StepFunctionClosure<P, O>]
+      | [func: StepFunctionClosure<P, O>]
   ) {
     super(
       scope,
@@ -1386,7 +1393,7 @@ export class StepFunction<
 
   public describeExecution = makeIntegration<
     "StepFunction.describeExecution",
-    (executionArn: string) => AWS.StepFunctions.DescribeExecutionOutput
+    (executionArn: string) => Promise<AWS.StepFunctions.DescribeExecutionOutput>
   >({
     kind: "StepFunction.describeExecution",
     appSyncVtl: this.appSyncIntegration({
@@ -1438,13 +1445,11 @@ export class StepFunction<
 
         const [arn] = args;
 
-        const result = await stepFunctionClient
+        return stepFunctionClient
           .describeExecution({
             executionArn: arn,
           })
           .promise();
-
-        return result;
       },
     },
     unhandledContext: (kind, contextKind) => {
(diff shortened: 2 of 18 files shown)


src/integration.ts
```diff
@@ -2,29 +2,48 @@ import { ApiGatewayVtlIntegration } from "./api";
...
   return isCallExpr(node) && isReferenceExpr(node.expr);
 }
 
+export function isIntegrationCallPattern(
+  node: FunctionlessNode
+): node is
+  | IntegrationCallExpr
+  | (AwaitExpr & { expr: IntegrationCallExpr })
+  | (PromiseExpr & { expr: IntegrationCallExpr })
+  | (AwaitExpr & { expr: PromiseExpr & { expr: IntegrationCallExpr } }) {
```
sam-goodwin, line 36: I'm curious - can this type be inferred from the impl?

src/step-function.ts
```diff
@@ -47,6 +47,16 @@ export type AnyStepFunction =
   | ExpressStepFunction<any, any>
   | StepFunction<any, any>;
 
+export type StepFunctionClosure<
+  Payload extends Record<string, any> | undefined,
+  Out
+> = (arg: Payload) => Promise<Out> | Out;
+
+type ParallelFunction<T> = () => Promise<T> | T;
```
sam-goodwin, line 55: Why not always a Promise

src/util.ts
```diff
@@ -141,6 +136,23 @@ export const isPrimitive = (val: any): val is PrimitiveValue => {
   );
 };
 
+export function isPromiseAll(expr: CallExpr): expr is CallExpr & {
```
sam-goodwin, line 139: Docs

src/util.ts
```diff
@@ -141,6 +136,23 @@ export const isPrimitive = (val: any): val is PrimitiveValue => {
...
+  expr: PropAccessExpr & {
+    name: "all";
+    parent: {
+      kind: "Identifier";
+      name: "Promise";
+    };
+  };
```
sam-goodwin, line 146: I do like being explicit like this. Can make understanding easier.

src/validate.ts
```diff
@@ -97,23 +113,97 @@ export function validate(
         ),
       ];
     }
+    if (ts.isCallExpression(node)) {
+      return validatePromiseCalls(node);
+    } else return [];
+  }
+
+  function validatePromiseCalls(node: ts.CallExpression): ts.Diagnostic[] {
```
sam-goodwin, line 121: Perhaps we should start documenting the syntax it catches

src/visit.ts
```diff
@@ -517,5 +535,53 @@ export function visitBlock(
   });
 }
 
+/**
+ * Starting at the root, explore the children without processing until one or more start nodes are found.
+ *
+ * For each Start nodes, apply {@link visitEachChild} to it with the given callback.
+ */
+export function visitSpecificChildren<T extends FunctionlessNode>(
```
sam-goodwin, line 543: What's a use case for this?

src/vtl.ts
```diff
@@ -261,9 +264,17 @@ export abstract class VTL {
...
-        return this.integrate(serviceCall, node);
+      if (isReferenceExpr(node.expr)) {
+        const ref = node.expr.ref();
+        if (isIntegration<Integration>(ref)) {
+          const serviceCall = new IntegrationImpl(ref);
+          return this.integrate(serviceCall, node);
+        } else {
+          throw new SynthError(
+            ErrorCodes.Unexpected_Error,
+            "Called references are expected to be an integration."
```
sam-goodwin, line 275: Is this error code too general?

src/vtl.ts
```diff
@@ -526,6 +537,32 @@ export abstract class VTL {
...
+        isCallExpr(node.parent.parent) &&
+        isPromiseAll(node.parent.parent)
+      ) {
+        return this.eval(node.expr);
+      }
+      debugger;
+      // TODO create error code
+      throw new SynthError(
+        ErrorCodes.Arrays_of_Integration_must_be_immediately_wrapped_in_Promise_all
+      );
```
sam-goodwin, line 565: What does it mean to do parallel promises in appsync?

sam-goodwin (changes requested):

src/asl.ts
```diff
@@ -403,13 +414,37 @@ export class ASL {
...
+                return node.expr.expr;
+              }
+            } else if (
+              isPromiseExpr(node) &&
+              isIntegrationCallExpr(node.expr)
+            ) {
+              return node.expr;
+            }
+            return node;
+          })();
```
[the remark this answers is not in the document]
thantos: I could, it was originally returning an array (start to end), but realized start was always `node`. Will change.

src/asl.ts
```diff
@@ -926,6 +1002,17 @@ export class ASL {
...
+      } else if (isPromiseAll(expr)) {
+        const values = expr.getArgument("values");
+        // just validate Promise.all and continue, will validate the PromiseArray later.
+        if (values?.expr && isPromiseArrayExpr(values?.expr)) {
+          return this.eval(values.expr, props);
+        }
+        debugger;
+        // TODO create error code
+        throw Error(
+          "`Promise.all` must wrap an expression that returns an array of promises like `Array.map`"
```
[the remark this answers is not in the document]
thantos: yeah, its validated, not sure how this was missed.

src/asl.ts
```diff
@@ -985,14 +1072,16 @@ export class ASL {
...
         };
-      } else if (isCallExpr(expr.right)) {
+      } else if (isAwaitExpr(expr.right) || isCallExpr(expr.right)) {
         return this.eval(expr.right, {
           ...props,
           ResultPath: ASL.toJsonPath(expr.left),
         });
       }
     } else if (isBinaryExpr(expr)) {
       // TODO
```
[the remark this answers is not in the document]
thantos: it just falls through and fails on any unsupported binaryExpr. we do separately support binary `=`

src/checker.ts
```diff
@@ -354,6 +363,51 @@ export function makeFunctionlessChecker(
...
 
+  function typeMatch(
+    type: ts.Type,
+    predicate: (type: ts.Type) => boolean
+  ): boolean {
+    if (type.isUnionOrIntersection()) {
+      return predicate(type) || type.types.some((t) => typeMatch(t, predicate));
+    }
+    return predicate(type);
+  }
```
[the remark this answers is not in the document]
thantos: I was not able to find any? Always surprised by what isn't there or easy to find.

src/checker.ts
```diff
@@ -354,6 +363,51 @@ export function makeFunctionlessChecker(
...
+    return checker.getFullyQualifiedName(symbol) === "Array";
+  }
+
+  function isPromiseArray(type: ts.Type): boolean {
+    const symbol = type.getSymbol();
+    if (!symbol) {
+      return false;
+    }
+    const typeParams = checker.getTypeArguments(type as ts.TypeReference);
+    if (isArraySymbol(symbol) && typeParams?.length === 1) {
```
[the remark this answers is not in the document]
thantos: In this case it returns false. and yeah, this is what found to do this, seemed to work.

src/checker.ts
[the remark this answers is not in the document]
sam-goodwin: I've noticed that not all functions on the checker are in the TypeChrcker type.

src/compile.ts
```diff
@@ -688,7 +685,15 @@ export function compile(
...
 
-          return getCall();
+          const call = getCall();
+
+          const type = checker.getTypeAtLocation(node);
+          const typeSymbol = type.getSymbol();
+          return typeSymbol && checker.isPromiseSymbol(typeSymbol)
+            ? newExpr("PromiseExpr", [call])
+            : checker.isPromiseArray(type)
+            ? newExpr("PromiseArrayExpr", [call])
```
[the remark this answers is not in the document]
thantos: yes, but we'd be left with just a call and not the array if we didn't wrap the call. This is mostly a cheap way to inject type information right now, we can remove it later.

src/error-code.ts
```diff
@@ -343,13 +343,13 @@ export namespace ErrorCodes {
    *
    * For example:
    * ```ts
-   * new AwsMethod({
-   *   request: ($input) => $AWS.DynamoDB.GetItem({
-   *     Table: table,
+   * new AwsMethod(
+   *   ($input) => $AWS.DynamoDB.GetItem({
+   *     TableName: table,
```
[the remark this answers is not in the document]
thantos: I know, most of them were updated, missed one.

src/expression.ts
```diff
@@ -446,5 +449,41 @@ export class TypeOfExpr extends BaseExpr<"TypeOfExpr"> {
...
+  constructor(readonly expr: Expr) {
+    super("PromiseArrayExpr");
+
+    expr.setParent(this);
+  }
+
+  public clone(): this {
+    return new PromiseArrayExpr(this.expr.clone()) as this;
+  }
+}
```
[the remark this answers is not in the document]
thantos: type information. agree they are not ideal. Wanted a way to cheaply and consistently find bad promise patterns. It also allows us to support integrations and calls that are not promises. I started with isPromise flags on various types, but that was cumberum and didn't support the Array<Promise> case.

src/function.ts
```diff
@@ -316,9 +316,12 @@ abstract class FunctionBase<in Payload, Out>
       /**
        * This method is called from the calling runtime lambda code (context) to invoke this lambda function.
        */
+      // @ts-ignore - Typescript fails when comparing Promise<O> with ReturnType<ConditionalFunction<P, O>> though they should be the same.
```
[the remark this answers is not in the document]
thantos: image

Error when removing the ts-ignore

image

I played with this for a long time, if I remove the conditional of the conditional function, it works fine.

src/integration.ts
thantos: No, because the user can choose the pattern. Await may be on a non-promise, a promise may be on a return, the integration may not return a promise.

src/step-function.ts
thantos: It allows for functions that don't return promise.

```ts
() => 1 + 2 // this fails without | T because it doesn't return a promise.
() => func() // promise
async () => {
    const await func();
}
```

src/function.ts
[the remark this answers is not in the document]
sam-goodwin: Looks to be a contra variant issue.

src/integration.ts
sam-goodwin: Oh I meant the return type signature

src/visit.ts
thantos:
```ts
new AwaitExpr(new PromiseExpr(new CallExpr()))
```

I want to explore the structure of Await and Promise, but don't start call the callback until we hit `CallExpr`.

Basically a range of nodes to take as is, without knowing their structure.

src/vtl.ts
thantos: I don't think a user should be able to create this right now. We can change it if that turns out to be wrong.

src/validate.ts
```diff
@@ -97,23 +113,97 @@ export function validate(
         ),
       ];
     }
+    if (ts.isCallExpression(node)) {
```
sam-goodwin, line 116: Do we validate that they don't use promise.then ?

src/vtl.ts
thantos: Think we should just fail if a `map` returns a promise or uses one?

src/validate.ts
thantos: It would just say that Promise.then is an unsupported call. I can add more specific errors.

src/function.ts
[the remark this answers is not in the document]
thantos: Its mostly an internal error so I left it. I did play with various combinations, but there was nothing to put contra variance on.

src/integration.ts
thantos: Actually updated this to a type now so we can use it for other things.

src/expression.ts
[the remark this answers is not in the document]
sam-goodwin: Cool, looking forward to integrating deepkit's type ast transformer https://github.com/functionless/functionless/issues/119#issuecomment-1156794411

src/vtl.ts
sam-goodwin: PromiseAll is evaluated above

src/validate.ts
sam-goodwin: In `validate.ts` or at synth time?

sam-goodwin (approved):

src/vtl.ts
thantos: Yeah, messed up here, there are two cases. 

1. Must wrap PromiseArrays with Promise.all
2. Must only use Promise.all to wrap PromiseArrays

Fixing

src/checker.ts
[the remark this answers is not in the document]
thantos: I do search in the code base when looking for behavior to see if there is something I can at least copy. Didn't find anything, I am sure it is in there and obscure.

src/validate.ts
thantos:
```ts
    + https://functionless.org/docs/error-codes#integration-must-be-immediately-awaited-or-returned
    +
    + 96   return await func().then((x) => x);
    +                  ~~~~~~~
    + test/test-files/step-function.ts:100:15 - error Functionless(10016): Integration must be immediately awaited or returned
```

I think this is OK for now.

thantos:
Build failing due to latest `localstack` cli regression: https://github.com/localstack/localstack/issues/6334

thantos:
BREAKING CHANGE: Most integrations return `Promise`s and will need to be awaited in order to work.
BREAKING CHANGE: Using `.map` in sfn and appsync (?) which calls integrations will need to use `Promise.all()` to join the concurrent requests.
BREAKING CHANGE: Using `.map` + `Promise.all` in sfn will use max concurrency while the a map that does not return a promise will continue to use 1 concurrency. Note: Use `$SFN.map` to configure the concurency and other aspects of the `Map` State.

PR #211, opened 2022-06-09
chore: move app sync tests to snapshots

thantos:
Part of #202 

* Reduces the size and fragility of the app sync tests by outputting a snapshot instead of hand written evaluation.
* Wraps the template execution into a app sync single test method to support simpler tests
* Support matchObject to test granular cases like CDK tokens without the previous verbosity

### DIFF
--- a/test/appsync.test.ts
+++ b/test/appsync.test.ts
@@ -15,10 +15,10 @@ describe("step function integration", () => {
       reflect(() => {
         machine({});
       }),
-      [
-        {
-          index: 1,
-          expected: {
+      {
+        executeTemplates: [
+          {
+            index: 1,
             match: {
               params: {
                 body: {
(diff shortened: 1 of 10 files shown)


test/appsync.test.ts
```diff
@@ -73,32 +39,24 @@ $util.qr($v1.put('stateMachineArn', '${machine.stateMachineArn}'))
...
+      [
+        {
+          index: 1,
+          expected: {
+            match: {
+              params: {
+                body: {
+                  stateMachineArn: machine.stateMachineArn,
+                },
+              },
```
sam-goodwin, line 55: Is this necessary? Seems verbose and redundant?

thantos: It lets us test that the CDK tokens are in the right place, because we strip them out for the snapshot.

thantos: I could remove the expected wrapper.

sam-goodwin (approved):

thantos:
Accidentally merged with a stray commit on my local main, removed the commit, rebased, force pushed.

Issue #128, opened 2022-05-07
Native Function Serialization can fail when async promises do not complete.


thantos:
CDK does [not officially support async operations](https://twitter.com/samgoodwin89/status/1516887131108438016?s=20&t=7GRGOQ1Bp0h_cPsJgFk3Ww) however, Functionless's native Functions (#103) rely on pulumi's async function serialization. 

We first mitigate this issue by:
1. starting an async promise and leaving it dangling. 
2. Registering the dangling promise with a global array of promises behind `Function.promises`
3. adding `asyncSynth` which awaits on `Function.promises` with `Promise.all`

This works when 1) asyncSynth is used 2) `Function.promises` is manually awaited on and 3) when the `cdk` cli and auto-synth is used.

The reason for 3 is that CDK's auto-synth is called as part of the node process exiting. When node exits, promises are allowed to complete (join). 

When one of the above strategies is not used, it is possible that the native functions will not be serialized and output, causing the lambda code to be invalid. 

We output the following text into the lambda: `"If you are seeing this in your lambda code, ensure generate is called and then consult the README."`

Obviously the failure case isn't user friendly.

This ticket is a placeholder to do one or more of the following:
1. Once CDK native has async support, use it (https://github.com/sam-goodwin/aws-cdk/pull/1)
2. replace the "if you are seeing" message with a poison pill that fails synth or deployment. 
3. Use a serialization function that is not async

volkanunsal:
@thantos 

Can you give an example for this:

```
If you are seeing this in your lambda code, ensure generate is called
```

thantos:
If you are using the CDK cli and auto-synth, there is nothing to do.

If you are programmatically calling synth for something like tests, you can either use `asyncSynth` (exported from Functionless) from an async scope, or await on `Promise.all(Function.promises)`

volkanunsal:
> If you are using the CDK cli and auto-synth, there is nothing to do.

Does this mean this can't be done... or that it's done automatically? If it's done automatically, I may be missing something because it's not happening for me.

sam-goodwin:
It means done automatically. The node process won't exit until all promises have completed (ie when the event loop is empty). 

What error are you seeing?

volkanunsal:
@sam-goodwin 

I'm not seeing any error during deployment, but the lambda code only contains the string:

```
If you are seeing this in your lambda code, ensure generate is called and then consult the README.
```

sam-goodwin:
[...]

PR #229, opened 2022-06-12
fix: function invalid/incomplete closure serialize poison pill

### LINKED ISSUE #223: Error Code page Gen missing from build path
After PR #221, the error code page changed, but was not included in the self-mutation.

Likely due to prettier.

* [ ] Error code page is always up to date in PR (either self-mutation or commit hook).

### LINKED ISSUE #227: Multiple calls to app.synth() breaks Function serialization
If the CDK `app.synth()` is called twice, such as having `autoSynth:true` and then manually calling it, the Function closure serializer fails to re-bind the function body, resulting in the dummy code being deployed.

See https://github.com/functionless/functionless/issues/128#issuecomment-1153030191

thantos:
Resolves #227, Closes #223 

* If the CDK app uses `Function`s and does not await the Function Closure Serialize functions (see #128), the synthesis will fail with validation errors.
* Removed the error code page from the repo, it will be auto generated when the website is generated.
  * Originally we wanted to add this generation to the pre-commit hook. However, the generation of this file requires the typedocs to be generated and that requires the while site to build. This is a slow process we should not wait on during pre-commit. 
  * Now the error code page will be generated just like the updated typedocs, during the netlfiy build. The Netlify output can be used to review the changes.


src/function.ts
```diff
@@ -608,6 +611,19 @@ export class Function<in Payload, Out, OutPayload extends Payload = Payload> ext
...
+        // In order for the synthesis to complete successfully 
+        // 1. Use autoSynth `new App({ autoSynth: true })` or `new App()` with the CDK Cli (`cdk synth`)
+        // 2. Use `await asyncSynth(app)` exported from Functionless in place of `app.synth()`
+        // 3. Manually await on the closure serializer promises `await Promise.all(Function.promises)`
+        // https://github.com/functionless/functionless/issues/128
+        _resource.node.addValidation({
+          validate: () => this.resource.node.metadata.find(m => m.type === FUNCTION_CLOSURE_FLAG) 
+            ? [] 
+            : [formatErrorMessage(ErrorCodes.Function_Closure_Serialization_Incomplete)]
+        });
```
sam-goodwin, line 625: When do these validations run?

thantos: Immediately before synth, after app.synth is called.

I'm pretty excited about the find. Feel like there will be more uses for this...

The cool thing is that the validations do not short circuit, cdk shows all errors for all constructs. 

Which means we could do something like error on an assertion about least privilege at the last moment. 

() => table.neverWrite() // fails during synth if the lambda role contains write

sam-goodwin: I was wondering how it plays with the promises? Wouldn't the validate have a chance of running before the promises have completed?

thantos: It should run as part of synth, so if the promises have been awaited before synth, the validation will pass and if the promises have not been allowed to complete, they will fail.

There is a chance that the validation succeeds because the promises happen to complete before synth is called, in which case the state is correct.

1. serialize 
2. synth
   3. validate (fail if 1 has not completed - only possible by luck)

1. serialize
2. await
3. synth
  4. validate (pass)
  5. output

sam-goodwin: What happens when using auto synth? The promises won't be awaited

thantos: The node process waits for the promises before executing synth. I tested this with the cli. 

`cdk synth` works as long as `app.synth()` isn't called manually.

sam-goodwin (approved):

Issue #61, opened 2022-03-30
Remaining $util helpers for AppSync


sam-goodwin:
We're missing various helpers from https://docs.aws.amazon.com/appsync/latest/devguide/resolver-util-reference.html

Need to provide complete list.

Anyway to code generate?

thantos:
Related: #247

PR #231, opened 2022-06-12
feat: VTL util.log types

volkanunsal:
Marks progress towards #61

### DIFF
--- a/src/appsync.ts
+++ b/src/appsync.ts
@@ -514,6 +514,13 @@ export interface $util {
   readonly time: time;
 
   /**
+   * The $util.log variable contains log methods to help log info and error messages.
+   *
+   * @see https://docs.aws.amazon.com/appsync/latest/devguide/utility-helpers-in-util.html
+   */
+  readonly log: log;
+
+  /**
    * Returns the input string as a JavaScript escaped string.
    */
   escapeJavaScript(js: string): string;
@@ -706,6 +713,25 @@ export interface $util {
    * Returns a string describing the multi-auth type being used by a request, returning back either "IAM Authorization", "User Pool Authorization", "Open ID Connect Authorization", or "API Key Authorization".
    */
   authType(): string;
+}
+
+export interface log {
+  /**
+   * Logs the string representation of the provided object to the requested log stream when request-level and field-level CloudWatch logging is enabled with log level ALL on an API.
+   */
+  info(obj: any): void;
+  /**
+   * Logs the string representation of the provided objects to the requested log stream when request-level and field-level CloudWatch logging is enabled with log level ALL on an API. This utility will replace all variables indicated by "{}" in the first input format string with the string representation of the provided objects in order.
+   */
+  info(message: string, ...rest: any[]): void;
+  /**
+   * Logs the string representation of the provided object to the requested log stream when field-level CloudWatch logging is enabled with log level ERROR or log level ALL on an API.
+   */
+  error(obj: any): void;
+  /**
+   * Logs the string representation of the provided objects to the requested log stream when field-level CloudWatch logging is enabled with log level ERROR or log level ALL on an API. This utility will replace all variables indicated by "{}" in the first input format string with the string representation of the provided objects in order.
+   */
+  error(message: string, ...rest: any[]): void;
 }
 
 export interface time {
--- a/test/__snapshots__/vtl.test.ts.snap
+++ b/test/__snapshots__/vtl.test.ts.snap
@@ -1,5 +1,53 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`$util.log.error(message) 1`] = `
+Array [
+  "{
+  \\"version\\": \\"2018-05-29\\",
+  \\"payload\\": null
+}",
+  "#return($util.log.error('hello world'))",
+]
+`;
+
+exports[`$util.log.error(message, ...Object) 1`] = `
+Array [
+  "{
+  \\"version\\": \\"2018-05-29\\",
+  \\"payload\\": null
+}",
+  "#set($v1 = {})
+$util.qr($v1.put('a', 1))
+#set($v2 = {})
+$util.qr($v2.put('b', 2))
+#return($util.log.error('hello world', [$v1, $v2]))",
+]
+`;
+
+exports[`$util.log.info(message) 1`] = `
+Array [
+  "{
+  \\"version\\": \\"2018-05-29\\",
+  \\"payload\\": null
+}",
+  "#return($util.log.info('hello world'))",
+]
+`;
+
+exports[`$util.log.info(message, ...Object) 1`] = `
+Array [
+  "{
+  \\"version\\": \\"2018-05-29\\",
+  \\"payload\\": null
+}",
+  "#set($v1 = {})
+$util.qr($v1.put('a', 1))
+#set($v2 = {})
+$util.qr($v2.put('b', 2))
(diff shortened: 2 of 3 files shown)


src/appsync.ts
```diff
@@ -370,12 +370,12 @@ export class AppsyncResolver<
                 // lambdaFunction().prop
                 // table.query().Items.size() //.size() is still a Call but not a call to a service.
                 return expr;
-              } else if (expr.kind === "PropAccessExpr") {
+              } else if (expr.kind === 'PropAccessExpr') {
```
thantos, line 373: Please ensure you are using the prettier profile with the project.

src/appsync.ts
```diff
@@ -708,6 +708,25 @@ export interface $util {
...
+  /**
+   * Logs the string representation of the provided object to the requested log stream when field-level CloudWatch logging is enabled with log level ERROR or log level ALL on an API.
+   */
+  error(obj: Record<string, any>): void;
+  /**
+   * Logs the string representation of the provided objects to the requested log stream when field-level CloudWatch logging is enabled with log level ERROR or log level ALL on an API. This utility will replace all variables indicated by "{}" in the first input format string with the string representation of the provided objects in order.
+   */
+  error(message: string, rest: Record<string, any>[]): void;
+}
+
```
thantos, line 736: Awesome! the recent utils additions to app sync are exciting to see.

src/appsync.ts
```diff
@@ -708,6 +715,25 @@ export interface $util {
...
+   */
+  info(message: string, rest: Record<string, any>[]): void;
+  /**
+   * Logs the string representation of the provided object to the requested log stream when field-level CloudWatch logging is enabled with log level ERROR or log level ALL on an API.
+   */
+  error(obj: Record<string, any>): void;
+  /**
+   * Logs the string representation of the provided objects to the requested log stream when field-level CloudWatch logging is enabled with log level ERROR or log level ALL on an API. This utility will replace all variables indicated by "{}" in the first input format string with the string representation of the provided objects in order.
+   */
+  error(message: string, rest: Record<string, any>[]): void;
```
thantos, line 734: `...rest: Record<string, any>[]`

volkanunsal:
Sorry about all the formatting changes. My prettier seems to be using a different config than yours. Maybe it'd be a good idea to include a prettier config for the project.

thantos (changes requested):
Please add tests. I just merged a change to use snapshot tests and an updated appSyncTest method for app sync/vtl that'll make testing app sync easier.

thantos:
> Sorry about all the formatting changes. My prettier seems to be using a different config than yours. Maybe it'd be a good idea to include a prettier config for the project.

Hmm.

Could have sworn we had one, will look into it. (not sure how it was working up until now).

volkanunsal:
There is a git commit hook that runs prettier, which uses an internal config probably.

volkanunsal:
Should be good now.

test/vtl.test.ts
```diff
@@ -359,6 +359,34 @@ test("$util.time.nowISO8601", () =>
     })
   ));
 
+test("$util.log.info(message)", () =>
+  appsyncTestCase(
+    reflect(() => {
+      return $util.log.info("hello world");
+    })
+  ));
```
sam-goodwin, line 367: Does this create snapshot tests? I don't see them in the PR

src/appsync.ts
```diff
@@ -708,6 +715,25 @@ export interface $util {
...
+   */
+  info(obj: Record<string, any>): void;
+  /**
+   * Logs the string representation of the provided objects to the requested log stream when request-level and field-level CloudWatch logging is enabled with log level ALL on an API. This utility will replace all variables indicated by "{}" in the first input format string with the string representation of the provided objects in order.
+   */
+  info(message: string, ...rest: Record<string, any>[]): void;
+  /**
+   * Logs the string representation of the provided object to the requested log stream when field-level CloudWatch logging is enabled with log level ERROR or log level ALL on an API.
+   */
+  error(obj: Record<string, any>): void;
```
sam-goodwin, line 730: Is it required to be an object? Can it be a primitive like string?

sam-goodwin (changes requested):

test/vtl.test.ts
volkanunsal: I can't run the tests locally, for some reason. It fails due to AWS account number not being valid. Is there another command to record the snapshots?

src/appsync.ts
volkanunsal: you're right. it's a string.

volkanunsal: ah, actually, I was wrong... it does say it's an object.

> $util.log.info(Object) : Void

volkanunsal: The second method signature handles the string usecase, though.

test/vtl.test.ts
sam-goodwin: Oh have you set up local stack? Our tests spin that up locally and deploy stacks to it. https://localstack.cloud/

We need to set up a contributing guide. Thanks for the reminder.

sam-goodwin: Tracking https://github.com/functionless/functionless/issues/234

src/appsync.ts
sam-goodwin: AWS's documentation uses `Object` as `any` - it's Java based, so `Object``` just means any value type can be passed in. Not `Record<string, any>`

sam-goodwin:
Thanks for this change @volkanunsal! Super useful feature. I wonder if we should compile console.log to $util.log?

Also, a suggestion for your development lifecycle. Don't use `git push --force`, that incremental changes are visible as individual commits on the PR. When we merge all the commits are squashed into one, so everything is clean.

thantos:
Something else we should keep in mind is that logging in appSync requires logging to be enabled. 

Should we:
1. validate that logging is enabled and fail if logging is used, but not enabled?
2. Turn on logging when logging is used? probably not just a flag.
3. Nothing, appSync doesn't describe the ramification of using `log` when logging is off, so I assume it just does nothing?

src/appsync.ts
thantos: Confirmed

```
$util.log.info(" 1 {} 2 {}", $v1.containsKey('key'), "hi")
```

out

```
c98f529-d1b3-4a03-a5ff-22853b3caa4e INFO - 1 true 2 hi
```

thantos: Note, anything in the first position when more than 1 argument is given appears to fail the log... (I tried passing a boolean).

sam-goodwin:
> Nothing, appSync doesn't describe the ramification of using log when logging is off, so I assume it just does nothing?

Test this. If true, then I think this should be the behavior.

volkanunsal:
FYI. I couldn't run the tests with `project test`, and had to use `project build` instead.

test/__snapshots__/vtl.test.ts.snap
```diff
@@ -1,5 +1,53 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`$util.log.error(message) 1`] = `
+Array [
+  "{
+  \\"version\\": \\"2018-05-29\\",
+  \\"payload\\": null
+}",
+  "#return($util.log.error('hello world'))",
```
sam-goodwin, line 9: Curious what this does at runtime? Return null?

sam-goodwin (approved):

Issue #238, opened 2022-06-12
Tracking ticket for Lambda/Function Closure Serializer robustness


thantos:
* [ ] #237

Issue #236, opened 2022-06-12
table.query is not a function


volkanunsal:
I tried using `Table` inside a `Function`, and got this error. The generated code seems to be missing every method. Did I do this right?

```ts
ddbTable.addGlobalSecondaryIndex({
  indexName: 'GSI1',
  partitionKey: { name: 'PK', type: aws_dynamodb.AttributeType.STRING },
  sortKey: { name: 'SK', type: aws_dynamodb.AttributeType.STRING },
  projectionType: aws_dynamodb.ProjectionType.ALL,
});

const table = new Table<ITable, 'PK', any>(ddbTable);

const defaultAuthFunction = new Function<AuthContext, AuthResponse>(
  this,
  'LambdaAuthorizer',
  {
    logRetention: 14,
  },
  async (c) => {
    const apiKey = c.authorizationToken;

    // Look up the account using the auth token.
    const item = table.query({
      index: 'GSI1',
      query: {
        expression: 'PK = :pk',
        expressionValues: { ':pk': { S: apiKey } },
      },
    }).items[0];

    // If it exists and it's a valid account, allow request.
    if (item?.entityType === 'Account') {
      const accountId = item.Id;
      return {
        isAuthorized: true,
        // Add account id and api key into the resolver context.
        resolverContext: { accountId, apiKey },
        deniedFields: [],
      };
    }

    //
    return {
      isAuthorized: false,
      resolverContext: {},
      deniedFields: [],
    };
  },
);
```

And here is the part of the generated code with the `__table` object.

```js
var __table = {};
var __table_query = {};
__table_query.kind = "Table.query";
__table_query.unhandledContext = __f5;
var __table_query_native = {call: undefined, preWarm: undefined};
__table_query.native = __table_query_native;
__table.query = __table_query;
```

thantos:
Table.* Doesn't work in function right now. 

You can use `$AWS.DynamoDB.Query()`

The reason is table.* Use the app sync contracts right now, we have a ticket to resolve this, but it's non trivial to recreate the next token and other parts of the app sync to dynamo contract.

The larger issue is that lambda doesn't give an error at compile/synth time which I'll look into.

volkanunsal:
Got it. I'll try that. `$AWS.DynamoDB.Query()` isn't a great option because it returns Dynamo types, which trip up Typescript when discriminating on tagged unions,  and I would have to erase the types entirely.

volkanunsal:
I'm getting this error now

```
[...]

Issue #239, opened 2022-06-12
Refactor Lambda/Function closure serializer field removal


thantos:
Field removal is verbose and manual in the `serialize` function. Make this configuration based rather than the order of the transformation and nested functions.

PR #243, opened 2022-06-13
feat: improve support for CDK tokens in serialized closure.

### LINKED ISSUE #237: Bug(Lambda): Serialized Closures which contain constructs with resolvable objects fail to deploy
From https://github.com/functionless/functionless/issues/236#issuecomment-1153278369

```ts
CfnSynthesisError: Resolution error: Supplied properties not correct for "CfnFunctionProps"
  environment: supplied properties not correct for "EnvironmentProperty"
    variables: element 'env__functionless315': [{"indexName":"GSI1","keySchema":[{"attributeName":"PK","keyType":"HASH"},{"attributeName":"SK","keyType":"RANGE"}],"projection":{"projectionType":"ALL"}}] should be a string.
```

IResolvables that resolve to non-strings fail to deploy as Lambda Env variables can only be strings.

thantos:
Closes #237 
Part of #238 
Solution for https://github.com/functionless/functionless/issues/236#issuecomment-1153278369

* [x] Resolve objects when possible
* [x] Handle unresolvable object
* [x] More tests
    * [x] Basic token types
    * [x] Table with GSI https://github.com/functionless/functionless/issues/236#issuecomment-1153278369
    * [x] Function https://github.com/functionless/functionless/issues/236#issuecomment-1153278369
    * [x] Late resolved objects
    * [x] Synth time generated strings (ex: arns)
* [x] Early resolve for strings when possible

BREAKING_CHANGE: `.resource` cannot be used on `EventBus`, `Function`, `Table`, `StepFunction`, or `ExpressStepFunction` inside of a Function Closure.
BREAKING_CHANGE: `.bus` on `EventBus` and `.rule` on `Rule` has been renamed to `.resource`


src/function.ts
```diff
@@ -788,6 +795,24 @@ export class CallbackLambdaCode extends aws_lambda.Code {
                 }
               }
 
+              return o;
+            };
+
+                        /**
```
sam-goodwin, line 801: Weird formatting

src/function.ts
```diff
@@ -760,13 +762,18 @@ export class CallbackLambdaCode extends aws_lambda.Code {
...
-                tokens = [...tokens, token];
-                return token;
+                const reversed = Tokenization.reverse(o)!;
+                if(Reference.isReference(reversed)) {
+                  tokens.push(reversed.toString());
+                  return reversed.toString();
+                } else if("value" in reversed) {
+                  return (reversed as unknown as {value: any}).value;
+                }
+                // TODO: fail at runtime and warn at compiler time when a token cannot be serialized
```
sam-goodwin, line 775: What cases are there?

test/function.localstack.test.ts
```diff
@@ -50,23 +51,25 @@ interface TestFunctionResource {
     payload?: I | ((context: Outputs) => I)
   ): void;
 
-  skip: <I, O, Outputs extends Record<string, string> = Record<string, string>>(
+  skip: <
+    I,
+    O, // Forces typescript to infer O from the Function and not from the expect argument.
```
sam-goodwin, line 56: How/why?

test/function.localstack.test.ts
```diff
@@ -603,7 +548,129 @@ localstackTestSuite("functionStack", (testResource, _stack, _app) => {
...
+
+      const flTable = new Table(table);
+
+      return new Function(
+        parent,
+        "function",
+        localstackClientConfig,
+        async () => {
+          get({
+            TableName: flTable,
```
sam-goodwin, line 614: Oh dang, this forces the entire table to be serialized?

test/function.localstack.test.ts
```diff
@@ -603,7 +548,129 @@ localstackTestSuite("functionStack", (testResource, _stack, _app) => {
...
+        outputs: { table: table.tableArn },
+      };
+    },
+    (outputs) => {
+      return outputs.table;
+    }
+  );
+
+  test(
+    "serialize token with lazy should fail",
```
sam-goodwin, line 656: Why should it fail? Cant we resolve the value?

test/function.localstack.test.ts
```diff
@@ -712,7 +779,54 @@ const testFunction = async (
...
+          timeout: Duration.seconds(20),
+        },
+        async () => {
+          const bus = new aws_events.EventBus(stack, "busbus");
+          return bus.eventBusArn;
+        }
+      );
+      await Promise.all(Function.promises);
+    }
+    // TODO: add error message
```
sam-goodwin, line 809: Should we catch this in the language service too?

src/function.ts
thantos: hmmm, can't wait for prettier to support 4.7!

src/function.ts
thantos: good question, part of the problem is that I don't know. `LazyAny` is one.

test/function.localstack.test.ts
thantos: yes, its pretty easy to do that and with the refactor to not re-write the closures, we'll almost always serialize the entire constructs.

test/function.localstack.test.ts
thantos: We can, though I am concerned that we'd be resolving the lazy too early.

Lazy relies on being resolved at the last moment to catch anything that may change with its contents (ex: adding more GSIs). However we current start the lambda serialization immediately and may end up with a race condition.

Need to test more.

test/function.localstack.test.ts
sam-goodwin: Do we get a chance to rewrite it in the serializer?

test/function.localstack.test.ts
thantos: What would we do? Warn when app.synth() is used?

sam-goodwin: Was just thinking of an error code in the IDE

sam-goodwin: This is for creating constructs in the function right?

test/function.localstack.test.ts
thantos: Yeah, we remove common Construct/Resource and CfnResource fields. Heuristic driving right now. #239

thantos: Which of course means that `table.resource.globalSecondaryIndices.length` would fail in a `Function` at runtime.

test/function.localstack.test.ts
thantos: Ohh, yeah. Yeah, we can detect new on Construct or Resource (we should create a FunctionlessResource) interface or something.

thantos: This test was just moved in this PR. https://github.com/functionless/functionless/issues/246

test/function.localstack.test.ts
sam-goodwin: Maybe we should start the closure serializer in `serialize` instead of constructor? That way the construct tree is fully instantiated at least.

thantos: Was thinking about that...

I don't think serialize would work.

* At bundling happens at the beginning
* The serialize hook seems to be closed off, get an error about using validate
* There is no way to await a promise. 

One thought was to start the promises at the end of construct creation, but before synth, but that would require an explicit command be added to start compilation and would take longer to complete. 

`await Functionless.compileFunctions`

sam-goodwin: What do you mean the serialize hook seems closed off?

thantos: Trying to use onSynthesize threw an error and said to use validate. 

Either way... The async issue is not something we can bypass easily.

test/function.localstack.test.ts
sam-goodwin: > Which of course means that `table.resource.globalSecondaryIndices.length` would fail in a `Function` at runtime.

I get that we erase the `globalSecondaryIndices` data but do we get a chance to inline the `.length` value?

thantos:
Collecting possible token:
* [Resource](https://cs.github.com/aws/aws-cdk/blob/49b397c343b1a2581b2f4d2c098729b21b4b0290/packages/@aws-cdk/core/lib/reference.ts?q=extends+Intrinsic#L11) - reference to a resource's attribute.
* [SecretValue](https://cs.github.com/aws/aws-cdk/blob/49b397c343b1a2581b2f4d2c098729b21b4b0290/packages/@aws-cdk/core/lib/secret-value.ts?q=extends+Intrinsic) - SecretManager, SSM, or plain text
* [Dynamic Reference](https://cs.github.com/aws/aws-cdk/blob/49b397c343b1a2581b2f4d2c098729b21b4b0290/packages/@aws-cdk/core/lib/cfn-dynamic-reference.ts?q=extends+Intrinsic) - 
* [Fn](https://cs.github.com/aws/aws-cdk/blob/49b397c343b1a2581b2f4d2c098729b21b4b0290/packages/@aws-cdk/core/lib/cfn-fn.ts?q=extends+Intrinsic)
* [Lazy](https://cs.github.com/aws/aws-cdk/blob/49b397c343b1a2581b2f4d2c098729b21b4b0290/packages/@aws-cdk/core/lib/lazy.ts#L334)

src/function.ts
thantos: Probably more... but what I found: https://github.com/functionless/functionless/pull/243#issuecomment-1157249992

test/function.localstack.test.ts
thantos: 1. How - I am not sure, but it worked. 
2. Why - Because I was unable to represent some tests and cases when inferring from the expected result.

src/error-code.ts
```diff
@@ -134,4 +134,123 @@ export namespace ErrorCodes {
...
+   * new Function(this, 'func', {
+   *    environment: {
+   *       STREAM_ARN: table.resource.tableStreamArn
+   *    }
+   * }, async () => {
+   *    const tableStreamArn = process.env.STREAM_ARN;
+   * });
+   * ```
+   */
+  export const Cannot_Use_Infra_Resource_In_Function = {
```
sam-goodwin, line 252: Make sure the error code matches the messagetext

`/Infra/Infrastructure/`

src/function.ts
```diff
@@ -752,13 +758,23 @@ export class CallbackLambdaCode extends aws_lambda.Code {
...
-                return token;
+                const reversed = Tokenization.reverse(o)!;
+                if (Reference.isReference(reversed)) {
+                  tokens.push(reversed.toString());
+                  return reversed.toString();
+                } else if (SecretValue.isSecretValue(reversed)) {
+                  throw new SynthError(
+                    ErrorCodes.Unsafe_use_of_secrets,
+                    "Found unsafe use of SecretValue token in a Function."
+                  );
```
sam-goodwin, line 772: Any plans to plumb this through or is it impossible because it needs to be fetched from the secret store and may need to be regularly rotated, so better to do explicit?

src/function.ts
```diff
@@ -783,6 +799,24 @@ export class CallbackLambdaCode extends aws_lambda.Code {
               return o;
             };
 
+            /**
+             * When the StreamArn attribute is used in a Cfn template, but streamSpecification is
+             * undefined, then the deployment fails. Lets make sure that doesn't happen.
+             */
+            const transformBus = (o: CfnResource): CfnResource => {
```
sam-goodwin, line 806: Is this a bus? Docs hint that it is a DynamoDB Strean?

src/function.ts
```diff
@@ -814,10 +847,32 @@ export class CallbackLambdaCode extends aws_lambda.Code {
...
+             */
+            const transformResource = (integ: unknown): any => {
+              if (
+                integ &&
+                (isFunction(integ) || isTable(integ) || isStepFunction(integ))
+              ) {
+                const { resource, ...rest } = integ;
+
+                return rest;
+              } else if (integ && isEventBus(integ)) {
```
sam-goodwin, line 865: nit: Do the "`integ` is truthy" check once

src/table.ts
```diff
@@ -106,11 +106,19 @@ export interface ITable<
   RangeKey extends keyof Item | undefined = undefined
 > {
   readonly kind: "Table";
+  /**
+   * This static property identifies this class as an EventBus to the TypeScript plugin.
+   */
+  readonly functionlessKind: typeof Table.FunctionlessType;
```
sam-goodwin, line 112: Where is this set?

src/validate.ts
```diff
@@ -74,6 +82,49 @@ export function validate(
...
 
+  function validateFunctionNode(node: FunctionInterface) {
+    const func =
+      node.arguments.length === 4 ? node.arguments[3] : node.arguments[2];
+
+    return [
+      // visit all other arguments
+      ...node.arguments
+        .filter((arg) => arg !== func)
+        .flatMap((arg) => validateEachChildRecursive(arg, visit)),
```
sam-goodwin, line 93: Why is `validateFunctionNode` responsible for recursively calling `visit`? This should be delegated higher, no?

src/validate.ts
```diff
@@ -74,6 +82,49 @@ export function validate(
...
+      // process the function closure
+      ...validateEachChildRecursive(func, validateFunctionClosureNode),
```
sam-goodwin, line 95: This is the only line that is specific to this function

src/validate.ts
```diff
@@ -74,6 +82,49 @@ export function validate(
...
+      checker.isStepFunction(node) ||
+      checker.isTable(node) ||
+      checker.isFunctionlessFunction(node)
+    ) {
+      if (
+        typescript.isPropertyAccessExpression(node.parent) &&
+        node.parent.name.text === "resource"
+      ) {
+        return [
+          newError(node, ErrorCodes.Cannot_Use_Infra_Resource_In_Function),
```
sam-goodwin, line 112: Should this be a warning?

src/validate.ts
```diff
@@ -74,6 +82,49 @@ export function validate(
...
+        ];
+      }
+    } else if (checker.isEventBus(node)) {
+      if (
+        typescript.isPropertyAccessExpression(node.parent) &&
+        node.parent.name.text === "bus"
```
sam-goodwin, line 118: Let's make the property names consistent.

sam-goodwin (changes requested):

src/error-code.ts
```diff
@@ -134,4 +134,123 @@ export namespace ErrorCodes {
...
+    code: 106,
+    messageText:
+      "Unsupported initialization of Resources in a Function closure",
+  };
+
+  /**
+   * Cannot use Infrastructure resource in Function closure.
+   *
+   * The `.resource` property of `Function`, `StepFunction`, `ExpressStepFunction`, and `Table` are not available
+   * in Native Function Closures. Likewise `.bus` is not available or `EventBus`es.
```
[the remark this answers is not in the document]
thantos: we could, quite a few references, but I agree with being consistent. Need to do the same to rule which exposes `.rule`.

src/function.ts
thantos: Have a ticket to plumb it through (the error code links to it). It should be possible, just need to turn it into our Secret object or similar. No reason we cannot send a hint with a the env variable that contains the ID

SSM: [id]
Secret: [id]

and then pull from secret manager during runtime. (also permissions)

src/function.ts
thantos: err, yes, also don't need that. Removed

src/validate.ts
thantos: I only want to validate the closure, not the whole new statement. 

Its valid to so something like

```ts
new Function(stack, 'id', { onSuccess: new EventBusSource(bus.resource) }, async () => {

});
```

Might need to refactor step functions for the same later too.

And its also valid to do something like 

```ts
new Function(stack, 'id', { onSuccess: new FunctionSource(new Function(...).resource) }, async () => {

});
```

where the nested function needs to be validated. So we need to ensure all of the args are validated.

The current step function code will run the step function validation over the entire `new StepFunction` including the args and does not validate the args with their own rules. 

Will add tests to prove.

src/validate.ts
thantos: No. Right now we always remove resource, so it will always fail. We can relax it later.

src/validate.ts
thantos: See?

```
    + test/test-files/step-function.ts:57:39 - error Functionless(100): Cannot perform arithmetic on variables in Step Function
    +
    + https://functionless.org/docs/error-codes#cannot-perform-arithmetic-on-variables-in-step-function
    +
    + 57     stateMachineName: `aNamedMachine${stack.region + 2}`,
```

```ts
new StepFunction(
  stack,
  "test",
  {
    stateMachineName: `aNamedMachine${stack.region + 2}`,
  },
  () => {}
);
```

src/validate.ts
thantos: yes, but visitEachChild doesn't support ignoring a specific child and we are looking for the `new Function` expression with arguments. So we are picking out the `new`, finding the special function argument, validating it separately and carrying on the "forEach" with the remaining arguments.

I updated step functions to do the same after finding the bugs in here: https://github.com/functionless/functionless/pull/243#discussion_r899402510

thantos: The other options is that we check each function argument to see if it has a parent of type function/step function and that its the right position in the arguments.  I kind of like this because it would be easier to refactor how a step function behavior or add more cases.

src/validate.ts
thantos: done

src/validate.ts
sam-goodwin: I don't think this belongs in the function. This is the outer function's responsibility.

```ts
.flatMap((arg) => validateEachChildRecursive(arg, visit)),
```

sam-goodwin (approved):

sam-goodwin:
Remember: breaking change

src/validate.ts
thantos: I think we'll end up removing the recursive checker once the rules get more complex, but it works for now. 

(for example, if you need to switch contexts in the step function closure validator, you'll double process everything in the closure for both the new context and the original one with no control over if both contexts are valid).

Not a real example, but lets say the body of an if was different.

```ts
if() { // original rules on the if
    // different rules in the body
}
```

The other option is to track state, but you cannot inject state back into the recursive call, so everything in the `if` body will be seen for both contexts and the original context will not know the nodes are any different.

The exception is to backtrack and look at the parents of the node to determine the context.

PR #251, opened 2022-06-15
chore: upgrade cdk to 2.28

thantos:


sam-goodwin:
I wonder why the upgrader is not doing this? https://github.com/functionless/functionless/pull/245

Also, what's specifically interesting about 2.28?

sam-goodwin:
> I wonder why the upgrader is not doing this? https://github.com/functionless/functionless/pull/245

I wonder if this has to do with how we're coding the versions in `.projenrc.js`?

thantos:
> I wonder if this has to do with how we're coding the versions in `.projenrc.js`?

yeah, but I am not sure projen would be able to upgrade appsync anyways, so we may need to keep doing this manually.

Also, as a library, eventually we'll need to maintain a min version instead of upgrading all of the time.

thantos:
> Also, what's specifically interesting about 2.28?

`SecretValue.isSecretValue` for finding secret token values.

thantos:
> > I wonder if this has to do with how we're coding the versions in `.projenrc.js`?
> 
> yeah, but I am not sure projen would be able to upgrade appsync anyways, so we may need to keep doing this manually.
> 
> Also, as a library, eventually we'll need to maintain a min version instead of upgrading all of the time.

if we cared enough, we could add a script to upgrade the min-cdk version number in the projen file. (or use a flag).

PR #262, opened 2022-06-16
chore: add prettier config

volkanunsal:


thantos:
You'll want to run `yarn default` to apply configuration changes locally.

Looks like you are just applying the defaults? With a `prettierrc.json` already generated, does this change change the formatting of any of the code.

volkanunsal:
> Looks like you are just applying the defaults? With a prettierrc.json already generated, does this change change the formatting of any of the code.

This shouldn't change any of the code. Like you said, it's using the defaults, but it's necessary to generate a `.prettierrc.json` file because contributors may have user defaults that differ from the project defaults and if they also have "format on save" enabled, saving the file will cause massive diffs.

thantos:
Not sure if you saw, but the main branch has a `prettierrc.json` now from #248 

After your suggestion and prettier supporting type 4.7, we cleaned up the config a bit.

thantos:
Looks like projen doesn't even add the defaults to the `prettierrc`, I am fine with this change as it makes the configuration explicit.

volkanunsal:
> Hard to discover. prettier: {} was enough to add the config file and wire up eslint.

How about using `prettier: true` instead? I think that config only needs a truthy value.

volkanunsal:
In fact, that property expects a boolean... There is another property for adding options, per the type def. 

Screen Shot 2022-06-16 at 8 25 40 AM

It's weird that `{}` doesn't cause a type error...

PR #273, opened 2022-06-18
feat: native functions bundle imports

### LINKED ISSUE #264: Is it possible to use a `Function` as a lambda authorizer?
Here is what I'm trying to do:

```ts
const defaultAuthFunction0 = new Function(
  this,
  'LambdaAuthorizer',
  {
    memorySize: 256,
    timeout: Duration.minutes(1),
    logRetention: 14
  },
  async () => ({isAuthorized: true})
);
const defaultAuthFunction = defaultAuthFunction0.resource;


const api = new appsync.GraphqlApi(this, 'API', {
  name: 'API',
  schema,
  authorizationConfig: {
    defaultAuthorization: {
      authorizationType: appsync.AuthorizationType.LAMBDA,
      lambdaAuthorizerConfig: {
        handler: defaultAuthFunction,
      },
    },
  },
  xrayEnabled: true,
  logConfig: {
    fieldLogLevel: FieldLogLevel.ALL,
  },
});
```

But I'm getting a `SynthError`

```
SynthError: Expected input function to Function to be compiled by Functionless. Make sure you have the Functionless compiler plugin configured correctly.
```

### LINKED ISSUE #269: Bug: Function does not bundle require node_modules
From https://github.com/functionless/functionless/issues/264#issuecomment-1158927324

The closure serializer leverages `functionless` as a module import, but we do not bundle any node modules currently.

```ts
function __f1(__0) {
  return (function() {
    with({ function_1: require("functionless/lib/function.js") }) {

return function /*preWarm*/(prewarmContext) {
                        prewarmContext.getOrInit(function_1.PrewarmClients.DYNAMO);
                    };

    }
  }).apply(undefined, undefined).apply(this, arguments);
}
```

We didn't catch this because because tests running locally inlined the contents of functionless instead of importing/requiring it.

thantos:
Fixes #269, Fixes #264 

* Uses esbuild to bundle imports discovered by closure serializer into the function output
* Add `deploymentOnlyModule` to move files to force the closure serialize to serialize the functionless code instead of importing. Importing Functionless can cause a large number of unnecessary imports to be bundled.
* [ ] Add tests and snapshots to show the result of bundling native functions from common imports and functionless integrations.


src/api.ts
```diff
@@ -1110,3 +1110,6 @@ export interface ApiGatewayResponseOverride {
    */
   status: number;
 }
+
+// to prevent the closure serializer from trying to import all of functionless.
+export const deploymentOnlyModule = true;
```
sam-goodwin, line 1115: Hmm this seems hacky. Can we use a comment instead? Are we sure it's necessary. Wouod really rather not have to scatter this around

thantos: This works with the closure serializer right now. I don't think the closure serializer can see comments. I wanted something that works and then will find ways to make it less verbose.

sam-goodwin:
18k lines added?

sam-goodwin:
> 18k lines added?

snapshotting the serialized bundle? Cool, makes a lot of sense.

src/api.ts
sam-goodwin: What's the root cause for the over-inclusion of functionless dependencies?

thantos:
> 

100% much easier to see. can't wait to make those smaller. Plus I need to bypass your new lines commit metrics. ;-)

sam-goodwin:
> Add tests and snapshots to show the result of bundling native functions from common imports and functionless integrations.

I think this is completed right? PR ready to merge?

Have you looked into why netlify is failing to build?

src/api.ts
sam-goodwin: Is `deploymentOnlyModule` a feature enabled by the closure serializer or something you added ust now?

sam-goodwin: nit: If we can control the name, I would prefer `synthTimeOnlyModule`.

sam-goodwin (approved):

src/node.ts
```diff
@@ -349,3 +349,6 @@ export abstract class BaseNode<
     }
   }
 }
+
+// to prevent the closure serializer from trying to import all of functionless.
+export const deploymentOnlyModule = true;
```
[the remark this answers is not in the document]
sam-goodwin: Got it, that's fine for now.

test/serialize.test.ts
```diff
@@ -0,0 +1,236 @@
+// 500k arbitrary max bundle size. Some functions may need more.
+
+import { Stack } from "aws-cdk-lib";
+import { AttributeType } from "aws-cdk-lib/aws-dynamodb";
+import axios from "axios";
+import { v4 } from "uuid";
+import { serialize, bundle, $AWS, Table } from "@fnls";
+
+// In that case increase explicitly.
+const BUNDLED_MAX_SIZE = 500 * 1024;
```
sam-goodwin, line 10: Is this excessive? I know we will slowly improve this over time, but 500KB seems excessive for the tests I see below.

thantos: Can lower it, was arbitrary

.projenrc.js
```diff
@@ -141,13 +143,20 @@ const project = new CustomTypescriptProject({
           exclude: ["./src/{,**}/*"],
         },
       ],
+      paths: {
+        "@fnls": ["lib/index"],
+      },
+      baseUrl: ".",
```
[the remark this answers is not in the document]
sam-goodwin: Oh, it's in `serialize.test.ts`. Is this because of how `yarn link` was hiding what a consumer's behavior would be?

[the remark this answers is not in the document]
sam-goodwin: Got it, now I understand. And it's only in `tsconfigDev` - which i'm happy about. Usually I hate path mappings but this is a useful application of them.

test/serialize.test.ts
thantos: lowered to 50k with the `Aaios` test using 500k as it is 300k bundled.

sam-goodwin: Axios is 300k bundled? gah-dam

PR #275, opened 2022-06-18
feat(appsync): remove addResolver from AppsyncResolver and lazy template evaluation

### LINKED ISSUE #65: AppSync: addResolver breaks if called before dependency is defined
If you call `.addResolver` on an AppsyncResolver before a dependency it references has been instantiated, an error is thrown.

```ts
new AppsyncResolver(() => {
  table.get()
}).addResolver(..);

// this is undefined at the time AppsyncResolver is synthesized
export const table = new f.Table(..);
```

### LINKED ISSUE #137: Remove addResolver utility from AppsyncResolver
The `addResolver` function from an `AppsyncResolver` is an unnecessary extra step for creating a field. Instead, we should change `AppsyncResolver` to be a `Construct` whose instantiation immediately adds Resolvers to a target API.

```ts
const getPerson = new AppsyncResolver(scope, id, {
  api,
  typeName: "Query",
  fieldName: "getPerson"
}, ($context) => {
  // impl
});
```

sam-goodwin:
Fixes #137 
Fixes #65

- [x] AppsyncResolver is now a Construct that takes as props the api, typeName, fieldName and resolve function (#137)
```ts
export const getDeletionStatus = new AppsyncResolver<
  { executionArn: string },
  string | undefined
>(stack, "getDeletionStatus", {
  api,
  typeName: "Query",
  fieldName: "getDeletionStatus",
  resolve: ($context) => {
    const executionStatus = deleteWorkflow.describeExecution(
      $context.arguments.executionArn
    );

    return executionStatus.status;
  },
});
```
- [x] Lazily evaluate VTL interpreters so that fields can be defined in any order (#65)
- [x] Move `addField` to its own class and also apply lazy evaluation

BREAKING CHANGE: AppsyncResolver is now a Construct and creates the resolvers within its constructor

### DIFF
--- a/src/appsync.ts
+++ b/src/appsync.ts
@@ -180,7 +180,7 @@ export class AppsyncResolver<
   Arguments extends ResolverArguments,
   Result,
   Source = undefined
-> extends Construct {
+> {
   /**
    * This static property identifies this class as an AppsyncResolver to the TypeScript plugin.
    */
@@ -238,14 +238,13 @@ export class AppsyncResolver<
         }),
     resolve: ResolverFunction<Arguments, Result, Source>
   ) {
-    super(scope, id);
     this.decl = validateFunctionDecl(resolve, "AppsyncResolver");
 
     const api = props.api ?? (scope as unknown as appsync.GraphqlApi);
 
     this.resolvers = memoize(() => synthResolvers(api, this.decl));
 
-    this.resource = new aws_appsync.CfnResolver(this, "Resource", {
+    this.resource = new aws_appsync.CfnResolver(scope, id, {
       apiId: api.apiId,
       typeName: props.typeName,
       fieldName: props.fieldName,
--- a/src/appsync.ts
+++ b/src/appsync.ts
@@ -147,14 +147,13 @@ export interface AppsyncResolverProps<>extends Pick<
 /**
  * An AWS AppSync Resolver Function derived from TypeScript syntax.
  *
- * First, you must wrap a CDK L2 Construct in the corresponding Functionless type-safe interfaces.
- * ```ts
- * const table = Table.fromTable<Person, "id">(new aws_dynamodb.Table(scope, "id", props));
- * ```
- *
- * Then, call the table from within the new AppsyncResolver:
  * ```ts
  * const getPerson = new AppsyncResolver<{id: string}, Person | undefined>(
+ *   scope, id, {
+ *      api,
+ *      typeName: "Query",
+ *      fieldName: "getPerson"
+ *   },
  *   ($context, id) => {
  *     const person = table.get({
  *       key: {
@@ -163,18 +162,6 @@ export interface AppsyncResolverProps<>extends Pick<
  *     });
  *     return person;
  *   });
- * ```
- *
- * Finally, the `getPerson` function can be used to create resolvers on a GraphQL API
- * ```ts
- * import * as appsync from "@aws-cdk/aws-appsync-alpha";
- *
- * const api = new appsync.GraphQLApi(..);
- *
- * getPerson.createResolver(api, {
- *   typeName: "Query",
- *   fieldName: "getPerson"
- * });
  * ```
  *
  * @functionless AppsyncFunction
@@ -259,7 +246,14 @@ export interface AppsyncFieldOptions
  * const api = new appsync.GraphQLApi(..);
  *
  * ```ts
- * const getPerson = new AppsyncResolver<{id: string}, Person | undefined>(
+ * const getPersonField = new AppsyncField<{id: string}, Person | undefined>(
+ *   {
+ *     api,
+ *     returnType: appsync.Field.string(),
+ *     args: {
+ *       argName: appsync.Field.string()
+ *     }
+ *   },
  *   ($context, id) => {
  *     const person = table.get({
  *       key: {
@@ -273,11 +267,7 @@ export interface AppsyncFieldOptions
  * // code first person type
  * const personType = api.addType(appsync.ObjectType(...));
  *
- * api.addQuery("getPerson", getPerson.getField(api, personType))
- * getPerson.createResolver(api, {
- *   typeName: "Query",
- *   fieldName: "getPerson"
- * });
+ * api.addQuery("getPerson", getPersonField)
  * ```
  *
  * @param api app sync API which the data sources construct will be added under.
--- a/src/appsync.ts
+++ b/src/appsync.ts
@@ -176,12 +176,47 @@ export class AppsyncResolver<
    */
   public static readonly FunctionlessType = "AppsyncResolver";
 
+  /**
+   * The AST form of this resolver.
+   */
   public readonly decl: FunctionDecl<
     ResolverFunction<Arguments, Result, Source>
   >;
 
+  /**
+   * The underlying {@link aws_appsync.CfnResolver}.
+   */
   public readonly resource;
+
+  /**
+   * A memoized function that computes the Appsync Resolvers.
+   *
+   * This function will be called from within a {@link Lazy} during CDK synthesis. It should not be
+   * called directly except during tests.
+   */
   public readonly resolvers;
+
+  /**
+   * Create a new AppsyncResolver and use the {@link scope} as the GraphQL API.
+   *
+   * If {@link props}.api exists, then it will be use as the API instead.
+   */
+  constructor(
+    scope: appsync.GraphqlApi,
+    id: string,
+    props: Omit<AppsyncResolverProps, "api">,
+    resolve: ResolverFunction<Arguments, Result, Source>
+  );
+
+  /**
+   * Create a new AppsyncResolver and use {@link props}.api as the GraphQL API.
+   */
+  constructor(
+    scope: Construct,
+    id: string,
+    props: AppsyncResolverProps,
+    resolve: ResolverFunction<Arguments, Result, Source>
+  );
 
   constructor(
     scope: Construct,
@@ -192,10 +227,12 @@ export class AppsyncResolver<
     super(scope, id);
     this.decl = validateFunctionDecl(resolve, "AppsyncResolver");
 
+    const api = props.api ?? (scope as unknown as appsync.GraphqlApi);
+
     this.resolvers = memoize(() => synthResolvers(props.api, this.decl));
 
     this.resource = new aws_appsync.CfnResolver(this, "Resource", {
-      apiId: props.api.apiId,
+      apiId: api.apiId,
       typeName: props.typeName,
       fieldName: props.fieldName,
       kind: Lazy.string({
@@ -229,12 +266,18 @@ export class AppsyncResolver<
   }
 }
 
+/**
+ * Properties for a new {@link AppsyncField}.
+ */
 export interface AppsyncFieldOptions
   extends Omit<
     appsync.ResolvableFieldOptions,
     keyof ReturnType<typeof synthResolvers>
   > {
-  api: appsync.GraphqlApi;
+  /**
+   * The API this {@link AppsyncField} should be added to.
+   */
+  readonly api: appsync.GraphqlApi;
 }
 
 /**
--- a/src/appsync.ts
+++ b/src/appsync.ts
@@ -1,4 +1,5 @@
 import * as appsync from "@aws-cdk/aws-appsync-alpha";
+import { aws_appsync, Lazy } from "aws-cdk-lib";
 import type { AppSyncResolverEvent } from "aws-lambda";
 import { Construct } from "constructs";
 import {
@@ -32,7 +33,12 @@ import {
 import { findDeepIntegration, IntegrationImpl } from "./integration";
 import { Literal } from "./literal";
 import { FunctionlessNode } from "./node";
-import { AnyFunction, isInTopLevelScope, singletonConstruct } from "./util";
+import {
+  AnyFunction,
+  isInTopLevelScope,
+  memoize,
+  singletonConstruct,
+} from "./util";
 import { visitEachChild } from "./visit";
 import { VTL } from "./vtl";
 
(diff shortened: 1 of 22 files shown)


src/appsync.ts
```diff
@@ -180,62 +198,36 @@ export class AppsyncResolver<
...
-        requestMappingTemplate: fields.requestMappingTemplate,
-        responseMappingTemplate: fields.responseMappingTemplate,
-        pipelineConfig: fields.pipelineConfig,
-      }
-    );
+  constructor(
+    scope: Construct,
+    id: string,
+    props: AppsyncResolverProps<Arguments, Result, Source>
+  ) {
```
thantos, line 208: Could the scope come from the api?

sam-goodwin: That would force all resolvers to exist in the same stack

src/compile.ts
```diff
@@ -220,14 +220,45 @@ export function compile(
       }
 
       function visitAppsyncResolver(call: ts.NewExpression): ts.Node {
-        if (call.arguments?.length === 1) {
-          const impl = call.arguments[0];
-          if (ts.isFunctionExpression(impl) || ts.isArrowFunction(impl)) {
+        if (call.arguments?.length === 3) {
+          const [scope, id, props] = call.arguments;
+          if (ts.isObjectLiteralExpression(props)) {
```
thantos, line 225: Need a friendly error when not an object?

src/compile.ts
```diff
@@ -220,14 +220,45 @@ export function compile(
...
+                ts.factory.updateObjectLiteralExpression(
+                  props,
+                  props.properties.map((prop) => {
+                    if (ts.isPropertyAssignment(prop)) {
+                      if (
+                        (ts.isIdentifier(prop.name) &&
+                          prop.name.text === "resolve") ||
+                        (ts.isStringLiteral(prop.name) &&
+                          prop.name.text === "resolve")
+                      ) {
```
thantos, line 242: join?

src/compile.ts
```diff
@@ -220,14 +220,45 @@ export function compile(
...
+                        const impl = prop.initializer;
+                        if (
+                          ts.isFunctionExpression(impl) ||
+                          ts.isArrowFunction(impl)
+                        ) {
```
thantos, line 247: Error when not a function?

thantos: and validate

src/compile.ts
```diff
@@ -220,14 +220,45 @@ export function compile(
...
+                          return ts.factory.updatePropertyAssignment(
+                            prop,
+                            prop.name,
+                            errorBoundary(() =>
```
thantos, line 251: The error boundary should be around anything that can error.

src/util.ts
```diff
@@ -27,6 +27,24 @@ import { FunctionlessNode } from "./node";
...
+  let t: T;
+  return () => {
+    if (!isComputed) {
+      t = f();
+      isComputed = true;
+    }
+    return t!;
+  };
+}
+
```
thantos, line 47: Nice, I should use this for the `resource` on Rule, which is also lazy resolved and memorized.

test-app/src/message-board.ts
```diff
@@ -59,95 +59,100 @@ const api = new appsync.GraphqlApi(stack, "Api", {
...
-      sk: {
-        S: "Post",
-      },
+const getPost = new AppsyncResolver<{ postId: string }, Post | undefined>(
+  stack,
+  "getPost",
+  {
+    api,
+    typeName: "Query",
+    fieldName: "getPost",
```
thantos, line 68: How does this work with `getField` now. Code first AppSync uses the `add*` methods on the api and just ties the resolver to a single field as the typeName may come from the schema objects.

It also requires adding return and args types to the resolver when binding.

I still think that would all look better being fluent.

```ts
const api = new GqLApi();

api.addResolver("Query", "getPost", { .. props }, () => {
   ....
});


api.addQuery("something", appsync.GrapQLType.string(), { args: {... } }, () => {
   ....
});
```

src/appsync.ts
```diff
@@ -180,62 +198,36 @@ export class AppsyncResolver<
...
-        dataSource: fields.dataSource,
-        requestMappingTemplate: fields.requestMappingTemplate,
-        responseMappingTemplate: fields.responseMappingTemplate,
-        pipelineConfig: fields.pipelineConfig,
-      }
-    );
+  constructor(
+    scope: Construct,
+    id: string,
+    props: AppsyncResolverProps<Arguments, Result, Source>
```
thantos, line 207: Our patterns has been to put the function on the method, should we do that here?

```ts
new AppSyncResolver(
   scope,
   id,
   api,
   {  }, // optional props
   ($context) => { ... }
)
```

thantos: This would match EB, Sfn, Function, etc.

test-app/src/message-board.ts
```diff
@@ -565,7 +576,7 @@ post.addField({
 
 api2.addQuery(
   "getPost",
-  getPostResolver.getField(api2, post.attribute(), {
+  getPost.getField(api2, post.attribute(), {
```
thantos, line 579: What does it mean to add the resolver twice here? Its already added to `Query getPost` on api above.

thantos (changes requested):

src/compile.ts
sam-goodwin: Ts compiler should error for us

test-app/src/message-board.ts
sam-goodwin: Yeah that's my outstanding task. I was thinking of creating its own class. We could also add methods to our own GraphQLApi class

src/appsync.ts
sam-goodwin: Ah that's true, although I did like the resolve function. We should be consistent

src/compile.ts
thantos: Are there not other valid ways? like references?

src/appsync.ts
thantos: I mean as an option, with event bus many of the methods have the scope as optional with the bus being used when not provided. obv just a nice to have.

src/compile.ts
sam-goodwin: It is.

src/compile.ts
sam-goodwin: Decided against this as it breaks our tests. Will instead rely on a static validation error. Also, moved the function to the 4th argument.

sam-goodwin: It breaks our tests because we have a util that passes the FunctionDecl and casts to any. This breaks.

test-app/src/message-board.ts
sam-goodwin: Not sure, I don't remember putting this here. I assume it would error. Should this be on api2?

test-app/src/message-board.ts
sam-goodwin: Moved it to `AppsyncField` class

src/appsync.ts
```diff
@@ -180,391 +193,385 @@ export class AppsyncResolver<
...
+ *
+ * // code first person type
+ * const personType = api.addType(appsync.ObjectType(...));
+ *
+ * api.addQuery("getPerson", getPerson.getField(api, personType))
+ * getPerson.createResolver(api, {
+ *   typeName: "Query",
+ *   fieldName: "getPerson"
+ * });
+ * ```
```
sam-goodwin, line 281: Oops, these are not updated. That reminds me - i need to update the website docs too

src/appsync.ts
```diff
@@ -152,419 +164,447 @@ export class AppsyncVTL extends VTL {
...
+    resolve: ResolverFunction<Arguments, Result, Source>
+  ) {
+    super(scope, id);
+    this.decl = validateFunctionDecl(resolve, "AppsyncResolver");
+
+    const api = props.api ?? (scope as unknown as appsync.GraphqlApi);
+
+    this.resolvers = memoize(() => synthResolvers(props.api, this.decl));
+
+    this.resource = new aws_appsync.CfnResolver(this, "Resource", {
```
thantos, line 234: why CfnResolver and not he L2?

src/appsync.ts
```diff
@@ -152,419 +164,447 @@ export class AppsyncVTL extends VTL {
...
-    if (resolverCount === 0) {
-      if (pipelineConfig.length !== 0) {
-        throw new Error(
-          `expected 0 functions in pipelineConfig, but found ${pipelineConfig.length}`
-        );
-      }
+  // mock integration
+  if (resolverCount === 0) {
+    if (pipelineConfig.length !== 0) {
+      throw new Error(
```
thantos, line 357: In my PR I removed this check. I am not sure I understand the purpose, do we think we'll mess up? Couldn't an integration provide only vtl in the future?

src/appsync.ts
```diff
@@ -152,419 +164,447 @@ export class AppsyncVTL extends VTL {
...
+              return returnValName;
+            } else if (isPropAccessExpr(expr)) {
+              return `${getResult(expr.expr)}.${expr.name}`;
+            } else if (isElementAccessExpr(expr)) {
+              return `${getResult(expr.expr)}[${getResult(expr.element)}]`;
             } else {
-              template.return("$null");
+              throw new Error(
+                `invalid Expression in-lined with Service Call: ${expr.kind}`
+              );
```
thantos, line 559: Synth error, also I don't understand how this happens?

src/appsync.ts
```diff
@@ -152,419 +164,447 @@ export class AppsyncVTL extends VTL {
...
+> extends appsync.ResolvableField {
+  /**
+   * This static property identifies this class as an AppsyncResolver to the TypeScript plugin.
+   */
+  public static readonly FunctionlessType = "AppsyncField";
+
+  constructor(
+    options: AppsyncFieldOptions,
+    resolve: ResolverFunction<Arguments, Result, Source>
+  ) {
```
thantos, line 333: I was trying to remember why field and resolver are different.

Ha...

https://github.com/aws/aws-cdk/issues/18351#issuecomment-1115266100

I don't like how buggy it is, ran into this again when working on the authorizer example.

Can you add a validator or something that checks to see if a resolver was actually created?

test/__snapshots__/vtl.test.ts.snap
```diff
@@ -749,8 +749,8 @@ $util.qr($v1.put('value', $null))
 exports[`property assignment of conditional expression 1`] = `
 Array [
   "{
-  \\"version\\": \\"2018-05-29\\",
-  \\"payload\\": null
+\\"version\\": \\"2018-05-29\\",
+\\"payload\\": null
```
thantos, line 753: You meant to do this?

test/test-files/appsync.ts
```diff
@@ -0,0 +1,66 @@
...
+  stack,
+  "getPost",
+  {
+    api,
+    typeName: "Query",
+    fieldName: "getPost",
+  },
+  // valid - inline arrow function
+  () => null
+);
```
thantos, line 24: Love these tests.

They do get stuck periodically though

image

thantos (changes requested):

test/__snapshots__/vtl.test.ts.snap
sam-goodwin: I don't recall changing the vtl logic

test/test-files/appsync.ts
sam-goodwin: Hmm that's odd, any idea why that would be? I've never seen that happen.

src/appsync.ts
```diff
@@ -152,419 +164,447 @@ export class AppsyncVTL extends VTL {
...
- * ```
- *
  * @functionless AppsyncFunction
  */
 export class AppsyncResolver<
   Arguments extends ResolverArguments,
   Result,
   Source = undefined
-> {
+> extends Construct {
```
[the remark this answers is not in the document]
sam-goodwin: Not sure we would ever call a resolved directly. I don't think it's even possible? We can execute graphql queries but not individual resolvers.

src/appsync.ts
sam-goodwin: Yeah, I don't really know what this is checking for either. Seems to check for a situation where we rendered a pipeline without ever calling a resolver? Seems like internal bug.

sam-goodwin: Which PR should we merge first?

src/appsync.ts
sam-goodwin: I haven't changed this code. It is a bit hacky - it is trying to support various different ways in which the result of an Integration call may be referenced.

E.g. 

```ts
call().result
call().result[0]
```

This will be fixed with your a-normal form change? https://github.com/functionless/functionless/pull/280

src/appsync.ts
sam-goodwin: Do you mean a node validator? There is no `Construct` here for me to add that validator to. Also, if the resolver isn't added, then the validator also wouldn't be added? Is this solved with just tests?

test/test-files/appsync.ts
thantos: I am not sure, never get an error, just spins. Running it on its own works fine, maybe a race condition?

test/__snapshots__/vtl.test.ts.snap
thantos: I think it was a literal string change.

src/appsync.ts
thantos: Hmm, you could put it on the api.

```ts
api.node.addValidator(() => {
    api.node.tryFindChild(`${props.typeName}${props.fieldName}Resolver`) ? [] : ["uhh ohh no resolver!"]
});
```

src/appsync.ts
thantos: Ohh, yeah, I actually removed this whole function, lol

thantos: in the async change.

src/appsync.ts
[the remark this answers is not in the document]
thantos: I guess if there is value in this being a construct, it is fine.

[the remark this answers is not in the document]
sam-goodwin: I'm also fine with moving it to `resource`. Keep things consistent.

src/appsync.ts
thantos: Makes sense, I think we'll need to do this in other places.

src/appsync.ts
sam-goodwin: I don't have a `typeName` or `fieldName` at this point. That happens later when you call `addQuery`. I don't think there's anything to validate here?

thantos (approved):

PR #280, opened 2022-06-19
fix: appsync and step function visitor hoist refactor

### LINKED ISSUE #212: Bug: App Sync: Nested Integrations
Currently returns a broken VTL template, should return 8 templates, 2 for each of the 3 integrations, one before and one after.

```ts
test("multiple nested integrations", () => {
  const machine = new StepFunction(stack, "machine", () => {});

  appsyncTestCase(
    reflect(() => {
      machine({ input: machine({ input: machine({ input: {} }) }) });
    }),
    {
      expectedTemplateCount: 8,
      executeTemplates: [
        {
          index: 1,
        },
      ],
    }
  );
});
```

thantos:
Closes #212

Refactor step functions and app sync interpreters to refactor nested tasks to a normal form. This change will make the #202 easier by providing a generic interface for re-writing the AST tree. Async introduces more complex statements that need to be hoisted to a-normal form which would have been previously difficult.

### DIFF
--- a/src/appsync.ts
+++ b/src/appsync.ts
@@ -28,12 +28,26 @@ import {
   isPropAccessExpr,
   isElementAccessExpr,
   isIfStmt,
+  isBlockStmt,
+  isStmt,
+  isForInStmt,
+  isForOfStmt,
 } from "./guards";
-import { findDeepIntegration, IntegrationImpl } from "./integration";
+import {
+  findDeepIntegration,
+  IntegrationImpl,
+  isIntegrationCall,
+} from "./integration";
 import { Literal } from "./literal";
 import { FunctionlessNode } from "./node";
-import { AnyFunction, isInTopLevelScope, singletonConstruct } from "./util";
-import { visitEachChild } from "./visit";
+import { BlockStmt } from "./statement";
+import {
+  AnyFunction,
+  DeterministicNameGenerator,
+  isInTopLevelScope,
+  singletonConstruct,
+} from "./util";
+import { visitBlock, visitEachChild } from "./visit";
 import { VTL } from "./vtl";
 
 /**
@@ -180,6 +194,8 @@ export class AppsyncResolver<
     ResolverFunction<Arguments, Result, Source>
   >;
 
+  private generatedNames = new DeterministicNameGenerator();
+
   constructor(fn: ResolverFunction<Arguments, Result, Source>) {
     this.decl = validateFunctionDecl(fn, "AppsyncResolver");
   }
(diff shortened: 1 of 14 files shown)


src/visit.ts
```diff
@@ -480,3 +490,29 @@ export function visitEachChild<T extends FunctionlessNode>(
...
+  cb: (stmt: Stmt, hoist: (expr: Expr) => Expr) => Stmt,
+  nameGenerator: DeterministicNameGenerator
+): BlockStmt {
+  return visitEachChild(block, (stmt) => {
+    const nestedTasks: FunctionlessNode[] = [];
+    function hoist(expr: Expr): Expr {
+      const id = new Identifier(nameGenerator.generateOrGet(expr));
+      nestedTasks.push(new VariableStmt(id.name, expr));
+      return id;
+    }
```
sam-goodwin, line 510: Why would we do it this way instead of transforming the tree before walking?

thantos: We are, this loop is running before walking it, it just reshapes the tree before the interpreter runs on it. Its 1:1 with the previous pre-interpreter transform, it just supports arbitrary depth, complexity, and location of integrations.

```ts
for(const i in [func(), func(), func(func())])
```

out

```ts
const 0_expr = func();
const 1_expr = func();
const 2_expr = func();
const 3_expr = func(2_expr);
for(const i in [0_expr, 1_expr, 3_expr])
```

sam-goodwin: I mean why is the hoist running within the generic visit function?

Also, how does it handle `ConditionExpr`?

```ts
for (const i in [cond ? expr_1 : expr_2])
```

thantos: > Also, how does it handle ConditionExpr?

it should? can add a test. It won't short circuit, but that's a risk of any of the hoisting. We should probably make the actually processing smarter and get rid of the a-normal form at some point, or also inline the conditional statements.

We could also handle specific cases differently like I'll show with async.

But I don't know the difference between

```ts
func(func2())
```

and 

```ts
func(x ? func() : func2())
```

Could also turn turnary statements into if-statements to maintain scoping instead of the a-normal form.

> I mean why is the hoist running within the generic visit function?

Its not running, its an optional callback. When processing a block statement, calls can be hoisted.

thantos: An attempt to generalize this

https://github.com/functionless/functionless/blob/main/src/asl.ts#L403-L450

thantos: lol

(step functions)

image

src/appsync.ts
```diff
@@ -295,20 +311,267 @@ export class AppsyncResolver<
     });
   }
 
-  private getResolvableFields(api: appsync.GraphqlApi) {
-    const resolverCount = countResolvers(this.decl);
+  private synthesizeFunctions(
```
sam-goodwin, line 314: FYI - I moved this out of the class as a module-scoped function

thantos: Which PR?

thantos: #275 I imagine, will review once I stop the self mutation loop

src/asl.ts
```diff
@@ -868,38 +834,32 @@ export class ASL {
...
+                  ResultPath: throwOrPass.ResultPath,
+                },
+              ],
+            };
+          } else {
+            return taskState;
+          }
         } else {
-          return taskState;
+          throw Error("");
```
sam-goodwin, line 862: What should error message be? Why did the logic change?

src/util.ts
```diff
@@ -273,3 +273,22 @@ export const evalToConstant = (expr: Expr): Constant | undefined => {
...
+   * Generate a deterministic and unique variable name for a node.
+   *
+   * The value is cached so that the same node reference always has the same name.
+   *
+   * @param node the node to generate a name for
+   * @returns a unique variable name that can be used in JSON Path
+   */
+  public generateOrGet(node: FunctionlessNode): string {
+    if (!this.generatedNames.has(node)) {
+      this.generatedNames.set(node, `${this.generatedNames.size}_tmp`);
```
sam-goodwin, line 290: Look at who is using mutability ;)

src/visit.ts
```diff
@@ -480,3 +490,29 @@ export function visitEachChild<T extends FunctionlessNode>(
...
+ * Provides the hoist function that allows hoisting expressions into variable statements above the current statement.
+ */
+export function visitBlock(
+  block: BlockStmt,
+  cb: (stmt: Stmt, hoist: (expr: Expr) => Expr) => Stmt,
+  nameGenerator: DeterministicNameGenerator
+): BlockStmt {
+  return visitEachChild(block, (stmt) => {
+    const nestedTasks: FunctionlessNode[] = [];
+    function hoist(expr: Expr): Expr {
```
sam-goodwin, line 506: Should the return type be an Identifier?

src/visit.ts
```diff
@@ -480,3 +490,29 @@ export function visitEachChild<T extends FunctionlessNode>(
   }
   return assertNever(node);
 }
+
+/**
+ * Like {@link visitEachChild} but it only visits the statements of a block.
+ *
+ * Provides the hoist function that allows hoisting expressions into variable statements above the current statement.
```
sam-goodwin, line 497: Pretty nice I must say. Neat.

test/__snapshots__/table.test.ts.snap
```diff
@@ -608,12 +608,12 @@ $util.qr($v5.put('consistentRead', $v1.get('consistentRead')))
 $util.qr($v5.put('select', $v1.get('select')))
 #end
 $util.toJson($v5)",
-  "#set( $context.stash.return__flag = true )
-#set( $context.stash.return__val = $context.result.items )
+  "#set( $context.stash.0_tmp = $context.result )
```
sam-goodwin, line 611: Is this a regression?

src/visit.ts
thantos: ohh, yeah, it could be, it was more complex at one point.

src/util.ts
thantos: I copied your code ;-) but in this case the risk seemed low in a small isolated private methods.

test/__snapshots__/table.test.ts.snap
thantos: Good question, I am going to say "no" because it follows the rules.

Previously 

```ts
return table.query().items
```

now

```ts
var 0_temp = table.query()
// template split
return 0_temp.items;
```

Because we say the only integration calls we keep inline are those directly on the statements. 

This strategy alights with arbitrary complexity before or after the integrations

```ts
return table.query.items[0].S;
```

src/util.ts
sam-goodwin: > I copied your code ;-)


Loool

src/asl.ts
thantos: got rid of `findIntegration` which will align better with validating and handling the async change. Will fix error.

sam-goodwin (approved):

PR #293, opened 2022-06-22
chore: upgrade closure-serializer and add more tests

thantos:
Uses the latest version of the closure serializer and adds more integration tests that output serialized and bundled handler code.

Use of Sfn in the closure has reduced from 1000s of lines to hundreds (less than 50k).


sam-goodwin:
> less than 50k

Is that what it is now or used to be? Still seems high? What's the remaining bloat?

test/__snapshots__/serialize.test.ts.snap
```diff
@@ -12380,6 +12380,1890 @@ function __f0() {
...
+                    return dynamo
+                        .putItem({
+                        ...rest,
+                        Item: Item,
+                        TableName: table.tableName,
+                    })
+                        .promise();
+                };
+
+    }
```
sam-goodwin, line 12585: Does it pull in all of the table's native methods even if they aren't used?

test/__snapshots__/serialize.test.ts.snap
```diff
@@ -12380,6 +12380,1890 @@ function __f0() {
...
+                    Entries: [
+                        {
+                            EventBusName: bus.eventBusName,
+                        },
+                    ],
+                });
+            };
+
+    }
+  }).apply(undefined, undefined).apply(this, arguments);
```
sam-goodwin, line 12723: Does this have any performance impact on V8's ability to optimize this code?

test/__snapshots__/serialize.test.ts.snap
```diff
@@ -12380,6 +12380,1890 @@ function __f0() {
 "
 `;
 
+exports[`serialize event bridge aws put events 1`] = `
```
sam-goodwin, line 12383: Looks significantly less than 50k, but i am totally eye balling.

test/serialize.test.ts
```diff
@@ -224,6 +232,104 @@ describe("serialize", () => {
...
+        () => () => {
+          return sfn({ input: {} });
+        },
+        []
+      );
+      expect(srlz).toMatchSnapshot();
+
+      const bundled = await bundle(srlz);
+      expect(bundled.text).toMatchSnapshot();
+      expect(bundled.text).toHaveLengthLessThan(BUNDLED_MAX_SIZE);
```
sam-goodwin, line 250: Should we do this per test instead of capping each test at the same amount?

sam-goodwin (approved):

test/__snapshots__/serialize.test.ts.snap
thantos: No idea, its what the closure serializer generates and have not played with it yet. Not 100% sure they need to wrap all of the function like this.

test/__snapshots__/serialize.test.ts.snap
thantos: yeah, I didn't look, just didn't fail the test (50k).

test/serialize.test.ts
thantos: I set 50k as the max and then up them as needed. Could try lower and see what breaks.

test/__snapshots__/serialize.test.ts.snap
thantos: Yes, originally I thought this was due to the namespace, but I changed to a constant and it still happened. Not sure why it cannot tree shake here.

test/__snapshots__/serialize.test.ts.snap
sam-goodwin: Yeah I know we inherited this. Thought I would ask the question anyway. I'm curious of the answer.

test/__snapshots__/serialize.test.ts.snap
sam-goodwin: Time to ratchet that down some more?

test/__snapshots__/serialize.test.ts.snap
thantos: yeah, same, slowly making changes.

thantos:
> > less than 50k
> 
> Is that what it is now or used to be? Still seems high? What's the remaining bloat?

looks like 10-15k is above average. with the tail being axios at 300k...

sam-goodwin (approved):

PR #295, opened 2022-06-23
fix: remove .only on function test, pin localstack

thantos:
* @thantos forgot to remove `.only` from a test. Do that now
* add eslint rule to remove `.only`
* pin localstack cli version https://github.com/localstack/localstack/issues/6334
* run localstack tests

### DIFF
--- a/test/localstack.ts
+++ b/test/localstack.ts
@@ -157,6 +157,7 @@ export const localstackTestSuite = (
   testResource.skip = (name, resources, test) => {
     tests.push({ name, resources, test: test as any, skip: true, only: false });
   };
+  // eslint-disable-next-line no-only-tests/no-only-tests
   testResource.only = (name, resources, test) => {
     tests.push({ name, resources, test: test as any, skip: false, only: true });
   };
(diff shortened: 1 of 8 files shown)


test/localstack.ts
```diff
@@ -157,6 +157,7 @@ export const localstackTestSuite = (
   testResource.skip = (name, resources, test) => {
     tests.push({ name, resources, test: test as any, skip: true, only: false });
   };
+  // eslint-disable-next-line no-only-tests/no-only-tests
   testResource.only = (name, resources, test) => {
```
sam-goodwin, line 161: Why is the linter finding this? Does it alert on any call of .only?

sam-goodwin (approved):

test/localstack.ts
thantos: not sure, actually, the plugin claims to search this list by default: `["describe", "it", "context", "test", "tape", "fixture", "serial"]`

But it "fixed" that line :(

thantos: Looks like the plugin is kind of dumb

https://github.com/levibuzolic/eslint-plugin-no-only-tests/blob/master/rules/no-only-tests.js#L65-L71

PR #299, opened 2022-06-24
fix(appsync): various appsync and related bugs

### LINKED ISSUE #8: Appsync: Support push with var-args
Support the following syntaxes:

```ts
list.push(a, b);
list.push(...a, b);
list.push(a, ...b);
list.push(a, ...b, c);
```

### LINKED ISSUE #68: Step Functions: Support destructured variable declarations
We need to support destructuring objects:

```ts
new StepFunction(({key} : Type => {
  const {val} = ..
}) 
```

### LINKED ISSUE #150: Correctly handle assignment expressions inside object puts in VTL
Syntax like this is uncommon but valid.

```
let y = 1;
return { x: (y = 2) };
```

This is converted to

```
#set($context.stash.y = 1)
#set($v1 = {})
#set($v2 = $context.stash.y = 2)
```

Which does not evaluate. We'll need to treat BinaryExpr with = as the op as a special case.

### LINKED ISSUE #232: App Sync: Support additional binary/unary expressions
Binary:
* `+=`
* `-=`
* `*=`
* `/=`
* `%=`
* `%`

Unary:
* `++` (pre and post)
* `--` (pre and post)

### LINKED ISSUE #235: Bug: App Sync Deconstruction fails at runtime
```ts
const {a, b} = $context.arguments;
```

results in

```vtl
#set($context.stash.{ a, b } = $context.arguments)
```

### LINKED ISSUE #247: App Sync: Compile console.log and console.error to `$util.log.*`
Extends #231 to support natural  typescript console commands.

### LINKED ISSUE #265: Bug: `Function` doesn't support blockless if statements
```
Error: a IfStmt's then must be a BlockStmt
```

Here is the code that produced it.

```ts
if (id && 'resolverContext' in id)
      return table.getItem({
        key: { pk: { S: `ACCOUNT#${id}` }, sk: { S: `ACCOUNT#${id}` } },
      });
```

Fixed it by adding a block after if.

### LINKED ISSUE #266: $util.error causes "Lexical error"
Here is my code

```ts
const getAccount = new AppsyncResolver<{}, IAccount>(($context) => {
    const identity = $context.identity;

    if (identity && 'resolverContext' in identity) {
      const { accountId } = identity.resolverContext;
      return table.getItem({
        key: {
          pk: { S: `ACCOUNT#${accountId}` },
          sk: { S: `ACCOUNT#${accountId}` },
        },
      });
    }

    return $util.error('Cannot find account.');
  });
```

And this is the error:

```
[{"path":["account"],"data":null,"errorType":"MappingTemplate","errorInfo":null,"locations":[{"line":1,"column":17,"sourceName":null}],"message":"Lexical error,   Encountered: \"{\" (123), after : \".\" at *unset*[line 11, column 21]"}]
```

### LINKED ISSUE #300: App Sync: Improve error and add validation for non-deterministic validation calls.
From: https://github.com/functionless/functionless/issues/266

thantos:
Closes #266 Closes #265 Closes #150 Closes #8 Closes #235 Closes #300
Partial #232

* [ ] #266 
* [x] #265 
* #232 
   * [x] AppSync
   * Sync - New PR
* [ ] #150 
* [x] #8 
* [x] #235
* [ ] #300

### DIFF
--- a/src/appsync.ts
+++ b/src/appsync.ts
@@ -552,8 +552,9 @@ function synthesizeFunctions(api: appsync.GraphqlApi, decl: FunctionDecl) {
             } = ${getResult(stmt.expr)} )\n{}`
           );
         } else {
-          throw new Error(
-            "only a 'VariableDecl', 'Call' or 'Return' expression may call a service"
+          debugger;
+          throw new SynthError(
+            ErrorCodes.Appsync_Integration_invocations_must_be_deterministic
           );
         }
 
--- a/src/appsync.ts
+++ b/src/appsync.ts
@@ -566,7 +566,6 @@ function synthesizeFunctions(api: appsync.GraphqlApi, decl: FunctionDecl) {
             } = ${getResult(stmt.expr)} )\n{}`
           );
         } else {
-          debugger;
           throw new SynthError(
             ErrorCodes.Appsync_Integration_invocations_must_be_deterministic
           );
--- a/src/appsync.ts
+++ b/src/appsync.ts
@@ -36,6 +36,7 @@ import {
   isAwaitExpr,
   isPromiseExpr,
   isBindingElem,
+  isBindingPattern,
 } from "./guards";
 import {
   findDeepIntegrations,
@@ -498,6 +499,18 @@ function synthesizeFunctions(api: appsync.GraphqlApi, decl: FunctionDecl) {
     resolverCount === 0
       ? new AppsyncVTL()
       : new AppsyncVTL(AppsyncVTL.CircuitBreaker);
+
+  if (
+    updatedDecl.parameters.length > 0 &&
+    isBindingPattern(updatedDecl.parameters[0].name)
+  ) {
+    template.evaluateBindingPattern(
+      updatedDecl.parameters[0].name,
+      "$context",
+      "$context.stash."
+    );
+  }
+
   const functions = updatedDecl.body.statements
     .map((stmt, i) => {
       const isLastExpr = i + 1 === updatedDecl.body.statements.length;
--- a/src/appsync.ts
+++ b/src/appsync.ts
@@ -35,6 +35,7 @@ import {
   isForOfStmt,
   isAwaitExpr,
   isPromiseExpr,
+  isBindingElem,
 } from "./guards";
 import {
   findDeepIntegrations,
@@ -138,7 +139,7 @@ export class AppsyncVTL extends VTL {
 
   protected dereference(id: Identifier): string {
     const ref = id.lookup();
-    if (isVariableStmt(ref) && isInTopLevelScope(ref)) {
+    if ((isVariableStmt(ref) || isBindingElem(ref)) && isInTopLevelScope(ref)) {
       return `$context.stash.${id.name}`;
     } else if (isParameterDecl(ref) && isFunctionDecl(ref.parent)) {
       // regardless of the name of the first argument in the root FunctionDecl, it is always the intrinsic Appsync `$context`.
--- a/src/appsync.ts
+++ b/src/appsync.ts
@@ -567,7 +567,7 @@ function synthesizeFunctions(api: appsync.GraphqlApi, decl: FunctionDecl) {
           );
         } else {
           throw new SynthError(
-            ErrorCodes.Appsync_Integration_invocations_must_be_deterministic
+            ErrorCodes.Appsync_Integration_invocations_must_be_unidirectional_and_defined_statically
           );
         }
 
--- a/src/checker.ts
+++ b/src/checker.ts
@@ -7,6 +7,7 @@ import { EventTransform } from "./event-bridge/transform";
 import { Function } from "./function";
 import { ExpressStepFunction, StepFunction } from "./step-function";
 import { Table } from "./table";
+import { anyOf } from "./util";
 
 /**
  * Various types that could be in a call argument position of a function parameter.
@@ -90,8 +91,43 @@ export type FunctionlessChecker = ReturnType<typeof makeFunctionlessChecker>;
 export function makeFunctionlessChecker(
   checker: ts.TypeChecker | tsserver.TypeChecker
 ) {
+  const isStatement = anyOf(
+    ts.isVariableStatement,
+    ts.isEmptyStatement,
+    ts.isExpressionStatement,
+    ts.isIfStatement,
+    ts.isDoStatement,
+    ts.isWhileStatement,
+    ts.isForStatement,
+    ts.isForInStatement,
+    ts.isForOfStatement,
+    ts.isContinueStatement,
+    ts.isBreakStatement,
+    ts.isReturnStatement,
+    ts.isWithStatement,
+    ts.isSwitchStatement,
+    ts.isLabeledStatement,
+    ts.isThrowStatement,
+    ts.isTryStatement,
+    ts.isDebuggerStatement
+  );
+
+  /**
+   * Used by AppSync to determine of a root statement is deterministic or now.
+   *
+   * func();
+   * const v = func();
+   * return func();
+   */
+  const isDeterministicStatement = anyOf(
+    ts.isExpressionStatement,
+    ts.isReturnStatement,
+    ts.isVariableStatement
+  );
+
   return {
     ...checker,
+    findParent,
     getApiMethodKind,
     getFunctionlessTypeKind,
     isApiIntegration,
@@ -99,6 +135,7 @@ export function makeFunctionlessChecker(
     isAppsyncResolver,
     isCDKConstruct,
     isConstant,
+    isDeterministicStatement,
     isEventBus,
     isEventBusWhenFunction,
     isFunctionlessFunction,
@@ -113,6 +150,7 @@ export function makeFunctionlessChecker(
     isPromiseSymbol,
     isReflectFunction,
     isRuleMapFunction,
+    isStatement,
     isStepFunction,
     isTable,
   };
@@ -398,6 +436,28 @@ export function makeFunctionlessChecker(
       });
     }
     return false;
+  }
+
+  /**
+   * Invokes a predicate on the ancestors of the given node until:
+   * * the predicate is true
+   * * there is no parent
+   * * or the limit parameter is hit
+   *
+   * @param limit - An upper bound node to stop processing at.
+   */
+  function findParent<T extends ts.Node>(
+    node: ts.Node,
+    predicate: (node: ts.Node) => node is T,
+    limit?: ts.Node
+  ): T | undefined {
+    if (!node.parent || (limit && limit === node.parent)) {
+      return undefined;
+    } else if (predicate(node.parent)) {
+      return node.parent;
+    } else {
+      return findParent(node.parent, predicate);
+    }
   }
 
   /**
--- a/src/checker.ts
+++ b/src/checker.ts
@@ -7,7 +7,6 @@ import { EventTransform } from "./event-bridge/transform";
 import { Function } from "./function";
 import { ExpressStepFunction, StepFunction } from "./step-function";
 import { Table } from "./table";
-import { anyOf } from "./util";
 
 /**
  * Various types that could be in a call argument position of a function parameter.
@@ -91,43 +90,8 @@ export type FunctionlessChecker = ReturnType<typeof makeFunctionlessChecker>;
 export function makeFunctionlessChecker(
   checker: ts.TypeChecker | tsserver.TypeChecker
 ) {
-  const isStatement = anyOf(
-    ts.isVariableStatement,
-    ts.isEmptyStatement,
-    ts.isExpressionStatement,
-    ts.isIfStatement,
-    ts.isDoStatement,
-    ts.isWhileStatement,
-    ts.isForStatement,
-    ts.isForInStatement,
-    ts.isForOfStatement,
-    ts.isContinueStatement,
-    ts.isBreakStatement,
-    ts.isReturnStatement,
-    ts.isWithStatement,
-    ts.isSwitchStatement,
-    ts.isLabeledStatement,
-    ts.isThrowStatement,
-    ts.isTryStatement,
-    ts.isDebuggerStatement
-  );
-
-  /**
-   * Used by AppSync to determine of a root statement is deterministic or now.
-   *
-   * func();
-   * const v = func();
-   * return func();
-   */
-  const isDeterministicStatement = anyOf(
-    ts.isExpressionStatement,
-    ts.isReturnStatement,
-    ts.isVariableStatement
-  );
-
   return {
     ...checker,
-    findParent,
     getApiMethodKind,
     getFunctionlessTypeKind,
     isApiIntegration,
@@ -135,7 +99,6 @@ export function makeFunctionlessChecker(
     isAppsyncResolver,
     isCDKConstruct,
     isConstant,
-    isDeterministicStatement,
     isEventBus,
     isEventBusWhenFunction,
     isFunctionlessFunction,
@@ -150,7 +113,6 @@ export function makeFunctionlessChecker(
     isPromiseSymbol,
     isReflectFunction,
     isRuleMapFunction,
-    isStatement,
     isStepFunction,
     isTable,
   };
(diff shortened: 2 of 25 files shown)


sam-goodwin:
Commit title should probably not have the word "random" on it

thantos:
> Commit title should probably not have the word "random" on it

No error from the linter, must be fine.

src/checker.ts
```diff
@@ -400,6 +438,28 @@ export function makeFunctionlessChecker(
...
+  ): T | undefined {
+    if (!node.parent || (limit && limit === node.parent)) {
+      return undefined;
+    } else if (predicate(node.parent)) {
+      return node.parent;
+    } else {
+      return findParent(node.parent, predicate);
+    }
+  }
+
```
sam-goodwin, line 462: This doesn't use any aspect of the type checker (only the ast). Does it belong in the checker?

src/compile.ts
```diff
@@ -556,10 +558,17 @@ export function compile(
...
+            return newExpr("VariableStmt", [
+              ts.factory.createStringLiteral(node.name.getText()),
+              ...(node.initializer ? [toExpr(node.initializer, scope)] : []),
+            ]);
+          } else {
+            return newExpr("VariableStmt", [
+              toExpr(node.name, scope),
+              toExpr(node.initializer, scope),
+            ]);
+          }
```
sam-goodwin, line 571: Our VariableStmt node is not 1:1 with typescript's, should we change it? Would that remove this special case code?

src/compile.ts
```diff
@@ -593,7 +628,9 @@ export function compile(
         } else if (ts.isPrefixUnaryExpression(node)) {
           if (
             node.operator !== ts.SyntaxKind.ExclamationToken &&
-            node.operator !== ts.SyntaxKind.MinusToken
+            node.operator !== ts.SyntaxKind.MinusToken &&
+            node.operator !== ts.SyntaxKind.MinusMinusToken &&
+            node.operator !== ts.SyntaxKind.PlusPlusToken
```
sam-goodwin, line 633: Should this be a constant like we did for binary operators?

src/compile.ts
```diff
@@ -608,6 +645,16 @@ export function compile(
             ),
             toExpr(node.operand, scope),
           ]);
+        } else if (ts.isPostfixUnaryExpression(node)) {
+          return newExpr("UnaryPostfixExpr", [
```
sam-goodwin, line 649: Should we mirror the name? PostfixUnaryExpr

src/compile.ts
```diff
@@ -675,21 +722,25 @@ export function compile(
...
+                  toExpr(varDecl, scope),
+                  toExpr(node.expression, scope),
+                  toExpr(node.statement, scope),
+                ]
+              );
             }
           }
+          throw new SynthError(
+            ErrorCodes.Unsupported_Feature,
+            "For in/of loops with expression initializers are not currently supported."
```
sam-goodwin, line 737: Link to issue?

src/compile.ts
```diff
@@ -675,21 +722,25 @@ export function compile(
...
+          );
+        } else if (ts.isForStatement(node)) {
+          throw new SynthError(
+            ErrorCodes.Unsupported_Feature,
+            "For loops are not currently supported. https://github.com/functionless/functionless/issues/303"
```
sam-goodwin, line 742: I worry about this error message as it could be confused for all types of for loops.

src/declaration.ts
```diff
@@ -83,5 +90,59 @@ export function validateFunctionlessNode<E extends FunctionlessNode>(
...
+  constructor(readonly bindings: (BindingElem | undefined)[]) {
+    super("ArrayBinding");
+    bindings.forEach((b) => b?.setParent(this));
+  }
+
+  public clone(): this {
+    return new ArrayBinding(this.bindings.map((b) => b?.clone())) as this;
+  }
+}
+
```
sam-goodwin, line 146: Docs with example syntax

src/error-code.ts
```diff
@@ -686,6 +686,76 @@ export namespace ErrorCodes {
...
+   *       await func();
+   *    }
+   *    // valid
+   *    return func();
+   * });
+   * ```
+   *
+   * Workaround:
+   *
+   * One workaround would be to invoke a lambda function which handles the non-deterministic parts of the workflow.
```
sam-goodwin, line 720: Another would be to evaluate an express step function

src/error-code.ts
```diff
@@ -686,6 +686,76 @@ export namespace ErrorCodes {
       type: ErrorType.ERROR,
       title: "Unable to find reference out of application function",
     };
+
+  /**
+   * Appsync Integration invocations must be deterministic.
```
sam-goodwin, line 691: I think deterministic is the wrong word. Each of these cases are deterministic based on input. I think you mean they must always be statically defined? Or "resolvers cannot be dynamically called in loops"

I'd consider avoiding this word and instead just exhaustively list what isn't supported.

src/error-code.ts
```diff
@@ -686,6 +686,76 @@ export namespace ErrorCodes {
...
+   *
+   * Generic error for unsupported features.
+   *
+   * See error message provided for more details.
+   */
+  export const Unsupported_Feature: ErrorCode = {
+    code: 10021,
+    type: ErrorType.ERROR,
+    title: "Unsupported feature",
+  };
```
sam-goodwin, line 758: Let's remove this error and instead keep the discipline or defining an error code for each case. The message is not helpful, which will bloat documentation of error codes, and the code that throws and error will still have to provide a description. Let's keep it consistent.

src/node.ts
```diff
@@ -29,7 +37,7 @@ export abstract class BaseNode<
   Kind extends FunctionlessNode["kind"],
   Parent extends FunctionlessNode | undefined = FunctionlessNode | undefined
 > {
-  abstract readonly nodeKind: "Err" | "Expr" | "Stmt" | "Decl";
+  abstract readonly nodeKind: "Err" | "Expr" | "Stmt" | "Decl" | "Node";
```
sam-goodwin, line 40: What is "Node"? That doesn't seem right, we only have errors, expressions, statements and declarations.

src/validate.ts
```diff
@@ -70,6 +70,17 @@ export function validate(
...
+    node: ts.Block,
+    cb: (node: ts.Statement) => T[]
+  ): T[] {
+    const results: T[] = [];
+    ts.forEachChild(node, (child) => {
+      results.push(...cb(child as ts.Statement));
+    });
+    return results;
+  }
+
```
sam-goodwin, line 83: Why make it specific to a block? This could be generic - collectEachChild

src/validate.ts
```diff
@@ -258,6 +286,35 @@ export function validate(
...
+         * expression - await func()
+         * return - return await func()
+         *
+         * Also ensure that we are not inside an inline conditional statement.
+         *
+         * const v = x ? await func() : await func2()
+         */
+        if (
+          rootStatement &&
+          (!checker.isDeterministicStatement(rootStatement) ||
```
sam-goodwin, line 304: Move this guard here. I don't think it belongs in the checker and it's too specific to appsync. Also, probably wouldn't call it deterministic.

src/visit.ts
```diff
@@ -334,13 +345,9 @@ export function visitEachChild<T extends FunctionlessNode>(
     const _else = node._else ? visitor(node._else) : undefined;
 
     ensure(when, isExpr, "a IfStmt's when must be an Expr");
-    ensure(then, isBlockStmt, "a IfStmt's then must be a BlockStmt");
+    ensure(then, isStmt, "a IfStmt's then must be a BlockStmt");
```
sam-goodwin, line 348: Why did the guard change but the error still says it must be a block? What's changed here?

test/test-files/appsync.ts
```diff
@@ -297,3 +297,65 @@ new AppsyncResolver(
     new aws_events.EventBus(stack, "");
   }
 );
+
+/**
+ * Unsupported - non-deterministic invocations
```
sam-goodwin, line 302: Again: the word deterministic is wrong. This is the difference between static/variable length resolvers. Everything here is deterministic

test/vtl.test.ts
```diff
@@ -160,6 +160,19 @@ test("if statement", () => {
...
+    reflect(async ($context) => {
+      if (1 === $context.arguments.num) return;
+      if (1 === $context.arguments.num) {
+      } else return;
+    })
+  );
+
+  // https://github.com/aws-amplify/amplify-category-api/issues/592
+  // testAppsyncVelocity(templates[1]);
+});
```
sam-goodwin, line 174: If I recall, compile.ts rewrites all blockless ifs to use blocks.

sam-goodwin (changes requested):

src/appsync.ts
```diff
@@ -552,8 +566,9 @@ function synthesizeFunctions(api: appsync.GraphqlApi, decl: FunctionDecl) {
             } = ${getResult(stmt.expr)} )\n{}`
           );
         } else {
-          throw new Error(
-            "only a 'VariableDecl', 'Call' or 'Return' expression may call a service"
+          debugger;
```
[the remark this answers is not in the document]
thantos: :-) I'll see an eslint rule to remove them.

src/checker.ts
```diff
@@ -90,15 +91,51 @@ export type FunctionlessChecker = ReturnType<typeof makeFunctionlessChecker>;
...
+    ts.isContinueStatement,
+    ts.isBreakStatement,
+    ts.isReturnStatement,
+    ts.isWithStatement,
+    ts.isSwitchStatement,
+    ts.isLabeledStatement,
+    ts.isThrowStatement,
+    ts.isTryStatement,
+    ts.isDebuggerStatement
+  );
```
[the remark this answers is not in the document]
thantos: This is for `ts.Node`, not `FunctionlessNode` unfortunately.

src/checker.ts
```diff
@@ -90,15 +91,51 @@ export type FunctionlessChecker = ReturnType<typeof makeFunctionlessChecker>;
...
+   *
+   * func();
+   * const v = func();
+   * return func();
+   */
+  const isDeterministicStatement = anyOf(
+    ts.isExpressionStatement,
+    ts.isReturnStatement,
+    ts.isVariableStatement
+  );
```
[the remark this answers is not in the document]
thantos: Guards have checks for `FunctionlessNode`, checker had `is*` statements for `ts.Node`. Did you want to mix them?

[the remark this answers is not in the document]
thantos: (Validator is written with Ts AST and the vtl transpiler is written with Functionless AST.)

src/checker.ts
thantos: There is quite a bit in the checker that does not use the wrapped checker. Did we want to remove them all? 

```ts
  function isStepFunction(node: ts.Node) {
    return (
      isFunctionlessClassOfKind(node, StepFunction.FunctionlessType) ||
      isFunctionlessClassOfKind(node, ExpressStepFunction.FunctionlessType)
    );
  }

  function isApiIntegration(node: ts.Node): node is ts.NewExpression {
    return (
      ts.isNewExpression(node) &&
      isFunctionlessClassOfKind(node.expression, ApiMethod.FunctionlessType)
    );
  }
```

I see some in the checker file, but not in the checker class.

```ts
/**
 * Check if a {@link token} is an {@link ArithmeticToken}: `+`, `-`, `*` or `/`.
 */
export function isArithmeticToken(
  token: ts.SyntaxKind
): token is ArithmeticToken {
  return ArithmeticOperators.includes(token as ArithmeticToken);
}
```

and some in util.ts

```ts
export function hasParent(node: ts.Node, parent: ts.Node): boolean {
  if (!node.parent) {
    return false;
  } else if (node.parent === parent) {
    return true;
  }
  return hasParent(node.parent, parent);
}
```

src/checker.ts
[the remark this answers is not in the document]
sam-goodwin: That tells me we are not classifying some of our nodes properly. They should always fit into the categories we have, no?

src/checker.ts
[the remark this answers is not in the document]
sam-goodwin: Oh I see, you're right lol. I've confused environments - another reason to consolidate ast one day

src/compile.ts
thantos: We can, it would be to just change "name" from a string to an `Identifier`.

src/compile.ts
thantos: Sure

src/error-code.ts
thantos: I was trying to be careful about adding too many details to the workarounds as a sfn and a lambda can do most of the same things (except math, logic, etc).

src/appsync.ts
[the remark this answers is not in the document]
thantos: https://github.com/functionless/functionless/pull/304

src/checker.ts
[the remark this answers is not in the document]
thantos: I am not sure, but I needed this for Validate (use `ts.Node`), not the transpilers (use `FunctionlessNode`).

I think our classifications are right? 

We just don't have

1. debugger
2. empty
3. for
4. switch
5. labelled

test/vtl.test.ts
thantos: It did not (thus the filed issue: https://github.com/functionless/functionless/issues/265) It does rewrite blockless functions to add a block with return. If we want to be 1:1, should we avoid rewriting?

src/validate.ts
thantos: The important part is that the callback acted on statements, not nodes. Updated to use `collectEachChild` internally.

src/visit.ts
thantos: if only we generated those errors instead of hand wrote them, haha...

test/test-files/appsync.ts
thantos: > In [mathematics](https://en.wikipedia.org/wiki/Mathematics), [computer science](https://en.wikipedia.org/wiki/Computer_science) and [physics](https://en.wikipedia.org/wiki/Physics), a deterministic system is a system in which no [randomness](https://en.wikipedia.org/wiki/Randomness) is involved in the development of future states of the system.

Well, not necessarily, I was considering non-deterministic to be a path with uses external inputs (or the result of invocations) where the logic of those invocations are unknown. (and could include "random" behavior).

I can use "unidirectional and defined statically" to match the AWS docs.

src/compile.ts
thantos: https://github.com/functionless/functionless/issues/305

src/checker.ts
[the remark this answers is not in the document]
sam-goodwin: Those are all declarations, statements or expressions. What is a Node?

test/vtl.test.ts
sam-goodwin: Avoiding rewrite would be good

src/compile.ts
thantos: I actually trimmed down the binary constant now and have them all using the same pattern of  using `ts.tokenToString`. I left the BinaryOp's rewriting of `===` and `!==` to `==` and `!=` for now. The individual transpilers need to fail when an op is unsupported instead of us validating typescript.

src/compile.ts
thantos: I called them "condition based for loops" and added `for(;;)`. We should just close #303 so we don't need to differentiate :-P.

src/error-code.ts
thantos: I didn't like creating a detailed error codes for features we need to obviously support. Adding a message with an issue link isn't hard to do inline when throwing. 

Having an error for each missing feature will bloat the error codes more than having a single error code with a replacement error message. 

```ts
ErrorCodes.For_Loops_Are_Unsupported
ErrorCodes.Switch_Statement_Are_Unsupported
ErrorCodes.For_Loop_Expressions_Initialization_Is_Unsupported
```

If we don't have a generic error message, we'll more likely just throw useless errors with no description. 

I'd rather spend the time cutting the ticket and implementing the feature than pruning a bunch of transient error codes. (I'd also like to not throw cryptic, unactionable errors). 

Proposal: 
* If it is a system limitation, a large feature, or a edge case we will likely not support for a while - Add an error code
* If it is a common use case that we have not gotten to implementing - add unsupported feature with a link to a ticket to resolve 
* If it is a case we don't expect to happen, use Unexpected Error with a description of why it shouldn't happen.

src/node.ts
thantos: What is `ObjectBindingPattern` then? In `ts`, it is just a `Node`.

```ts
    export interface ObjectBindingPattern extends Node {
        readonly kind: SyntaxKind.ObjectBindingPattern;
        readonly parent: VariableDeclaration | ParameterDeclaration | BindingElement;
        readonly elements: NodeArray<BindingElement>;
    }
```

`Node` is a utility entity which is part of the tree, but only existings in specific contects.

Object and Array Binding:
* Do not declare (the `BindingElm` does)
* Are not a statement, cannot existing in a block
* Are not an expression, cannot exist in a ExprStmt, etc.
* Not an error, obviously

thantos: When we add `Switch` (#306 ), case and default are other cases.
```ts
    export interface CaseBlock extends Node {
        readonly kind: SyntaxKind.CaseBlock;
        readonly parent: SwitchStatement;
        readonly clauses: NodeArray<CaseOrDefaultClause>;
    }
```

thantos: Catch also fits this description.

```ts
    export interface CatchClause extends Node {
        readonly kind: SyntaxKind.CatchClause;
        readonly parent: TryStatement;
        readonly variableDeclaration?: VariableDeclaration;
        readonly block: Block;
    }
```

sam-goodwin (approved):

PR #304, opened 2022-06-26
chore: no-debugger

thantos:
Debugger statements will trigger for consumers as well. Enforce removing from code before push.


sam-goodwin:
Does the linter stop me using them during testing?

sam-goodwin (approved):

thantos:
> Does the linter stop me using them during testing?

No. If you have the eslint plugin on it'll show an error, but they are not automatically removed or anything.

PR #311, opened 2022-06-27
feat: support higher-order functions and function references

### LINKED ISSUE #148: Native Function Passed Closures
The current implementation of `Function` (#103) assumes that all compilation of the function closure are present in the immediate closure. 

However, closures may be passed around and may themselves contain integrations.

```ts
// 1. closure is passed into the method.
class Functor<T> {
	map<U>(f: ((T) => U)): Function {
		new Function(..., async () => {
			f();
		})
	}
}

// 2. Integration is used inside of a class method
class OurWrapper {
	constructor(private table: Table) {}

	get(key: string) {
		return this.table.getItem({ Key: { S: key } }, ...);
	}
}

const table = new Table();

// 3. Integration is used within a closure method
const deleter = (key: string) => table.deleteItem({ key: { S: key } });

const wrapper = new OurWrapper(table);

new Function(async () => { const item = wrapper.getItem() });

const functor = new Functor();
functor.map(() => table.updateItem(...));
```

Examples 1, 2, and 3 would not work rigth with the current implementation. The integrations within the passed closures would not be discovered by the compiler which relies on parsing the immediate closure to find integrations. 

```ts
// a simple working example
const table = new Table();
new Function(async () => {
	table();
})

// compiled
new Function(new NativeFunctionDecl(
	[], // parameters
	async () => table.native.call(), // the transformed closure
	[table] // the integrations
));
```

```ts
// doesn't work, 
// the table integration isn't found
// the table integration isn't transformed
const deleter = (key: string) => table.deleteItem({ key: { S: key } });
new Function(async () => {
	deleter("some key");
})

// compiled
const deleter = (key: string) => table.deleteItem({ key: { S: key } });
new Function(new NativeFunctionDecl(
	[],
	async () => {
		deleter("some key");
	},
	[]
))
```

#### 1 - attempt to compile all closures
```ts
const deleter = new NativeFunctionDecl(
	[],
	(key: string) => table.native.call.deleteItem({ key: { S: key } }),
	[table]
);

new Function(new NativeFunctionDecl(
	[],
	async () => {
		// somehow transform the deleter function
		deleter("some key");
	},
	[]
));
```

Optimization 1.2 - Only compile closures that contain integtrations, just an optimization. 

Concerns:
1. Would need to complile for each integration
2. More error prone as the compiletions may fail even when not used
3. Slower (?) compilation

#### 2 - Compile on Demand based on usage.

```ts
const deleter = (key: string) => table.deleteItem({ key: { S: key } });

const deleter_native = new NativeFunctionDecl(
	[],
	(key: string) => table.native.call.deleteItem({ key: { S: key } }),
	[table]
);

// when the usage of deleter is found, then compile it.
new Function(new NativeFunctionDecl(
	[],
	async () => {
		// somehow transform the deleter function
		deleter_native("some key");
	},
	[]
));
```

```ts
class Functor<T> {
	map<U>(f: ((T) => U)): Function {
		new Function(..., async () => {
			f();
		})
	}
}

const func = new Functor();
const table = new Table();
const table2 = new Table();

// would need to know that map contains a Function that may need this to be compiled.
func.map(x => table.getItem(...));
func.map(x => table2.getItem(...));
```
Concerns:
1. Closures maybe be variables themselves, how do we trace back to the closures to trasform?

### LINKED ISSUE #309: Support references to functions
We only support converting inline functions to Functionless AST, so a reference to a function won't be usable for DSL generation or IAM Policy inference.

```ts
const foo = () => { .. }
function foo() { .. }

// these are invalid - foo must be inlined
new Function(scope, id, foo);
new StepFunction(scope, id, foo);
```

### LINKED ISSUE #310: Lambda: detect transitive dependencies from higher-order functions
We currently do not support detecting dependencies to infer IAM Policies and automate environment variable/client instantiation if the dependency is contained within another function called by the closure.

- [ ] detect dependencies used by another function (declared outside of its scope)
```ts
const table = new Table(..);

function foo() {
  return $AWS.DynamoDB.GetItem({
    Table: table,
    ...
  });
}

new Function(scope, id, () => {
  return foo();
});
```

- [ ] detect dependencies passed to a function as input
```ts
function foo(table: Table) {
  return $AWS.DynamoDB.GetItem({
    Table: table,
    ...
  });
}

const table = new Table(..);

new Function(scope, id, () => {
  // dependency passed to foo
  return foo(table);
});
```

- [ ] detect dependencies returned from another funciton
```ts
const table = new Table(..);

function getTable() {
  return table;
}

new Function(scope, id, () => {
  return $AWS.DynamoDB.GetItem({
    // dependency returned by a function call
    Table: getTable(),
    ...
  });
});
```

sam-goodwin:
Fixes #309
Fixes #310
Fixes #148


sam-goodwin:
Closing in favor of https://github.com/functionless/functionless/pull/337

PR #312, opened 2022-06-27
fix: lambda invoke from step function, start step function localstack tests

### LINKED ISSUE #307: Step functions lambda integration doesn't result valid result
Testing against localstack, may be their issue too. We'll see.

```
    Expected: "hello world"
    Received: "$.Payload"
```

The result is the contents of the `ResultSelector` field.

According to AWS, the result selector should be a object with fields.

> Use the ResultSelector field to manipulate a state's result before ResultPath is applied. The ResultSelector field lets you create a collection of key value pairs, where the values are static or selected from the state's result. The output of ResultSelector replaces the state's result and is passed to ResultPath.

Our definition:

```ts
exports[` 2`] = `
Object {
  "StartAt": "return await func()",
  "States": Object {
    "return await func()": Object {
      "End": true,
      "Parameters": Object {
        "FunctionName": "\${Token[TOKEN.242]}",
        "Payload": undefined,
      },
      "Resource": "arn:aws:states:::lambda:invoke",
      "ResultPath": "$",
      "ResultSelector": "$.Payload",
      "Type": "Task",
    },
  },
}
`;
```

thantos:
Fixes #307 

* Add some localstack tests for step functions (Standard only, [Express is not supported by localstack](https://github.com/localstack/localstack/issues/5258))
   * No response from localstack in 6mo, unlikely to see this.
* Lambda integration uses the direct resource reference which accepts the payload directly and returns the response
  * Analyze the parameter to determine how to build the task.
* Fix `$AWS.Lambda.Invoke` which was using the wrong parameter name.


sam-goodwin:
> fix: fix

src/aws.ts
```diff
@@ -422,18 +423,18 @@ export namespace $AWS {
...
           throw new Error(
-            "property 'FunctionName' must reference a functionless.Function"
+            "property 'Function' must reference a functionless.Function"
           );
         }
         const functionRef = functionName.ref();
         if (!isFunction(functionRef)) {
           throw new Error(
-            "property 'FunctionName' must reference a functionless.Function"
+            "property 'Function' must reference a functionless.Function"
```
sam-goodwin, line 437: Did we somehow miss these or is this a merge problem?

src/function.ts
```diff
@@ -383,15 +387,30 @@ abstract class FunctionBase<in Payload, Out>
...
+      } else if (isStringLiteralExpr(payloadArg)) {
+        return {
+          Parameters: payloadArg.value,
+        };
+      } else {
+        return {
+          Parameters: ASL.toJson(payloadArg),
+        };
+      }
+    })();
```
sam-goodwin, line 409: Use ternary instead of an instantly called function closure?

test/__snapshots__/step-function.localstack.test.ts.snap
```diff
@@ -0,0 +1,386 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[` 1`] = `
```
sam-goodwin, line 3: Is this missing a description?

test/__snapshots__/step-function.test.ts.snap
```diff
@@ -11,13 +11,9 @@ Object {
         "States": Object {
           "return task(item)": Object {
             "End": true,
-            "Parameters": Object {
-              "FunctionName": "__REPLACED_TOKEN",
-              "Payload.$": "$.item",
-            },
-            "Resource": "arn:aws:states:::lambda:invoke",
```
sam-goodwin, line 18: Ah, was I calling the AWS SDK integration?

test/function.localstack.test.ts
```diff
@@ -131,7 +100,7 @@ localstackTestSuite("functionStack", (testResource, _stack, _app) => {
   // eslint-disable-next-line no-only-tests/no-only-tests
   test.only = _testFunc(testResource.only);
 
-  test(
+  test.only(
```
sam-goodwin, line 103: oops?

src/function.ts
thantos: I can, yeah. This will likely get much more complex and at one point I was trying to generate multiple states. Right now I guess it would be a mostly readable ternary.

test/__snapshots__/step-function.localstack.test.ts.snap
thantos: The self-mutation should remove these. I was comparing the snapshots outside of the tests and jest didn't like that (even though I was giving it a name).

test/__snapshots__/step-function.test.ts.snap
thantos: No, the "optimized", but also returning the whole response wrapper when we didn't need it. The largest issue was the `ResultPath` being a string and not an object. It turned out easier to correctly format the input than to manipulate the response.

test/function.localstack.test.ts
thantos: Why didn't that get removed? strange... thanks for the catch.

src/aws.ts
thantos: Not sure! This is why tests that actually execute the code is important.

thantos: I think it was missed, the last change on that line was 2 months ago.

sam-goodwin (approved):

test/step-function.localstack.test.ts
```diff
@@ -0,0 +1,219 @@
...
+        const [funcRes, outputs] =
+          res instanceof StepFunction ? [res, {}] : [res.sfn, res.outputs];
+        return {
+          outputs: {
+            function: funcRes.resource.stateMachineArn,
+            ...outputs,
+          },
+          extra: {
+            definition: JSON.stringify(funcRes.definition),
+          },
```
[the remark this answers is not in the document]
thantos: The tokens in output are resolved, which create quite a few patterns of random values. tokens are much easier to find and we don't need the resolved tokens in the snapshot.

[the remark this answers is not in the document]
thantos: Function name, arns, etc where creating randomized snapshots.

sam-goodwin (approved):

PR #313, opened 2022-06-27
fix: issue 308 - SFN wait error

### LINKED ISSUE #308: Bug: Step function fails to deploy with `$SFN.wait()`
> Invalid State Machine Definition: 'SCHEMA_VALIDATION_FAILED: Field 'ResultPath' is not supported at /States/$SFN.wait
For(10)

Failing on `ResultPath: null`

thantos:
Fixes #308

### DIFF
--- a/src/asl.ts
+++ b/src/asl.ts
@@ -897,10 +897,31 @@ export class ASL {
         const ref = expr.expr.ref();
         if (isIntegration<Integration>(ref)) {
           const serviceCall = new IntegrationImpl(ref);
-          const taskState = <State>{
-            ...serviceCall.asl(expr, this),
-            ...props,
-          };
+          const partialTask = serviceCall.asl(expr, this);
+
+          const { End, Next } = props;
+
+          const taskState = ((): State => {
+            switch (partialTask.Type) {
+              case "Wait":
+                return {
+                  ...(partialTask as Omit<Wait, "Next">),
+                  ...{ End, Next },
+                };
+              case "Choice":
+              case "Fail":
+              case "Succeed":
+                return partialTask as Choice | Fail | Succeed;
+              case "Task":
+              case "Parallel":
+              case "Pass":
+              case "Map":
+                return {
+                  ...partialTask,
+                  ...props,
+                } as Task | ParallelTask | Pass | MapTask;
+            }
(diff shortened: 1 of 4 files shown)


test/step-function.localstack.test.ts
```diff
@@ -214,4 +214,61 @@ localstackTestSuite("sfnStack", (testResource, _stack, _app) => {
     },
     "hello world"
   );
+
+  test(
+    "call lambda $SFN wait",
+    (parent) => {
+      return new StepFunction(parent, "sfn2", async () => {
+        $SFN.waitFor(1);
```
sam-goodwin, line 222: Should waitFor return a promise?

src/asl.ts
```diff
@@ -897,10 +897,31 @@ export class ASL {
...
+                return partialTask as Choice | Fail | Succeed;
+              case "Task":
+              case "Parallel":
+              case "Pass":
+              case "Map":
+                return {
+                  ...partialTask,
+                  ...props,
+                } as Task | ParallelTask | Pass | MapTask;
+            }
```
sam-goodwin, line 923: style: Please no switch - it's so fugly.

sam-goodwin: I don't understand what this code is doing. What are these cases that need to be handled?

sam-goodwin (changes requested):

src/asl.ts
thantos: Uhgg, but the `if else` was so verbose. Fine.

when `ResultPath` was `null`, on the `Wait` the deployment was failing. So this only injects the allowed properties into the nodes returned by the integrations.

some combination of ResultPath, End, Next based on what the state can handle.

sam-goodwin: Perhaps add some comments because it's not obvious.

thantos: :+1:

test/step-function.localstack.test.ts
thantos: What would it mean to not wait on a blocking statement? I saw it as `Wait` waits synchronously. It also a good way to show integrations can be both async and sync.

sam-goodwin (approved):

PR #315, opened 2022-06-28
feat(step-functions): major refactor - one to many expressions to states - reduce evaluation functions

### LINKED ISSUE #72: Step Functions: support `return typeof id`?
do you support `return typeof id`?

__Originally posted by @thantos in https://github.com/sam-goodwin/functionless/pull/36#discussion_r838919167__

### LINKED ISSUE #79: Step Functions: wait/task token.
Update the Lambda interface to support waitForTask tokens.

### LINKED ISSUE #209: Bug: Step Functions does not support returning template strings
```
`template me ${input.value}`
-----
Error: cannot eval expression kind 'TemplateExpr'
```

```ts
test("`template me ${input.value}`", () => {
  const { stack } = initStepFunctionApp();
  const definition = new ExpressStepFunction<{ value: string }, string>(
    stack,
    "fn",
    (input) => {
      return `template me ${input.value}`;
    }
  ).definition;

});
```

### LINKED ISSUE #279: Bug: Step Functions: Support Object Literals in For Loops
```ts
for (const i in { id: task(input) }) {
```

Throws

```console
expression kind 'ObjectLiteralExpr' cannot be evaluated to a JSON Path expression.
```

### LINKED ISSUE #314: Step Functions: Validate for undefined assignment
Step functions fails when accessing undefined values. Functionless can attempt to be safe when setting undefined values, but there is nothing we can do about variable assignment of undefined values. 

We will add validation for explicit undefined assignment.

```ts
const func = new Function(stack, 'func', () => { return undefined; })
new StepFunction<{ val: string | undefined }, undefined>(stack, 'sfn', async (input) => {
   const v = {
      // invalid - could be undefined
      val: input.val
   }
   // invalid, function outputs undefined.
   const output = await func();
   // invalid - could be undefined
   return input.val;
});
```

* Variable assignment
* Array literal initialization
* Object literal initialization
* Parameters

> Note: It is valid to omit parameters or attributes that are undefined.

thantos:
Closes #209 Closes #314 Closes #279 Resolves #79 Closes #72

Depends on: #325 

* New Features
   * Binary assignment `const b = true && false`
   * Null coalesce `val ?? val2`
   * return template string `return `${anything}`;
   * `return typeof anything` and `const type = typeof anything`
   * variable comparison `var1 == var2`
   * `<` `<=` `>=` `>`
   * integration invocation anywhere `if(await func()) { ... }`
   * `in`
* ASLGraph - Internal API 
   * removed `toJsonPath` and `toJson` in favor of `eval` and the `evalExpr` or `evalContext` wrappers.
      * the goal here was to reduce the number of duplicate transpilation logics acting on the expressions. Each unique logic requires the consumer to know what they want as output for all expressions in the tree. Now eval returns a consistent APi that handles most use cases. The only special use case right now is `toCondition` which may also be able to be squashed to better optimize for conditionals. (later)
   * `eval` now returns one of `AslConstant`, `Variable`, or `CompoundState`
      * `AslConstant` - Number, string, boolean, null, object, array
      * Variable - jsonPath
      * CompoundState - one or more SubStates which return some output (AslConstant or Variable) 
   * `SubStates` (and `CompoundStates`) are flattened into a `States` map (string to `State`). SubState is a sub-graph with a start state. `Next` in sub-states can use local state names, top level state names, relative state names (`../state name`), or `ASL.DeferNext` (or leave Next as undefined).
   * `ASL.DeferNext` or `undefined` - when a sub-state contains states with deferred Next values, the requester is responsible for replacing the Deferred Next with the actual next value. In move cases `joinSubStates` does this by wiring an array of substates to each other. Statements call joinSubStates or `applyDeferNext` to wire the sub-states up to the next statement.
   * Moved the `input` into a special property `$.fnl_context.input`
   * Added a global constant for `null` - Null is difficult to manufacture in sfn. Often null means something different like clear the result output. A constant location to get a null value eliminates that need. 
* Validation, Errors, QOL things
   * Add validation for unsupported binary operations.
   * [x] Validate putEvent object error
   * [x] improve step function validation
   * [x] errors
   * [x] Fix bus.putEvent issue caused by hoist
   * [x] Add more comments and document the new Asl API.


src/asl.ts
```diff
@@ -428,21 +428,52 @@ export class ASL {
...
             ? hoist(updatedChild)
             : updatedChild;
+        } else if (isBinaryExpr(node)) {
+          const updated = visitEachChild(node, (expr) =>
+            normalizeAST(expr, hoist)
+          );
+          return hoist && self.doHoist(node) ? hoist(updated) : updated;
+        } else if (
+          isArrayLiteralExpr(node) &&
+          !node.items.some((i) => isFunctionExpr(i))
```
sam-goodwin, line 438: nit: 

```ts
.some(isFunctionExpr)
```

src/asl.ts
```diff
@@ -428,21 +428,52 @@ export class ASL {
...
+        ) {
+          // if the array contains functions, just leave it, it is probably for Parallel.
+          // This is short sighted, but should work now
```
sam-goodwin, line 441: do we need an issue to track this?

src/asl.ts
```diff
@@ -428,21 +428,52 @@ export class ASL {
...
 
     const start = this.transition(this.decl.body);
     if (start === undefined) {
       throw new Error("State Machine has no States");
     }
 
+    /**
+     * Add some hard to manufacture constants to the machine.
+     *
+     * TODO: only add this when used?
```
sam-goodwin, line 462: I think this is covered by https://github.com/functionless/functionless/pull/316

src/asl.ts
```diff
@@ -497,12 +528,63 @@ export class ASL {
     return this.stateNames.get(stmt)!;
   }
 
-  public execute(stmt: Stmt): States {
+  private rewriteNext(
+    state: State,
+    scope: string,
+    scopeNames: Set<string>
+  ): State {
```
sam-goodwin, line 535: Need some documentation on why we need to re-write. Examples/overall architecture information?

sam-goodwin: This interpreter is starting to become difficult to track. Are we sure we need to re-write states like this versus perform pre-processing on the tree such that it can be mapped directly to ASL?

src/asl.ts
```diff
@@ -497,12 +528,63 @@ export class ASL {
...
+    } else if (!("Next" in state)) {
+      return state;
+    }
+    return {
+      ...state,
+      Next: state.Next ? updateIfInScopeNames(state.Next) : state.Next,
+    };
+  }
+
+  private scopedStateName(scope: string, localName: string) {
```
sam-goodwin, line 558: docs

src/asl.ts
```diff
@@ -497,12 +528,63 @@ export class ASL {
...
+    if (localName === "default") {
+      return scope;
+    }
+    return `${localName}__${scope}`;
+  }
+
+  private normalizeStates(name: string, states: State | States) {
```
sam-goodwin, line 565: docs

src/asl.ts
```diff
@@ -497,12 +528,63 @@ export class ASL {
...
+    if ("Type" in states) {
+      return {
+        [name]: states as State,
+      };
+    } else {
```
sam-goodwin, line 570: Is there an edge case here when the name of a state is `Type`?

src/asl.ts
```diff
@@ -1090,7 +1185,74 @@ export class ASL {
...
+          assignTrue: {
+            Type: "Pass",
+            Result: true,
+            ...props,
+          },
+          assignFalse: {
+            Type: "Pass",
+            Result: false,
+            ...props,
+          },
```
sam-goodwin, line 1204: Do these get re-written for uniqueness?

src/asl.ts
```diff
@@ -1090,7 +1185,74 @@ export class ASL {
...
+          takeLeft: {
+            Type: "Pass",
+            InputPath: left,
+            ...props,
+          },
+          takeRight: {
+            Type: "Pass",
+            ...ASL.toTaskInput(expr.right),
+            ...props,
+          },
```
sam-goodwin, line 1247: If these get re-written, then I totally understand why you added that pass to re-write state names - isolation.

src/asl.ts
```diff
@@ -1090,7 +1185,74 @@ export class ASL {
...
+        };
+      }
+
+      // TODO: assert never
+      throw new SynthError(
+        ErrorCodes.Unsupported_Feature,
+        `Step Function does not support operator ${expr.op}`
+      );
```
sam-goodwin, line 1255: assert never throwing this SynthError would be a nice addition

src/asl.ts
```diff
@@ -1090,7 +1185,74 @@ export class ASL {
...
+            Type: "Pass",
+            ...ASL.toTaskInput(expr.right),
+            ...props,
+          },
+        };
+      }
+
+      // TODO: assert never
+      throw new SynthError(
+        ErrorCodes.Unsupported_Feature,
```
sam-goodwin, line 1253: I thought we removed the `Unsupported_Feature` error code? Can we avoid it and deprecate ASAP. It's too general and encourages laziness.

src/error-code.ts
```diff
@@ -760,6 +760,96 @@ export namespace ErrorCodes {
...
+   *    * ```ts
+   * const func = new Function(stack, 'func', () => { return undefined; })
+   * new StepFunction<{ val: string | undefined }, null>(stack, 'sfn', async (undefined) => {
+   *    const v = {
+   *       // valid - could replace null with any defined value
+   *       val: input.val ?? null
+   *    }
+   *
+   *    // valid - could replace null with any defined value
+   *    const output = (await func()) ?? null;
```
sam-goodwin, line 814: This works? That's cool AF

src/error-code.ts
```diff
@@ -760,6 +760,96 @@ export namespace ErrorCodes {
...
+   *    if(input.val) {
+   *        v = {
+   *           val: input.val
+   *        }
+   *    } else {
+   *        v = {}
+   *    }
+   *
+   *    // valid - could replace null with any defined value
+   *    const output = await func();
```
sam-goodwin, line 836: Didn't we just say that this is invalid?

test/__snapshots__/step-function.localstack.test.ts.snap
```diff
@@ -1,8 +1,1113 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`binaryOps logic 1`] = `
```
sam-goodwin, line 3: I worry about the readability of these state machines

sam-goodwin: Also, cost. Do we have an idea of what % overhead we add in terms of transitions? I see a lot of them, but I assume that's more a measure of the number of branches, not the number of transitions expected in a single execution?

sam-goodwin (changes requested):

test/__snapshots__/step-function.localstack.test.ts.snap
thantos: What's more important, TS coverage syntax or readability and cost? 

Def thinking about cost and readability, but also doing what we need to make things work in native step functions.

src/error-code.ts
thantos: The function signature changed.

src/asl.ts
thantos: I'm changing this. In fact, I am rewriting all of it. Lol. I hope.

src/asl.ts
thantos: Yep

test/__snapshots__/step-function.localstack.test.ts.snap
sam-goodwin: Coverage for sure.

thantos: Yeah. Sfn is hard... I'm refactoring right now in a way that can be a bit smarter. Not leveraging a-normal form when we don't need to, etc. Allows the nodes to optimize themselves.

src/asl.ts
thantos: I had responded to that. I think the alternative is highly granular error codes that we will want to resolve anyways.

src/step-function.ts
```diff
@@ -528,23 +555,40 @@ abstract class BaseStepFunction<
...
     };
+
+    return context.evalContext(call, (evalExpr) => {
+      // evaluate each of the input expressions,
+      // returning an object assignment with the output value { input.$: $.inputLocation }
+      // and a state object containing the output and/or a sub-state with additional required nodes to add to the
+      // machine
+      const evalInputs = Object.entries(inputs)
+        .filter(([, expr]) => !!expr)
+        .map(([key, expr]) => context.toJsonAssignment(key, evalExpr(expr!)));
```
sam-goodwin, line 571: Should this be done as part of the interpreter instead of in the integration?

src/util.ts
```diff
@@ -333,3 +333,18 @@ export class DeterministicNameGenerator {
 
 // to prevent the closure serializer from trying to import all of functionless.
 export const deploymentOnlyModule = true;
+
+export const invertBinaryOperator = (op: BinaryOp): BinaryOp => {
```
sam-goodwin, line 337: What's this for?

test-app/src/message-board.ts
```diff
@@ -1,668 +1,675 @@
...
+//   AppsyncResolver,
+//   Function,
+//   StepFunction,
+//   Table,
+//   EventBus,
+//   Event,
+//   ExpressStepFunction,
+//   AppsyncField,
+//   AppsyncContext,
+// } from "functionless";
```
sam-goodwin, line 24: What happened in this file? Looks like it was nuked?

test/step-function.test.ts
```diff
@@ -80,6 +80,77 @@ test("return PropAccessExpr", () => {
   expect(normalizeDefinition(definition)).toMatchSnapshot();
 });
 
+test("return ElementAccessExpr", () => {
```
sam-goodwin, line 83: How sure are we that these new tests still run properly? Is there value in having snapshot tests without runtime tests?

sam-goodwin (changes requested):

src/asl.ts
```diff
@@ -655,96 +740,119 @@ export class ASL {
...
+          Object.assign(states, this.evalStmt(curr));
+        }
+
+        const next =
+          curr === undefined
+            ? this.next(stmt)
+            : // there was an else
+              this.getStateName(curr);
+
+        return ASLGraph.applyDeferNext(
```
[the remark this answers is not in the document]
thantos: Changed to `updateDeferredNextStates`. It updates all of the states with `ASLGraph.DeferNext` as the `Next` value to whatatever the given `Next` or `End` values are.

src/asl.ts
```diff
@@ -655,96 +740,119 @@ export class ASL {
...
+          { Next: next },
+          {
+            startState: "choose",
+            states: {
```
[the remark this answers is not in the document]
thantos: I actually like of like that the ASL keys are capitalized and that our special fields are lower cased. At the end of transpilation, only capitalized fields should be left.

src/asl.ts
```diff
@@ -767,41 +875,56 @@ export class ASL {
...
-      if (throwTransition === undefined) {
-        return {
-          [this.getStateName(stmt)]: {
+        const throwTransition = this.throw(stmt);
+        if (throwTransition === undefined) {
+          if (!isIdentifier(updated.expr)) {
+            throw new SynthError(
+              ErrorCodes.StepFunctions_error_name_and_cause_must_be_constant
+            );
+          }
```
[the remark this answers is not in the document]
thantos: Hmm, well we were taking the name, but I realized now that would not allow for property access. I changed it back to use `exprToString`. which could result in some ugly names, but will be unlikely.

src/asl.ts
```diff
@@ -872,55 +995,214 @@ export class ASL {
...
     return assertNever(stmt);
   }
 
+  /**
+   * Recursively evaluate a single expression, building a single {@link ASLGraph.NodeResults} object.
+   * All SubStates generated during evaluation will be merged into into a {@link ASLGraph.OutputSubState} along with the output
+   * of the handler callback.
+   *
+   * @param expr - Expression to evaluate.
+   * @param contextNode - Optional node to associate with the output state. This node may be ued to name the resulting state.
```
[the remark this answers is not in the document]
thantos: image
I am, lol

src/aws.ts
```diff
@@ -440,15 +440,17 @@ export namespace $AWS {
...
-          },
-        };
+
+        return context.evalExpr(payload, (output) => {
+          return context.outputState({
+            Type: "Task",
+            Resource: "arn:aws:states:::lambda:invoke",
+            Parameters: {
+              FunctionName: functionRef.resource.functionName,
+              ...context.toJsonAssignment("Payload", output),
```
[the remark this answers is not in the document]
thantos: It optionally adds the `.$` based on the output being a `JsonPath` or a `Value` (I added this method a few months ago, just changed it to support `ASLGraph.Output` instead of Expr.

src/asl.ts
```diff
@@ -872,55 +995,214 @@ export class ASL {
...
+    ? ASLGraph.OutputSubState
+    : ASLGraph.NodeResults {
+    const [node, handler] = isNode(nodeOrHandler)
+      ? [nodeOrHandler, maybeHandler!]
+      : [expr, nodeOrHandler];
+
+    const state = this.eval(expr);
+    const output = ASLGraph.getAslStateOutput(state);
+
+    const exprState = handler(output);
```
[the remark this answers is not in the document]
thantos: It accepts an output and returns a state/substate. It was an attempt to reduce the leakiness as the consumer doesn't not need to handle an of the states generated by child expressions. This method merged the output of `handle` with the states generated by `eval`.

In most cases, the generated child states are not needed by the consumer, they just need to be joined/returned.

src/aws.ts
```diff
@@ -576,15 +578,21 @@ function makeDynamoIntegration<
...
+
+      return context.evalExpr(renamedExpr, (output) => {
+        return context.outputState(
+          ASLGraph.applyConstantOrVariableToTask(
+            {
+              Type: "Task",
+              Resource: `arn:aws:states:::aws-sdk:dynamodb:${operationName}`,
+            },
+            output
+          )
```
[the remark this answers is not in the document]
thantos: It was simpler before! Before `eval` couldn't return multiple states, json path, and or constant values. Which meant that some values could not be passed to this integration, etc.

The callbacks (there is only one callback here) are an attempt at reducing the leakiness of the API (see [comment above)](https://github.com/functionless/functionless/pull/315#discussion_r917298393)).

If we wanted to break this out, it could be written as...

```ts
const exprState = context.eval(renamedExpr);
const output = ASLGraph.getAslStateOutput(exprState);

const tempHeap = context.newHeapVariable();

const task = {
    Type: "Task",
    Resource: `arn:aws:states:::aws-sdk:dynamodb:${operationName}`,
    ResultPath: tempHeap,
    ...ASLGraph.isJsonPath(output) ? { InputPath: output.jsonPath } : { Parameters: output.value }
};

const joined = ASLGraph.joinSubStates(expr, exprState, state);

return {
   ...joined,
   output: {
       jsonPath: tempHeap
   }
};
```

Updated the `apply*` names to `ASLGraph.taskWithInput` and `ASLGraph.passWithInput`.

src/error-code.ts
```diff
@@ -760,6 +760,295 @@ export namespace ErrorCodes {
...
+   *   });
+   * });
+   *
+   * const sfn = new StepFunction(stack, "sfn", async () => {
+   *   const event = { source: "lambda", "detail-type": "type", detail: {} };
+   *   await sender(event);
+   * });
+   * ```
+   *
+   * The limitation is due to Step Function's lack of optional or default value retrieval for fields. Attempting to access a missing field in ASL leads to en error. This can be fixed using Choice/Conditions to check for the existence of a single field, but would take all permutations of all optional fields to support optional field at runtime. Due to this limitation, we currently compute the transformation at compile time using the fields present on the literal object. For more details and process see: https://github.com/functionless/functionless/issues/101.
```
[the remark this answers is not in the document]
thantos: Ahh, I was thinking that the breaks would appear in the html, but I see that single line breaks are ignored.

ex:

```ts
  /*
   * Functionless processes your application code to find infrastructure references, wire up permissions, build clients, and more.
   * All infrastructure must be created outside of your application logic closures.
   * References to those resources (Integrations) must be traceable back to their location outside of your application functions.
   */
```

image


Neat.

src/error-code.ts
```diff
@@ -760,6 +760,295 @@ export namespace ErrorCodes {
...
+   * ```ts
+   * new StepFunction<{ value: undefined }, { error?: string }>(this, 'sfn', async (input) => {
+   *    // invalid - the error cause is not constant
+   *    return {
+   *       error: input.value
+   *    }
+   * });
+   * ```
+   *
+   * TODO: add validation
```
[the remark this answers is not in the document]
thantos: :+1: Missed searching for TODO in this file, haha...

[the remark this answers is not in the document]
thantos: Also part of the reason for https://github.com/functionless/functionless/pull/325

src/error-code.ts
```diff
@@ -760,6 +760,295 @@ export namespace ErrorCodes {
...
+   *
+   *    const arr = [1];
+   *    // valid
+   *    arr[0]
+   *    // invalid
+   *    const zero = 0;
+   *    arr[zero];
+   * });
+   *
+   * Workaround - use lambda
```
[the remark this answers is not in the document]
thantos: Updated a few other places too

src/error-code.ts
```diff
@@ -760,6 +760,295 @@ export namespace ErrorCodes {
...
+   */
+  export const StepFunctions_error_name_and_cause_must_be_constant: ErrorCode =
+    {
+      code: 10024,
+      type: ErrorType.ERROR,
+      title: "StepFunctions error name and cause must be constant",
+    };
+
+  /**
+   * Error is thrown when an array or object is incorrectly accessed.
```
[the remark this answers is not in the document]
thantos: I was intending this error to represent any invalid collection access in our transpilers with the details for each resource. I think we'll regret having granular error message for each error type for each resource.

src/error-code.ts
```diff
@@ -760,6 +760,295 @@ export namespace ErrorCodes {
...
+   * new StepFunction(stack, 'sfn', () => {
+   *    // valid
+   *    await accessor(obj, a);
+   * });
+   * ```
+   */
+  export const Invalid_collection_access: ErrorCode = {
+    code: 10025,
+    type: ErrorType.ERROR,
+    title: "Invalid collect access",
```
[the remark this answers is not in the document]
thantos: Do we want granular error codes for every resource? Will we have a separate code for invalid collection access in appsync or api gateway?

src/error-code.ts
```diff
@@ -760,6 +760,295 @@ export namespace ErrorCodes {
...
+   *
+   * Generic error code for when the user provided an unexpected input as documented and reflected in the types.
+   *
+   * See the error message for more details.
+   */
+  export const Invalid_Input: ErrorCode = {
+    code: 10027,
+    type: ErrorType.ERROR,
+    title: "Invalid Input",
+  };
```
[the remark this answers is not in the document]
thantos: So, there are three ways an error can happen

1. Unexpected - 500 type error, we didn't think the error could happen
2. Unsupported - user would think it should work because it is valid ts, but our transpiler or the service doesn't support it. This is a majority of our granular errorrs.
3. Invalid Input - 400 type errors, we told the user what to send via the types, but somehow they send something else.

What should we throw in this case?

```ts
      return context.evalExpr(seconds, call, (secondsOutput) => {
        if (
          ASLGraph.isValue(secondsOutput) &&
          typeof secondsOutput.value === "number"
        ) {
          return context.voidState({
            Type: "Wait" as const,
            Seconds: secondsOutput.value,
          });
        } else if (ASLGraph.isJsonPath(secondsOutput)) {
          return context.voidState({
            Type: "Wait" as const,
            SecondsPath: secondsOutput.jsonPath,
          });
        }

        throw new SynthError(
          ???,
          "Expected the first parameter (seconds) to $SFN.waitFor to a number or a variable."
        );
      });
```

We could:
1. merge 1 and 3, all errors whether caused by the user or the library throw `Unexpected`
2. Create granular errors for any 400 type case, even when typescript should catch it. (unless ts-ignore or similar could be used.)
    3.  ex 1: "$SFN.WaitFor expected a variable or a string"
    4. ex 2: "$SFN.WaitUntil expected a variable or a number"
3. Less granular error for any 400 type cases
    3.  ex 1: "Invalid input to a step function integration"
    4. ex 2: "Invalid input to an app sync validation"
6. Create a new, generic error code for 400 type errors.
   7. Invalid Input - Message: "$SFN.WaitFor expected a variable or a string"
   8. Invalid Input - Message: "$SFN.WaitUntil expected a variable or a number"
10. Just throw ugly errors, this is ts's job, we don't think it will ever happen.

I like option 6 personally, granular will be a PITA and I like to separate 500 from 400 type errors as the messaging is much different.

src/step-function.ts
thantos: You mean like wrap each of the states in new types and do some of the leaky transformation later?

```ts
const state = ASLGraph.Pass = {
   Type: "Task",
   parameters: {
       StateMachineArn: this.resource.stateMachineArn,
       Input: input ? evalExpr(input) : undefined,
       Name: name ? evalExpr(name) : undefined,
       TraceHeader: traceHeader ? evalExpr(traceHeader) : undefined,
   }
}
```

later...


```ts
const task = ASLGraph.graphTaskToTask(graphTask);
```

```ts
function graphTaskToTask(task: ASLGraph.Task): Task {
   return {
      Type: "Task",
      ...,
      Parameters: // ... determine if the values need `.$` or not
   }
}
```

src/util.ts
thantos:
```ts
/**
 * Inverts directional binary operators.
 * 
 * Useful when normalizing the processing of the left and right of a binary operator.
 *
 * < -> >
 * > -> <
 * >= -> <=
 * <= -> >=
 */
```

test-app/src/message-board.ts
thantos: woops.

test/step-function.test.ts
thantos: Generally I have been adding more verbose and granular tests here and adding similar, but less granular cases to the runtime tests.

src/asl.ts
thantos: No longer an issue. Sub-states now contains a property called  `states` with the dynamic keys.

src/asl.ts
thantos: Coming in https://github.com/functionless/functionless/pull/323

src/step-function.ts
```diff
@@ -257,30 +276,38 @@ export namespace $SFN {
...
-            ? "$$.Map.Item.Index"
-            : arrayPath,
-        ])
-      ),
-    };
+
+    return context.evalExpr(array, call, (arrayOutput) => {
+      if (!ASLGraph.isJsonPath(arrayOutput)) {
+        // TODO
+        throw new Error();
```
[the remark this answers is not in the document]
thantos: Managed to remove the error and support constant arrays in `$SFN.Map`!

```ts
      return new StepFunction(parent, "sfn2", async () => {
        return $SFN.map([1, 2, 3], (n) => {
          return `n${n}`;
        });
      });
```

src/error-code.ts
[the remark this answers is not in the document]
sam-goodwin: My main comment was that the `title: "Invalid collect access"` does not match the `const Invalid_collection_access`

PR #316, opened 2022-06-28
feat(step-functions): move input to dynamic context

thantos:
* Move the input to a context variable in the machine.
* Prepares step functions for #68

### DIFF
--- a/src/asl.ts
+++ b/src/asl.ts
@@ -361,7 +361,7 @@ export class ASL {
    */
   static readonly ContextName = "Amazon States Language";
   /**
-   * Tag this instance with its Functionless Context ({@link ASL.ContextName})
+   * Tag this instance with its Functionless Context ({@link this.ContextName})
    */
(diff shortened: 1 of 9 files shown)


src/asl.ts
```diff
@@ -381,6 +381,8 @@ export class ASL {
   private readonly stateNamesCount = new Map<string, number>();
   private readonly generatedNames = new DeterministicNameGenerator();
 
+  private needsFunctionlessContext: boolean = false;
```
sam-goodwin, line 384: comment to add some context on what this is used for? Define functionless context - I assume this is for lexical scoping and constants?

sam-goodwin (approved):
How confident are we that this won't break runtime? Do we have runtime tests yet?

thantos:
> How confident are we that this won't break runtime? Do we have runtime tests yet?

We do! At least a few. It shouldn't break runtime. (step-functions.localstack.test.ts)

PR #325, opened 2022-07-09
fix: test for validation and testing of all error codes

thantos:
* Adds automated tests which tests for the presence of error codes in the validate output.
* Adds some missing error code validation
  * Expected object input
  * Unable to find out of scope reference
* Fixed validation error for event bus transform (was returning the rule error)

When new error codes are introduced, the PR should include one of
1. New Validation
   1. And add at least one test case in the `test/test-files` that can cause the error
3. Add the error code to the `skipErrorCodes` in `validate.test.ts` with an explanation as to why the code cannot be validated for.

image


sam-goodwin:
> When new error codes are introduced, the PR should include one of

Add this to contributing guide?

sam-goodwin (approved):

CONTRIBUTING.md
```diff
@@ -180,6 +180,18 @@ export const My_New_Error: ErrorCode {
 
 **TSDocs** - Above the ErrorCode, TS docs should be used to describe the error in detail. This can contain markdown, which will be turned into markdown in the docs. Include code examples of what not to do is possible, go into detail on the issue, and explain any workarounds or escape hatching available to bypass the issue.
 
+2. [Optional] - Add validation
+
+Any ErrorCode which can be determined statically should have static validation added to `validate.ts`. `validate.ts` uses the Typescript AST to find errors which would be thrown later. It feels both the `functionless` api (`npx functionless`) and the [`@functionless/language-service`](https://github.com/functionless/functionless-language-service).
```
sam-goodwin, line 185: > it feels

?

src/aws.ts
```diff
@@ -400,7 +400,7 @@ export namespace $AWS {
      * @see https://docs.aws.amazon.com/lambda/latest/dg/API_Invoke.html
      */
     export const Invoke = makeIntegration<
-      "Lambda.Invoke",
+      "$AWS.Lambda.Invoke",
```
sam-goodwin, line 403: This seems like an unrelated change?

src/validate.ts
```diff
@@ -224,127 +271,116 @@ export function validate(
...
+            ts.isIfStatement,
+            ts.isDoStatement,
+            ts.isWhileStatement,
+            ts.isForStatement,
+            ts.isForInStatement,
+            ts.isForOfStatement,
+            ts.isSwitchStatement,
+            ts.isLabeledStatement,
+            ts.isTryStatement
+          );
```
sam-goodwin, line 334: Declare this at the module scope?

src/validate.ts
```diff
@@ -224,127 +271,116 @@ export function validate(
...
+           * Also ensure that we are not inside an inline conditional statement.
+           *
+           * const v = x ? await func() : await func2()
+           */
+          if (
+            rootStatement &&
+            (isControlFlowStatement(rootStatement) ||
+              findParent(
+                node.expression,
+                ts.isConditionalExpression,
```
sam-goodwin, line 353: Should this be `isControlFlowStatement`?

sam-goodwin: I.e., detect any integration within a control flow statement?

sam-goodwin (changes requested):

src/aws.ts
thantos: Just making them consistent. The values were not used before.

src/validate.ts
thantos: just copied from compile, but yeah, will do.

src/validate.ts
thantos: not sure, just copied your code from compile so I could use it in validate, didn't read it :-) will look.

src/aws.ts
thantos: Wanted them to match to make this part cleaner

image

src/validate.ts
thantos: ohh, sorry, this was my code. 

Not sure what you are pointing out. It fails on control flow (like `if`/`for`) or in ternary (`conditionalexpr`). This is unchanged, just moved.

sam-goodwin: Is rootStatement guaranteed to be at the top level of the function? Or could it be inside another statement?

PR #326, opened 2022-07-11
feat(step-functions): for in/of loop return, break, continue, assign

### LINKED ISSUE #319: Step Functions: Return from a for loop
```ts
        const b = ["b"].map((v) => {
          for (const i of [1, 2, 3]) {
            if (i === 3) {
              return `${v}${a}${i}`;
            }
          }
          return "boo";
        });
```

thantos:
Closes #319 

Depends on: #329

* [x] return
* [x] continue
* [x] break
* [x] assign to upper scopes
* [x] Tests


thantos:
Redesigning how or loops work

Current: Map state iterators over each item in an array. The map context is used to assign item or item index.

* ForInExpr: Needs index (`i`) and item (`arr[i]`) - `for(const i in arr)` 
* ForOfExpr: Needs item (`item`) - `for(const item in arr)`

* continue: continue to the next state, skipping all future lines in the block
* break: short circuit - continue to the line after the current for loop scope, skipping all future lines in the block
* return: short circuit - return from the current functional scope, breaking out of all nested scopes that are not function scopes. Return any data up/out.

Problems with map (Current):
* Cannot both short circuit (break/return) and return data (return)
* Cannot manipulate state outside of the map
  * Could be solved by some special return envelope
* State mutation cannot be shared between map states

# Alternatives: 

## tail recursion on array, slicing off of the array using json path

Problems:
* How do we get the index (`i`) for `ForInExpr`?

Solution:
* Use Map to compute an array of indices before `ForInExpr`, iterator over the new array. 

### ForOf

image

1. Assign array to temp variable `$temp_arr = $arr`
2. Check for empty array `IsPresent($.temp_arr[0])`
   3. When true, continue
   4. When false, go to 6
3. Assign the item variable `$item = $temp_arr[0]`
4. Do ForOf body...
    1. Continue - jump to 5
    2. Break - jump to 6
    3. Return - create return state for the current function scope
5. Slice temp_arr `$temp_arr = $temp_arr[1:]`, go to 2
6. continue to post for loop - Next: DeferNext
    1. Clean up scope variables?

### ForIn

image

1. Use `Map` state to create an array that pairs index to item and assign to $temp_arr (`.map((item, index) => ({ item, index}));`)
2. Check for empty array `IsPresent($.temp_arr[0])`
   3. When true, continue
   4. When false, go to 6
3. Assign the item variable `$0_i = $temp_arr[0].item` and `$i =$temp_arr[0].index`
4. Do ForOf body...
    1. Continue - jump to 5
    2. Break - jump to 6
    3. Return - create return state for the current function scope
5. Slice temp_arr `$temp_arr = $temp_arr[1:]`, go to 2
6. continue to post for loop - Next: DeferNext
    1. Clean up scope variables?

sam-goodwin:
make sure to use feature scopes, e.g. `feat(step-functions):`

test/step-function.localstack.test.ts
```diff
@@ -1105,4 +1167,37 @@ localstackTestSuite("sfnStack", (testResource, _stack, _app) => {
...
+  test(
+    "for control and assignment",
+    (parent) => {
+      return new StepFunction<{ arr: number[] }, string>(
+        parent,
+        "fn",
+        async (input) => {
+          let a = "";
+          for (const i in input.arr) {
+            if (i === "2") {
```
sam-goodwin, line 1180: comparing with string?

sam-goodwin (approved):

test/step-function.localstack.test.ts
thantos: `i` is the index which is a string. see the for of below comparing with a number.

PR #323, opened 2022-07-07
Sfn binary

thantos:
Closes #232 

Depends on #315 

* Binary Expressions - mostly errors 
   * [x] `+` | `-` | `/` | `*` | `+=` | `-=` | `*=` | `/=`
* Unary Expressions
  * [ ] `!` expand support
  * [x] Errors - `-` | `++` | `--`
* [ ] Validation/Language Service

PR #329, opened 2022-07-13
feat(step-functions): refactor evalStmt to return a sub-state, remove node state cache

thantos:
Depends on: #323

Before, `evalStmt` returned a map of state name to state. This update uses the `ASLGraph` to instead support:

* `NodeState`  - single state
* `SubState` - a local sub-graph of states
* undefined - no states

This allowed us to remove most of the node based prediction logic. 

* State names are purely cosmetic, the node-state name cache has been removed (unique state name cache remains)
* States are wired together by use of local or relative state names that are resolved to unique nodes at resolution time
* We search the graphs for orphaned states and remove them to prevent step function failures (DFS to find reachable states)
* We search the graphs for empty states that can be collapsed 

```ts
{
   1: { ..., Next: 3 },
   2: { ..., Next: 3 },
   3: { Type: "Pass" , Next: 4 },
   4: { ..., ??? }
}

=>

{
   1: { ..., Next: 4 },
   2: { ..., Next: 4 },
   4: { ..., ??? }
}
```


src/asl.ts
```diff
@@ -390,15 +417,28 @@ export class ASL {
...
+   * A pointer to the nearest break point.
+   *
+   * For and While loops should implement this state.
+   */
+  private static BreakNext: string = "__BreakNext";
+
+  /**
+   * A pointer to the nearest catch state.
+   */
+  private static CatchState: string = "__catch";
```
sam-goodwin, line 440: Should these be `readonly`?

src/asl.ts
```diff
@@ -479,21 +514,15 @@ export class ASL {
         },
         ResultPath: FUNCTIONLESS_CONTEXT_JSON_PATH,
         OutputPath: "$",
-        Next: start,
+        Next: ASLGraph.DeferNext,
```
sam-goodwin, line 517: Could we use `undefined` as `ASLGraph.DeferNext` if `End: undefined`?

src/asl.ts
```diff
@@ -716,124 +706,123 @@ export class ASL {
...
-          choices.push({
-            Next: this.getStateName(curr.then),
-            ...cond,
-          });
-          curr = curr._else;
-        }
-        if (isBlockStmt(curr)) {
-          Object.assign(states, this.evalStmt(curr));
+        if (!body) {
+          // TODO: test me
```
sam-goodwin, line 718: ?

sam-goodwin (approved):
beautiful

src/asl.ts
thantos: the logic is (`Next` or `Default` for `Choice`)
* implicit - `Next === undefined && End === undefined`
* explicit - `Next = ASLGraph.DeferNext`

I've been trying to use the explicit case where I can, but some of the states just have `Next: undefined`.

PR #331, opened 2022-07-14
feat(step-functions): ternary, join, bug fixes

### LINKED ISSUE #73: Step Functions: gracefully handle empty blocks
The Step Function interpreter does not handle all cases when a `BlockStmt` is empty - it does handle cases like an empty catch (i think), but we need to enumerate the cases thoroughly.

### LINKED ISSUE #78: Step Functions: `undefined` vs `null`
In Step Functions, we currently use `null` in places where we really mean `undefined` in TypeScript. This works for many cases but is definitely a gotcha that could hurt people. How do we model `undefined` in Step Functions?

An object with a missing is "undefined", but how do we return `undefined` as the whole result? We can't - only `null` can be returned.

### LINKED ISSUE #84: Step Functions: chained map, filter, etc.
Optimize chained map/filter calls into single Map tasks with each of the steps inlined to avoid multiple traversals.

```ts
list.map(A).filter(B)
```

Becomes one Map Task
```
Type: Map,
Iterator: {
  A,
  B
}
```

### LINKED ISSUE #210: Step Functions: filters with sub-tasks
```
await Promise.all(input.list.map((item) => task(item)))).filter
-----
Error: .filter with sub-tasks are not yet supported
```

```ts
test("await Promise.all(input.list.map((item) => task(item)))).filter", () => {
  const { stack, task } = initStepFunctionApp();
  const definition = new ExpressStepFunction<
    { list: string[] },
    (number | null)[]
  >(stack, "fn", async (input) => {
    return (await Promise.all(input.list.map((item) => task(item)))).filter(
      (item) => item !== null
    );
  }).definition;

});
```

### LINKED ISSUE #281: Step Function: Support tunary/conditionalexpr

thantos:
Closes #281 Closes #210 Closes #78 Closes #73
Partial #84

* `join`
* ternary
* bug: out of scope in map states
* removed most of the `skip`s in step function tests.

### DIFF
--- a/src/asl.ts
+++ b/src/asl.ts
@@ -1482,6 +1482,8 @@ export class ASL {
             );
           }
         }
+      } else if (isJoin(expr)) {
+        return this.joinToStateOutput(expr);
       } else if (isPromiseAll(expr)) {
         const values = expr.getArgument("values");
         // just validate Promise.all and continue, will validate the PromiseArray later.
@@ -1491,7 +1493,9 @@ export class ASL {
         throw new SynthError(ErrorCodes.Unsupported_Use_of_Promises);
       }
       throw new Error(
-        `call must be a service call or list .slice, .map, .forEach or .filter, ${expr}`
+        `call must be a service call or list .slice, .map, .forEach or .filter: ${exprToString(
+          expr
+        )}`
       );
     } else if (isVariableReference(expr)) {
       if (isIdentifier(expr)) {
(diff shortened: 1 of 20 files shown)


src/asl.ts
```diff
@@ -1255,7 +1255,9 @@ export class ASL {
 
     if (isPromiseExpr(expr)) {
       // if we find a promise, ensure it is wrapped in Await or returned then unwrap it
-      if (isAwaitExpr(expr.parent) || isReturnStmt(expr.parent)) {
+      if (
+        expr.hasOnlyAncestors(isConditionExpr, anyOf(isReturnStmt, isAwaitExpr))
```
sam-goodwin, line 1259: Does it stop at the first one it finds?

thantos: The first predicate is a passthrough and the second it stops at

src/asl.ts
```diff
@@ -2099,6 +2164,137 @@ export class ASL {
     }
   }
 
+  private joinToStateOutput(
```
sam-goodwin, line 2167: Big function. Could use some documentation

sam-goodwin (approved):

src/asl.ts
sam-goodwin: I mean at the top. A summary of what syntax it enables. The docs inside are great

Issue #305, opened 2022-06-26
Support expression initializers in `for in` and `for of` loops.


thantos:
```ts
let x;
for(x of arr) {
  ...
}
console.log(x);
```

Issue #303, opened 2022-06-26
For loops


thantos:
Compile doesn't support for loops. Support in applicable runtime functions.

* [ ] Step Functions
* [ ] AppSync (no integrations)
* [ ] Api Gateway (no no integrations)
* [ ] Validate
  * [ ] Event Bus
  * [ ] App sync - no integrations
  * [ ] api gateway - no integrations

sam-goodwin:
To clarify: for-in and for-of are supported

sam-goodwin:
No idea how step functions would support this without polyfilling with lambda

thantos:
Well you can use a (conditional) for loop, but the conditional clause cannot use arithmetic.

PR #332, opened 2022-07-17
feat(step-functions): for-loops and for-in/of ident initializer

thantos:
Partial #305 Partial #303 

```ts
for(const i = arr[0]; arr ; arr = arr.slice(1)) {
    await func(i);
}
```

```ts
let i = "";
for(i in arr) {
   await func(i);
}
await func(i);
```


src/asl.ts
```diff
@@ -723,6 +724,65 @@ export class ASL {
...
+          Type: "Pass",
+        };
+      }
+
+      return this.evalContextToSubState(stmt, (evalExpr) => {
+        const initializer = stmt.variableDecl
+          ? isVariableStmt(stmt.variableDecl)
+            ? this.evalStmt(stmt.variableDecl)
+            : evalExpr(stmt.variableDecl)
+          : undefined;
```
sam-goodwin, line 743: What does it mean for `variableDecl` to not be a `VariableStmt`?

src/visit.ts
```diff
@@ -285,27 +287,38 @@ export function visitEachChild<T extends FunctionlessNode>(
...
-    }
+    const body = blockStmtVisitor(node.body, visitor);
 
     return (
       isForInStmt(node)
         ? new ForInStmt(variableDecl, expr, body)
         : new ForOfStmt(variableDecl, expr, body)
     ) as T;
+  } else if (isForStmt(node)) {
+    const body = blockStmtVisitor(node.body, visitor);
```
sam-goodwin, line 305: Why not just recurse instead of separating out logic into `blockStmtVisitor`?

src/asl.ts
```diff
@@ -723,6 +724,65 @@ export class ASL {
...
+    } else if (isForStmt(stmt)) {
+      const body = this.evalStmt(stmt.body);
+
+      if (!body) {
+        // if the body is empty, the for expression is evaluated,
+        // and then continue on. This state will be optimized out later.
+        return {
+          Type: "Pass",
+        };
+      }
```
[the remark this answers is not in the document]
thantos: yeah? ha... I don't think we want to be responsible for non-terminating machine detection. If the user wrote infinitely executing typescript code, the machine will do the same.

src/asl.ts
thantos:
```ts
let i;
for(i in [1,23]) {}
return i;
```

thantos: In the TS AST, the initializer is a `VariableDeclList` or `Expr`, but in the case of for in and for of, it can only be (I think?) `[VariableDecl]` or `Identifier`.

I don't like that they don't restrain the types.

src/asl.ts
```diff
@@ -723,6 +724,65 @@ export class ASL {
...
+                  { Type: "Pass" as const },
+            // then run the body
+            body: ASLGraph.updateDeferredNextStates(
+              { Next: "increment" },
+              body
+            ),
+            // then increment (or do nothing)
+            increment: incrementState,
+            // return back to check
+            // TODO: clean up?
```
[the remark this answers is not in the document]
thantos: yeah, there is a general task to correctly reflect scope and garbage collection.

src/visit.ts
thantos: The visitor doesn't recurse, it handles each expr's children by passing it to the visitor and then validates the response. In the case of the block, you had allowed the normalization of a single or multiple exprs to block or a block itself being returned.

```ts
  let block = visitor(node);
  if (Array.isArray(block)) {
    block = flatten(block);
    ensureItemOf(block, isStmt, "Statements in BlockStmt must be Stmt nodes");
    return new BlockStmt(block);
  }
  ensure(block, isBlockStmt, `Body in ${node.kind} must be a BlockStmt`);
  return block;
```

If we recursed, the BlockStmt itself would never be passed to the visitor as the BlockStmt handler only passes the children down to the visitor, but the `For*` pass the BlockStmt itself down and accept multiple nodes as a result.

thantos: Updated all cases of block handling to support returning an array of stmt or a blockstmt now. (it was inconsistent)

src/asl.ts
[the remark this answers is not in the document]
thantos: You are right @sam-goodwin, a for loop with an empty body still executes the init, condition, and incrementor. Will ensure it does that.

src/asl.ts
```diff
@@ -688,15 +692,40 @@ export class ASL {
...
+                      ResultPath: `$.${stmt.variableDecl.name}`,
+                      Next: "assignValue",
+                    },
+                    assignValue: {
+                      Type: "Pass",
+                      InputPath: `${tempArrayPath}[0].item`,
+                      ResultPath: `$.0__${stmt.variableDecl.name}`,
+                      Next: "body",
+                    },
+                  },
```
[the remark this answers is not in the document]
thantos: So, I had it as one step, it assigned `$.[var]` to the `{index, item}` object and then all reference were re-written as `$.[var].item` or `$.[var].index`.

It is not possible to set two top level variables in one state in ASL. you can overwrite the entire top level (`ResultPath: "$"`) or set a single property in the top level (`ResultPath: "$.[var]"`).

This didn't work when the `var` was an identifier with a lifetime beyond the for loop though. 

```ts
let i = "";
i = `${i}something`;
for(i in [1,2,3]) {
   ...
}
return i;
```

Because now the for loop would overwrite the `i` variable with an object instead of the index string.

One option would have been to "fix" the variable at the end of for loop (`$.[var] = $.[var].index`), but I decided to keep the variable clean (`$.[var]` always is the index of the for loop iteration, or the value in the case of a `ForOfStmt`.)

Options:
1. Current State - `$.[var]` (index) and `$.0__[var]` (item) - set in 2 states on each iteration.
2. Object and Fix - `$.[var] = { index, item }` for each iteration, after the for loop `$.[var] = $.[var].index`. - Variable may be shadowed internally, want to keep the model simple, prefer not to "fix" after.
3. Do not support using an identifier in `ForIn` - Prefer to support all TS features, no matter the cost.
4. Different behavior for `Identifier` and `VariableStmt` - Thought this would be too messy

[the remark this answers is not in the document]
sam-goodwin: >Do not support using an identifier in ForIn - Prefer to support all TS features, no matter the cost

Is the cost only incurred in particular circumstances, or for all `for` loops?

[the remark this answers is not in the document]
sam-goodwin: Only `for-in` from what I can see. Why does the value need to be set? Can't we replace all references to `array[i]` with the JSON path to the value?

src/asl.ts
```diff
@@ -3354,16 +3461,89 @@ export namespace ASLGraph {
...
 
+    /**
+     * Find the updated next value for all of the empty states.
+     * If the updated Next cannot be determined, do not remove the state.
+     */
+    const emptyTransitions = Object.fromEntries(
+      Object.entries(emptyStates).flatMap(([name, { Next }]) => {
+        const getNext = (
+          transition: string,
+          seen: string[] = []
```
sam-goodwin, line 3472: How long do we expect the chain to get? If it got really long, i'd recommend an immutable Set over an array for `O(lg n)` instead of `O(n)`

src/compile.ts
```diff
@@ -573,6 +573,12 @@ export function compile(
           node.declarationList.declarations.length === 1
         ) {
           return toExpr(node.declarationList.declarations[0], scope);
+        } else if (ts.isVariableDeclarationList(node)) {
+          return newExpr("VariableList", [
```
sam-goodwin, line 577: Shouldn't it be called `VariableDeclList`?

src/statement.ts
```diff
@@ -75,6 +76,21 @@ export class VariableStmt<
   }
 }
 
+export type VariableListParent = ForStmt;
+
+export class VariableList extends BaseNode<"VariableList", VariableListParent> {
+  readonly nodeKind: "Node" = "Node";
+
+  constructor(readonly decls: VariableStmt[]) {
```
sam-goodwin, line 84: It shouldn't be `VariableStmt`inside the list.

It's

```ts
VariableStmt(VariableDeclList(VariableDecl[]))
```

src/visit.ts
sam-goodwin: Should we name it with the verb as the first word

```ts
visitBlockStmt(..)
```

src/visit.ts
```diff
@@ -563,10 +566,34 @@ export function visitEachChild<T extends FunctionlessNode>(
...
+      variables,
+      isVariableStmt,
+      "Variables in a VariableList must be of type VariableStmt"
+    );
+    return new VariableList(variables) as T;
   }
   return assertNever(node);
 }
 
+function blockStmtVisitor(
```
sam-goodwin, line 581: better name: `visitBlockStmt`

src/asl.ts
thantos: I had a set, but changed to array to simplify. I don't expect it to get long, the array is reset for each state. Maybe 3-8 state names max. Average 0-1.

src/compile.ts
thantos: It could, we called `VariableDeclaration` a `VariableStmt`, so I thought `VariableList` was more neutral.

src/statement.ts
thantos: yeah, `VariableStatement` and `VariableDeclaration` are currently merged. `VariableStmt` matches the `VariableDeclaration`.

In a for loop, it contains a `VariableDeclarationList` -> `VariableDeclaration`.

src/declaration.ts
```diff
@@ -216,5 +228,46 @@ export class ArrayBinding extends BaseNode<"ArrayBinding"> {
   }
 }
 
+export type VariableDeclParent = ForInStmt | ForOfStmt | CatchClause;
+
+export class VariableDecl<
+  E extends Expr | undefined = Expr | undefined
+> extends BaseDecl<"VariableDecl", VariableDeclParent> {
+  constructor(readonly name: string | BindingPattern, readonly expr: E) {
```
sam-goodwin, line 236: `expr` should probably be `initializer`

src/declaration.ts
```diff
@@ -216,5 +228,46 @@ export class ArrayBinding extends BaseNode<"ArrayBinding"> {
...
+    super("VariableDecl");
+    if (isBindingPattern(name)) {
+      name.setParent(this);
+    }
+    if (expr) {
+      expr.setParent(this);
+    }
```
sam-goodwin, line 243: This can be simplified into one `if`

src/declaration.ts
```diff
@@ -216,5 +228,46 @@ export class ArrayBinding extends BaseNode<"ArrayBinding"> {
...
+
+  public clone(): this {
+    return new VariableDecl(
+      isBindingPattern(this.name) ? this.name.clone() : this.name,
+      this.expr?.clone()
+    ) as this;
+  }
+}
+
+export type VariableDeclListParent = ForStmt;
```
sam-goodwin, line 254: It can also be inside a `VariableStmt` right?

sam-goodwin (changes requested):

sam-goodwin (approved):

PR #335, opened 2022-07-20
feat(step-functions): json parse and stringify

thantos:
```ts
JSON.stringify
JSON.parse
```


src/asl.ts
```diff
@@ -1516,6 +1535,72 @@ export class ASL {
...
+      ) {
+        if (expr.expr.name === "stringify") {
+          const heap = this.newHeapVariable();
+
+          const objParamExpr = expr.args[0].expr;
+          if (!objParamExpr || isUndefinedLiteralExpr(objParamExpr)) {
+            // return an undefined variable
+            return {
+              jsonPath: heap,
+            };
```
sam-goodwin, line 1551: nice

src/asl.ts
```diff
@@ -1516,6 +1535,72 @@ export class ASL {
...
+          }
+
+          return this.evalExpr(
+            objParamExpr,
+            (_, { normalizeOutputToJsonPath }) => {
+              const objectPath = normalizeOutputToJsonPath().jsonPath;
```
sam-goodwin, line 1557: So, if i understand correclty, `normalizeOutputToJsonPath` is a way for the evaluator to "pull forward" the value as a JSON path only when needed. What happens if it is never called? Does the value not get evaluated and stored on the heap?

src/asl.ts
```diff
@@ -1516,6 +1535,72 @@ export class ASL {
...
+        } else if (expr.expr.name === "parse") {
+          const heap = this.newHeapVariable();
+
+          const objParamExpr = expr.args[0].expr;
+          if (!objParamExpr || isUndefinedLiteralExpr(objParamExpr)) {
+            // return an undefined variable
+            return {
+              jsonPath: heap,
+            };
+          }
```
sam-goodwin, line 1582: Redundant with `JSON.stringify`

src/asl.ts
```diff
@@ -1516,6 +1535,72 @@ export class ASL {
...
+                  "string.$": `States.StringToJson(${objectPath})`,
+                },
+                ResultPath: heap,
+                output: {
+                  jsonPath: `${heap}.string`,
+                },
+              };
+            }
+          );
+        }
```
sam-goodwin, line 1603: Looks almost exactly the same as JSON.stringify

test/__snapshots__/step-function.test.ts.snap
```diff
@@ -7483,6 +7511,85 @@ Object {
...
+exports[`stringify undefined 2 1`] = `
+Object {
+  "StartAt": "return JSON.stringify(undefined)",
+  "States": Object {
+    "return JSON.stringify(undefined)": Object {
+      "End": true,
+      "InputPath": "$.heap0",
+      "ResultPath": "$",
+      "Type": "Pass",
+    },
```
sam-goodwin, line 7588: What does this output? I suspect it will be an error or `null`?

sam-goodwin (changes requested):

src/asl.ts
```diff
@@ -382,6 +382,21 @@ export const isWaitState = (state: State): state is Wait => {
...
+   * If the output was a {@link ASLGraph.JsonPath}, the output is returned.
+   */
+  normalizeOutputToJsonPath: () => ASLGraph.JsonPath;
+  /**
+   * Callback provided to inject additional states into the graph.
+   * The state will be joined (@see ASLGraph.joinSubStates ) with the previous and next states in the order received.
+   */
+  addState: (state: ASLGraph.NodeState | ASLGraph.SubState) => void;
+}
+
```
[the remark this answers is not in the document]
thantos: It was a repeated pattern. 

1. If the output is already jsonpath, use the json path,
2. if the output is a constant, put it into a variable and use the new variable.

src/asl.ts
```diff
@@ -1433,66 +1457,61 @@ export class ASL {
...
+                    callbackfn.parameters.map((param, i) => [
+                      `${param.name}.$`,
+                      i === 0
+                        ? "$$.Map.Item.Value"
+                        : i == 1
+                        ? "$$.Map.Item.Index"
+                        : listPath,
+                    ])
+                  ),
+                },
```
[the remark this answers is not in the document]
thantos: It isn't specifically, but the output normalization pattern kept coming up and I was sick of copying and pasting the same code.

src/asl.ts
thantos: Its tied to the state returned, so it will be executed immediately before the operation (ex: stringify).

When we write the expression handlers, we ensure the heap value is only created as needed. In the cases with short circuiting, we do not hoist to the top of the expression, we localization the heap assignment. This is why the consuming expression determines how to use the output rather than the expression that generates the output. Sometimes the output is ignored (ExprStmt), sometimes its used as is (lambda input), sometimes it needs to be normalized (JSON.stringify).

```ts
JSON.stringify({ a: "a" });
```

result (as ts because I don't want to write out ASL)

```ts
const heap1 = { a: "a" } // an object literal cannot be an input to States.JsonToString, need a json path
const heap2 = { string: JSON.stringify(heap1) } // ugly, but we do this because step functions does not support intrinsic functions uniformly. (can't be in inputpath, result path, etc).
return heap2.string; // return a reference to the resulting string
```

```ts
x ? JSON.stringify({ a: "a" }) : JSON.stringify({ b: "b" });
```

result (as ts because I don't want to write out ASL)

```ts
let temp;
if(x) {
   const heap1 = { a: "a" } // an object literal cannot be an input to States.JsonToString, need a json path
   const heap2 = { string: JSON.stringify(heap1) } // ugly, but we do this because step functions does not support intrinsic functions uniformly. (can't be in inputpath, result path, etc).
   temp = heap2.string; // return a reference to the resulting string
} else {
   const heap3 = { b: "b" } // an object literal cannot be an input to States.JsonToString, need a json path
   const heap4 = { string: JSON.stringify(heap3) } // ugly, but we do this because step functions does not support intrinsic functions uniformly. (can't be in inputpath, result path, etc).
   temp = heap4.string; // return a reference to the resulting string
}
```

src/asl.ts
thantos: Actually need to update this to error, parse on nothing or undefined is an error.

src/asl.ts
thantos: hmm, they did end up similar. will merge.

test/__snapshots__/step-function.test.ts.snap
thantos: This would be a validation error, but its tough to catch without having types at transpile time. 

```ts
return JSON.stringify(undefined) // fails
return JSON.stringify(undefined) ?? null // succeeds
```

sam-goodwin (approved):

PR #336, opened 2022-07-20
feat(step-functions): support context object in sfn

thantos:
Adds access to the [context object](https://docs.aws.amazon.com/step-functions/latest/dg/input-output-contextobject.html) in step functions.

```ts
new StepFunction(stack, 'sfn', (input, context) => {
   return context.Execution.Name; // return the given or generated name of the execution.
})
```

> Note: `$$.Execution.Input` was left off as the input is present at `(input) =>{}`

> Note: `$$.Name` and `$$.Task` have been left out as their fields are abstracted from the use in Functionless. State names are dynamically generates, will change, and should not be relied on. The point at which these field, like name and time will be computed may also change. `$$.Task` will be used later (#79 ) to handle paused operations and will not be exposed directly to the user (maybe?).


src/asl.ts
```diff
@@ -1499,8 +1499,20 @@ export class ASL {
         // If the identifier references a parameter expression and that parameter expression
         // is in a FunctionDecl and that Function is at the top (no parent).
         // This logic needs to be updated to support destructured inputs: https://github.com/functionless/functionless/issues/68
-        if (ref && isParameterDecl(ref) && isFunctionDecl(ref.parent)) {
+        if (
+          ref &&
+          isParameterDecl(ref) &&
+          isFunctionDecl(ref.parent) &&
```
sam-goodwin, line 1505: With my change, this is changing - we can no longer rely on the top being a function decl. Need to check for parent is undefined.

src/step-function.ts
```diff
@@ -46,10 +46,34 @@ export type AnyStepFunction =
...
+    /**
+     * Format: ISO 8601
+     */
+    readonly StartTime: string;
+    readonly RoleArn: string;
+  };
+  readonly StateMachine: {
+    readonly Name: string;
+    readonly Id: string;
+  };
```
sam-goodwin, line 70: Should we document each of the fields using the official docs as a reference? Also, perhaps have separate interface declarations instead of nested object literal types.

sam-goodwin (changes requested):

src/asl.ts
thantos: Wouldn't parent be undefined for any top level function? Wouldn't we'd want `ref.parent === sfn.decl`?

src/step-function.ts
thantos: we could, you thinking `context.executionName` `context.stateMachineName` and then map the parameters to their json path?

`executionName => $$.Execution.Name`

sam-goodwin (approved):

src/asl.ts
thantos: `&& ref.parent === this.decl`

sam-goodwin (approved):

PR #337, opened 2022-07-20
feat: simplify compile-phase by annotating all functions with their AST

sam-goodwin:
This is the first step in supporting higher-order functions. Instead of only converting a function within specific contexts, such as `new Function` or `new StepFunction`, now all function declarations, function expressions and arrow expressions are wrapped in a call to `functionless.associateAST(func, new FunctionExpr([..], ..))` which associates a reference to a Function to its `FunctionExpr` or `FunctionDecl` AST form. Behind the scenes, a global `WeakMap` stores the association.

Fixes https://github.com/functionless/functionless/issues/309 - by consequence, we now support passing references to closures.

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -4,7 +4,7 @@ import {
   APIGatewayProxyResult,
   APIGatewayEventRequestContext,
 } from "aws-lambda";
-import { FunctionDecl, validateFunctionDecl } from "./declaration";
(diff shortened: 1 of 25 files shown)


sam-goodwin:
CLosed in favor of https://github.com/functionless/functionless/pull/372

PR #342, opened 2022-07-21
fix(compile): update external module indicator

### LINKED ISSUE #341: Bug: injected import("functionless") is not re-written as require if file has no imports
We have an edge-case bug in our transformer where `import * as functionless from "functionless"` is not re-written to a commonJS `require("functionless")` if the source file we're transforming doesn't import anything on its own.

```ts
function foo() {}
```

Our transformer re-writes this to:
```ts
import functionless_1 from "functionless";
functionless_1.associateAST(foo, new functionless_1.FunctionDecl(..))
function foo() { }
```

This should be re-written as:
```ts
const functionless_1 = __importStar(require("functionless"));
..
```

If you add any `import` to the source file, then the import is transformed properly.
```ts
import path from "path";
```

Perhaps this is a tsc optimization. We are using ts-patch which hacks in the transformers, so the changed source is unexpected.

thantos:
Closes #341

Typescript [determines whether or not to mutate the module part of a file](https://cs.github.com/microsoft/TypeScript/blob/28dc248e5c500c7be9a8c3a7341d303e026b023f/src/compiler/transformers/module/module.ts#L58-L64) based on a few factors, including [if the module is an external module](https://cs.github.com/microsoft/TypeScript/blob/28dc248e5c500c7be9a8c3a7341d303e026b023f/src/compiler/utilities.ts#L815-L817).

This computation is only done on [file parse](https://github.com/microsoft/TypeScript/blob/main/src/compiler/parser.ts#L769-L775), but sometimes functionless may change a non-external module into an external module when adding the functionless import at the top of the file (see #341). 

Added a hack to re-compute the module-ness of the file after we make our changes.

```ts
      functionlessContext.requireFunctionless &&
        // @ts-ignore
        !updatedSourceFile.externalModuleIndicator &&
        // @ts-ignore
        updatedSourceFile.setExternalModuleIndicator &&
        // @ts-ignore
        updatedSourceFile.setExternalModuleIndicator(updatedSourceFile);
```

### DIFF
--- a/src/compile.ts
+++ b/src/compile.ts
@@ -61,7 +61,7 @@ export function compile(
       }
 
       const functionlessContext = {
-        requireFunctionless: true,
+        requireFunctionless: false,
         get functionless() {
           this.requireFunctionless = true;
           return functionless;
--- a/src/compile.ts
+++ b/src/compile.ts
@@ -61,7 +61,7 @@ export function compile(
       }
 
       const functionlessContext = {
-        requireFunctionless: false,
+        requireFunctionless: true,
         get functionless() {
           this.requireFunctionless = true;
           return functionless;
@@ -83,7 +83,7 @@ export function compile(
         (stmt) => visitor(stmt) as ts.Statement
       );
 
-      return ts.factory.updateSourceFile(
+      const updatedSourceFile = ts.factory.updateSourceFile(
         sf,
         [
           // only require functionless if it is used.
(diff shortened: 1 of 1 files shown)


src/compile.ts
```diff
@@ -99,6 +99,24 @@ export function compile(
...
+       * https://github.com/microsoft/TypeScript/blob/main/src/compiler/parser.ts#L769-L775
+       */
+
+      functionlessContext.requireFunctionless &&
+        // @ts-ignore
+        !updatedSourceFile.externalModuleIndicator &&
+        // @ts-ignore
+        updatedSourceFile.setExternalModuleIndicator &&
+        // @ts-ignore
+        updatedSourceFile.setExternalModuleIndicator(updatedSourceFile);
```
sam-goodwin, line 116: Put the call to set in an if block instead of &&

sam-goodwin (commented):
Can we add a test?

sam-goodwin:
Rolling this into https://github.com/functionless/functionless/pull/337

PR #346, opened 2022-07-22
feat: remove type information from FunctionlessAST

### LINKED ISSUE #343: Remove `PropAccess` and `ElementAccess` `.type`
* [ ] Update EventBridge pattern to not use `.type`
  * determine if the type could cause failures at runtime instead or fail when ambiguous only.
* [ ] Update EventBridge transform to not use `.type`
* [ ] Remove `.type` from Prop and Element access AST objects.

sam-goodwin:
Closes #343 

This change removes the type information from Functionless AST and updates the Event Bridge logic that relied on it. We are removing type information from the FunctionlessAST so that it can be generated on-demand with an efficient require-hook transformer written in SWC. Interpreters should no longer depend on type information for their logic. If type information needs to be injected, then we will at it as a tsc-transformer, but that should be for use-cases more akin to deepkit.io, not for the entire AST tree.

Truthy conditions have regressed - we no longer allow for a boolean value to be evaluated as truthy, instead we will only check `{exists: true}` which will match false

A tsc-level diagnostic has been added to warn the user and suggest a workaround.
```ts
bus.when("rule", (event) => <any>event);
bus.when("rule", (event) => event.detail.a);
bus.when("rule", (event) => !event.detail.a);
bus.when("rule", (event) => !event.detail.a || event.detail.b === "b");
bus.when("rule", (event) => event.detail.a === "b" || !event.detail.b);
bus.when("rule", (event) => event.detail.a && event.detail.b === "b");
bus.when("rule", (event) => event.detail.a === "a" && event.detail.b);
bus.when("rule", (event) => event.detail.a || event.detail.b === "b");
bus.when("rule", (event) => event.detail.a === "a" || event.detail.b);
```


src/event-bridge/target-input.ts
```diff
@@ -170,18 +169,13 @@ export const synthesizeEventBridgeTargets = (
...
-        throw Error(
-          "Addition operator is only supported to concatenate at least one string to another value."
-        );
+        const val = `${exprToInternalLiteral(expr.left)}${exprToInternalLiteral(
+          expr.right
+        )}`;
+        return {
+          value: val,
+          type: "string",
+        };
```
thantos, line 178: would adding two numbers give the expected result?

src/validate.ts
```diff
@@ -735,8 +735,43 @@ export function validate(
...
+        // a && a.b
+        // a || a.b
+        (ts.isBinaryExpression(parent) &&
+          (parent.operatorToken.kind ===
+            ts.SyntaxKind.AmpersandAmpersandToken ||
+            parent.operatorToken.kind === ts.SyntaxKind.BarBarToken)) ||
+        // !a
+        // !a.b
+        (ts.isPrefixUnaryExpression(parent) &&
+          parent.operator === ts.SyntaxKind.ExclamationToken)
```
thantos, line 765: Wouldn't falsey work? `[{ exists: false }, 0, "", false]`

test/__snapshots__/validate.test.ts.snap
```diff
@@ -483,6 +483,66 @@ https://functionless.org/docs/error-codes#eventbus-rules-do-not-support-integrat
 
 [7m39[0m   return (func() as any) === \\"x\\";
 [7m  [0m [91m          ~~~~~~[0m
+[96mtest/test-files/event-bus.ts[0m:[93m43[0m:[93m34[0m - [91merror[0m[90m Functionless(10030): [0mEvent Bridge does not support a truthy comparison in
+
+https://functionless.org/docs/error-codes#event-bridge-does-not-support-a-truthy-comparison-in
+
+[7m43[0m bus.when(\\"rule\\", (event) => <any>event);
```
thantos, line 490: This should be a different error? Maybe its fine for now, now sure what it means for the event not to exist.

thantos (changes requested):

test/__snapshots__/validate.test.ts.snap
sam-goodwin: I just wanted to test when it was an `identifier`, perhaps it doesn't make sense but i as being thorough - truthy is not allowed

src/event-bridge/target-input.ts
sam-goodwin: Are you suggesting I error? if `expr.left` or `expr.right` aren't strings?

sam-goodwin: There is no way to know if this is two strings being concatenated unless one is a literal string type.

sam-goodwin: Figured out what you mean. Updated it to require left or right be a StringLiteralExpr

src/validate.ts
sam-goodwin: Deferring this to https://github.com/functionless/functionless/issues/365

thantos (approved):

PR #347, opened 2022-07-22
feat: add missing AST nodes to FunctionlessAST for 1:1 parity with TS

sam-goodwin:
This change will bring 1:1 parity of the FunctionlessAST with the TypeScript AST so that our annotation process can support all valid syntax. Interpreters can still ignore unsupported statements, but we should be able to totally represent programs with our AST.

BREAKING CHANGE: FunctionlessAST contract changed to be 1:1 with ES AST.


src/asl.ts
```diff
@@ -1060,6 +1076,32 @@ export class ASL {
...
       });
+    } else if (isDebuggerStmt(stmt) || isEmptyStmt(stmt)) {
+      return {
+        startState: "pass",
+        states: {
+          pass: {
+            Type: "Pass",
+          },
+        },
+      };
```
thantos, line 1087: nit: you can return `undefined` or `{ Type: "Pass" }`

src/asl.ts
```diff
@@ -3912,12 +3954,33 @@ function toStateName(node: FunctionlessNode): string {
...
+      isDebuggerStmt(node) ||
+      isDefaultClause(node) ||
+      isLabelledStmt(node) ||
+      isMethodDecl(node) ||
+      isPropDecl(node) ||
+      isSuperKeyword(node) ||
+      isSwitchStmt(node) ||
+      isWithStmt(node)
+    ) {
+      throw new SynthError(ErrorCodes.Unsupported_Feature, "");
```
thantos, line 3983: at least throw the kind?

src/compile.ts
```diff
@@ -817,10 +832,82 @@ export function compile(
...
+            [
+              // name
+              node.name ?? ts.factory.createIdentifier("undefined"),
+              // extends
+              node.heritageClauses?.flatMap((clause) =>
+                clause.token === ts.SyntaxKind.ExtendsKeyword &&
+                clause.types[0].expression !== undefined
+                  ? [toExpr(clause.types[0].expression, scope)]
+                  : []
+              )[0] ?? ts.factory.createIdentifier("undefined"),
```
thantos, line 864: why `[0]`?

src/visit.ts
```diff
@@ -241,6 +272,49 @@ export function visitEachChild<T extends FunctionlessNode>(
     }
 
     return new CatchClause(variableDecl, block) as T;
+  } else if (isClassDecl(node) || isClassExpr(node)) {
+    const heritage = node.heritage ? visitor(node.heritage) : undefined;
+
+    ensure(
+      heritage,
```
thantos, line 279: This will fail if undefined

src/visit.ts
```diff
@@ -563,6 +668,94 @@ export function visitEachChild<T extends FunctionlessNode>(
...
+        return updatedClause;
+      } else {
+        ensure(
+          updatedClause,
+          isSwitchClause,
+          "must be a CaseClause or DefaultClause"
+        );
+        return [updatedClause];
+      }
+    });
```
thantos, line 693: I've seen this pattern multiple times now, can be abstract it?

src/visit.ts
```diff
@@ -563,6 +668,94 @@ export function visitEachChild<T extends FunctionlessNode>(
...
+
+    const defaultClauses = clauses.filter(isDefaultClause);
+    if (defaultClauses.length === 1) {
+      if (!isDefaultClause(clauses[clauses.length - 1])) {
+        throw new Error(`only the last SwitchClause can be a DefaultClause`);
+      }
+    } else if (defaultClauses.length > 1) {
+      throw new Error(
+        `there must be 0 or 1 DefaultClauses in a single SwitchStmt, but found ${defaultClauses.length}`
+      );
```
thantos, line 703: SynthError?

thantos (changes requested):

src/asl.ts
sam-goodwin: `undefined` seems beter?

src/compile.ts
sam-goodwin: In the TS AST, the `heritageClauses` can be `extends` or `implements` - an array. Here, I am using `flatMap` to filter out only the `extends` clause which will result in an array of length 1 or an empty array if there is no `extends` clause. `[0]` grabs the value if it exists, otherwise I returned `undefined`.

src/visit.ts
sam-goodwin: Replaced with `.flatMap` and a helper everywhere

src/visit.ts
sam-goodwin: Not right now - we don't have SynthError for `visit`, plus these are internal errors. If we hit these errors it is entirely our fault, not the user's.

thantos: There is a `ErrorCodes.Unexpected_Error` for that purpose, you could just catch all errors in here and throw that.

src/asl.ts
```diff
@@ -4224,6 +4293,16 @@ function exprToString(expr?: Expr): string {
     return `await ${exprToString(expr.expr)}`;
   } else if (isPromiseExpr(expr) || isPromiseArrayExpr(expr)) {
     return exprToString(expr.expr);
+  } else if (isThisExpr(expr)) {
+    return "this";
+  } else if (isClassExpr(expr)) {
+    throw new Error(`ClassDecl is not supported in StepFunctions`);
```
thantos, line 4299: nit `SynthError` with unsupported?

src/compile.ts
```diff
@@ -826,10 +845,93 @@ export function compile(
...
+          ]);
+        } else if (ts.isDefaultClause(node)) {
+          return newExpr("DefaultClause", [
+            ts.factory.createArrayLiteralExpression(
+              node.statements.map((stmt) => toExpr(stmt, scope))
+            ),
+          ]);
+        } else if (ts.isWithStatement(node)) {
+          node.expression;
+          node.statement;
```
thantos, line 922: ??

src/guards.ts
```diff
@@ -10,66 +10,45 @@ export function isNode(a: any): a is FunctionlessNode {
...
-      isPropAssignExpr(a) ||
-      isReferenceExpr(a) ||
-      isStringLiteralExpr(a) ||
-      isTemplateExpr(a) ||
-      isTypeOfExpr(a) ||
-      isUnaryExpr(a) ||
-      isPostfixUnaryExpr(a) ||
-      isUndefinedLiteralExpr(a))
-  );
+  return isNode(a) && a.nodeKind === "Expr";
```
thantos, line 13: nice

thantos (approved):

src/compile.ts
sam-goodwin: Was checking their types LOL

thantos (approved):

thantos (changes requested):

src/asl.ts
```diff
@@ -696,9 +736,9 @@ export class ASL {
             assign: isForOfStmt(stmt)
               ? {
                   Type: "Pass",
-                  node: stmt.variableDecl,
+                  node: stmt.initializer,
                   InputPath: `${tempArrayPath}[0]`,
-                  ResultPath: `$.${stmt.variableDecl.name}`,
+                  ResultPath: `$.${stmt.initializer.getName()}`,
```
thantos, line 741: Why can't we be type safe here? This will be nasty when we go to add BindingElm and need to check the type anyways (soon).

```ts
if(isVariableDecl(stmt.initializer) && isBindingElm(stmt.initializer.name)) {
    throw...
}
const initializerName = stmt.initializer.name.name; 
```

thantos: Also I realize I wasn't checking this before in a typesafe way, good argument for name being an identifier... and disappointing....

src/visit.ts
```diff
@@ -414,6 +466,19 @@ export function visitEachChild<T extends FunctionlessNode>(
...
+  } else if (isPropDecl(node)) {
+    const name = visitor(node.name);
+    const initializer = node.initializer
+      ? visitor(node.initializer)
+      : undefined;
+    ensure(
+      name,
+      isPropName,
+      "a PropDecl's name must be an Identifier, StringLiteralExpr or ComputedPropNameExpr"
+    );
```
thantos, line 478: will fail if undefined.

src/visit.ts
```diff
@@ -634,6 +775,21 @@ export function visitBlock(
   });
 }
 
+function ensureSingleOrArray<T>(
```
thantos, line 778: document?

src/vtl.ts
```diff
@@ -345,8 +361,16 @@ export abstract class VTL {
     } else if (isBreakStmt(node)) {
       return this.add("#break");
     } else if (isCallExpr(node)) {
-      if (isReferenceExpr(node.expr)) {
-        const ref = node.expr.ref();
+      let expr = node.expr;
+      while (isParenthesizedExpr(expr)) {
+        // unwrap the (<expr>)
+        expr = expr.expr;
+      }
```
thantos, line 368: nit: add an unwrap function to `ParenthesizedExpr`?

```ts
unwrap(): Expr | undefined {
   if(isParenthesizedExpr(this.expr)) {
       return this.expr.unwrap()
   }
   return this.expr;
}
```

src/vtl.ts
```diff
@@ -481,18 +507,30 @@ export abstract class VTL {
...
+            return `${this.eval(expr.expr)}.add(${this.eval(
+              node.args[0].expr
+            )})`;
+          } else {
+            return `${this.eval(expr.expr)}.addAll(${this.addAll(
+              node.args
+                .map((arg) => arg.expr)
+                .filter((e): e is Expr => e !== undefined)
+            )})`;
+          }
```
thantos, line 529: Seems like this is missing the case when there is a mix of spread and not spread. Wouldn't the logic just be to addAll for spread and add single for other args?

thantos: Ohh, I see that this.addAll does that magically. Maybe document?

thantos (changes requested):

src/expression.ts
```diff
@@ -350,17 +427,18 @@ export class ObjectLiteralExpr extends BaseExpr<"ObjectLiteralExpr"> {
   }
 }
 
+export type PropName =
```
thantos, line 430: Also `NumberLiteral`

`export type PropertyName = Identifier | StringLiteral | NumericLiteral | ComputedPropertyName | PrivateIdentifier`

src/declaration.ts
```diff
@@ -220,13 +297,16 @@ export class ArrayBinding extends BaseNode<"ArrayBinding", VariableDecl> {
...
+    super("ArrayBinding", arguments);
   }
 
   public clone(): this {
     return new ArrayBinding(this.bindings.map((b) => b?.clone())) as this;
   }
+
+  public getName(): never {
+    throw new Error(`an ArrayBinding does not have a name`);
+  }
```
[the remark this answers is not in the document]
sam-goodwin: Removed.

src/node.ts
```diff
@@ -56,7 +59,38 @@ export abstract class BaseNode<
...
+        return;
+      } else if (isNode(node)) {
+        node.setParent(this as FunctionlessNode);
+      } else if (Array.isArray(node)) {
+        node.forEach(setParent);
+      }
+    };
+    for (const arg of args) {
+      setParent(arg);
+    }
```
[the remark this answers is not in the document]
sam-goodwin: I don't understand, what's wrong?

src/asl.ts
```diff
@@ -460,7 +491,9 @@ export class ASL {
     this.decl = visitEachChild(decl, function normalizeAST(node):
       | FunctionlessNode
       | FunctionlessNode[] {
-      if (isBlockStmt(node)) {
+      if (isParenthesizedExpr(node)) {
+        return visitEachChild(node.expr, normalizeAST);
+      } else if (isBlockStmt(node)) {
```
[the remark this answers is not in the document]
sam-goodwin: Removed, i was thinking of adding these defensively but decided against it.

src/node.ts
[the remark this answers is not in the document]
thantos: Nothing wrong, just saying your function also supports arrays of arguments so it could be written as:

```ts
const setParent = (node: any) => {
      if (!node) {
        return;
      } else if (isNode(node)) {
        node.setParent(this as FunctionlessNode);
      } else if (Array.isArray(node)) {
        node.forEach(setParent);
      }
    };
setParent(args);
```

thantos (approved):

src/node.ts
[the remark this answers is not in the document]
sam-goodwin: Actually, IArguments is not an array, it is only Iterable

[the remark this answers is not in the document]
sam-goodwin: Let me take another look, perhaps I got confused by a type error. I do remember doing it your way to begin with

src/vtl.ts
```diff
@@ -395,42 +419,46 @@ export abstract class VTL {
...
+            currentValue = this.newLocalVarName(),
+            currentIndex,
+            array,
+          ] = fn.parameters.map((param) => {
+            if (isIdentifier(param.name)) {
+              return `$${param.name.name}`;
+            } else {
+              throw new SynthError(
+                ErrorCodes.Unsupported_Feature,
+                "Binding variable assignment is not currently supported in Event Bridge rules and input transforms. https://github.com/functionless/functionless/issues/302"
```
thantos, line 451: VTL?

PR #356, opened 2022-07-24
fix(step-functions): synth time error on undefined assignment

### LINKED ISSUE #353: Step Function: `let v = undefined` assigns entire state to variable.
```ts
let hotelReservationResult = undefined;
```

```json
{
  "Type": "Pass",
  "ResultPath": "$.hotelReservationResult",
  "Next": "try"
}
```

thantos:
Fixes #353


sam-goodwin (approved):

thantos:
This problem is larger than I originally thought.

Step functions has poor to no support for undefined values.

We made the decision to not conflate undefined and null, rather to not support cases where step functions cannot represent undefined.

Undefined in sfn is represented as a missing field and can never be used as a value.

Not supported:
* Assignment `const a = undefined;` or `a = undefined`
* Object Literals `{ a: undefined }` (note: supporting undefined literals here would not be hard, but support for reference to something that is possible undefined is impossible without encoding all possible undefined paths as conditional statements).
* Array literals `[undefined]`  - It is not possible to define a positionally undefined value in an array, the only values that are valid in an array are present values of type, `string`, `boolean`, `number`, `null`, `array`, `object`, or `json path reference`.
* Integration inputs `task(undefined)` - This could be left up to the integration as the integration will see the Undefined Literal in the literal case, but a possibly undefined reference would need to be handled.

### Handling references to possibly undefined

What would it look like to handle possibly undefine reference?

Lets say we knew fields of an object literal that were undefined, we'd need to conditional on all permutations of possibly undefined values, leaving any that are undefined off of the assignment.

```ts
const obj = {
   a: maybeUndefinedA,
   b: maybeUndefinedB
}
``B

```json
{
   "Type": "Choice",
   "Choices": [{
       "And": [{ IsPresent: "$.maybeUndefinedA" }, { IsPresent: "$.maybeUndefinedB" }],
        "Next": "setAandB"
   }, {
        "IsPresent": "$.maybeUndefinedA"
        "Next": "setA"
   }, {
        "IsPresent": "$.maybeUndefinedB"
        "Next": "setB"
   } ],
   "Next": "setEmpty"
}
```

### Ternary

Another complicated case is Ternary. For ternary, we normalize the two sides of the expression by assigning the output of the left and the right to the same heap variable and then return said heap variable to whatever the parent of the expression is (including nothing in the case of `ExprStmt`).


```
if(condition)
   heap <- [left]
else
   heap <- [right]

left side of ternary <- heap
```

This essentially creates an assignment case, even though it isn't obvious. Whats worse is this creates and assignment even when the value is never used.

Only call `task` when `x` is truthy.

```ts
x ? task() : undefined;
```

When an undefined literal is used, this should be fine.

But this currently either fails or assigns the entire state to that heap variable.

What if the task could return `undefined`?

```ts
x ? task("a") : task("b") // possible undefined on both sides.
```

thantos:
## Current Design

Only allow `undefined` literal in conditional expression. All other use cases will fail.

```ts
// valid
if(x === undefined){}

//invalid
undefined
let x = undefined;
x = undefined; 
x ? undefined : y;
func(undefined)
{
   x : undefined
}
[undefined]
```

Type validation is also used to catch referenced based undefined so they can be handled

```ts
func = () => undefined | number

const x = func() // invalid, may be undefined
const x = func() ?? null // valid
```

src/asl.ts
```diff
@@ -1375,19 +1375,28 @@ export class ASL {
...
+  private eval(
+    expr: Expr,
+    allowUndefined: boolean = false
+  ): ASLGraph.NodeResults {
     // first check to see if the expression can be turned into a constant.
     const constant = evalToConstant(expr);
     if (constant !== undefined) {
-      const value = constant.constant as any;
+      const value = constant.constant;
+      if (!allowUndefined && typeof value === "undefined") {
```
sam-goodwin, line 1389: Why use typeof?

sam-goodwin (approved):

src/asl.ts
thantos: good question

PR #360, opened 2022-07-26
fix: support standard step function in apig, fix input escaping

thantos:
* Adds support for calling standard step functions in an api gateway method.
* Fixes an issue where the input to the machine was not correctly encoded when invoking the stepfunction.


src/step-function.ts
```diff
@@ -1272,6 +1273,49 @@ class BaseStandardStepFunction<
...
+        action: "StartExecution",
+        integrationHttpMethod: "POST",
+        options: {
+          ...options,
+          credentialsRole,
+          passthroughBehavior: aws_apigateway.PassthroughBehavior.NEVER,
+        },
+      });
+    },
+  };
```
sam-goodwin, line 1317: Should we share this across express and standard?

sam-goodwin (approved):

src/step-function.ts
thantos: uhhgg, I was trying to be lazy and got caught.

PR #361, opened 2022-07-26
fix: apig vtl stringify

thantos:
Do things that don't scale?

```ts
new AwsMethod(
  {
    httpMethod: "POST",
    resource: saga,
  },
  (
    $input: ApiGatewayInput<
      ApiRequest<
        {},
        Reservation & { [index: string]: ApiBody },
        { run_type?: RunType }
      >
    >
  ) => {
    return sagaFunction.func({
      input: {
        reservation: $input.data!,
        run_type: $input.params("run_type")!,
      },
    });
  },
  (result) => {
    return result.data;
  }
);
```

```velocity
{
"stateMachineArn":"arn:aws:states:us-east-1:180727192481:stateMachine:sagasagaFunction54D97EEF-B6WfpKLDY3QV",
"input":"{\"reservation\":$util.escapeJavaScript($input.json('$')),\"run_type\":\"$input.params('run_type')\"}"
}
```

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -654,6 +654,37 @@ export class APIGatewayVTL extends VTL {
   }
 
   /**
+   * Attempt to return the expression as a valid escaped json string.
+   *
+   * ```ts
+   * {
+   *    x: input
+   * }
+   * ```
+   *
+   * =>
+   *
+   * ```ts
+   * { "x": $input.json('$') }
+   * ```
+   *
+   * =>
+   *
+   * ```ts
+   * "{ \"x\": $util.escapeJavaScript($input.json('$')) }"
+   * ```
+   */
+  public stringify(expr: Expr): string {
+    const json = this.exprToJson(expr);
+    return `"${json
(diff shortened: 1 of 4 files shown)


src/api.ts
```diff
@@ -653,6 +653,37 @@ export class APIGatewayVTL extends VTL {
     return super.eval(node as any, returnVar);
   }
 
+  /**
+   * Attempt to return the expression as a valid escaped json string.
```
sam-goodwin, line 657: Does it fail sometimes?

sam-goodwin (approved):

src/api.ts
thantos: No idea. Don't have exhaustive tests or runtime tests, but it works for the two cases I tried. I assume it will need to get more complex as more cases are found or we re-architect to support the various output cases (ex: late resolution with a sub-ast instead of in place transform that can account for the need to be a string instead of an object.)

PR #363, opened 2022-07-26
feat(api-gateway,event-bridge): apigateway to event bus put events integration

thantos:
Depends on #347

* Support sending API gatgeway requests to Event Bridge
    * Caveat: only supports object literals
* Bug: API Gateway integration strings in SFN were incorrectly stringified (from #361).


src/event-bridge/event-bus.ts
```diff
@@ -345,6 +356,100 @@ abstract class EventBusBase<in Evnt extends Event, OutEvnt extends Evnt = Evnt>
...
+              throw new SynthError(
+                ErrorCodes.Expected_an_object_literal,
+                "API Gateway Integration with EventBus.putEvents expects object literals with no computed properties"
+              );
+            }
+
+            const objectProps = arg.properties.map((prop) => {
+              if (
+                !isPropAssignExpr(prop) ||
+                !(isIdentifier(prop.name) || isStringLiteralExpr(prop.name))
```
sam-goodwin, line 386: nit: are the parens necessary?

sam-goodwin: Would it be easier to check for what's allowed instead of the `!`? I am confused.

sam-goodwin: what about numeric properties?

test/api.test.ts
```diff
@@ -213,6 +214,63 @@ test("AWS integration with Standard Step Function using input data", () => {
...
+            body: {
+              event: { source: string; "detail-type": string; detail: any };
+            };
+          }>
+        ) => bus.putEvents($input.data.event),
+        () => {
+          return "success";
+        }
+      )
+  ).toThrow("object literal");
```
sam-goodwin, line 271: Is this the whole message it throws?

sam-goodwin (approved):

src/event-bridge/event-bus.ts
thantos: image

Looks like that is our miss

`export type PropertyName = Identifier | StringLiteral | NumericLiteral | ComputedPropertyName | PrivateIdentifier`

thantos:
Rebased onto #347 to reflect the change before it gets pushed.

sam-goodwin (commented):
Looks like there are changes brought across from my ast change? Merge problems?

Issue #364, opened 2022-07-27
Support de-structured parameter declarations


sam-goodwin:
- [ ] Appsync VTL
- [ ] API GW VTL
- [ ] Step Function ASL
- [ ] Event Bridge
```ts
function foo({a: b = c}, [d: e = f], ...g) {}
```

thantos:
https://github.com/functionless/functionless/issues/68

Also AppSync supports destructured (#299, #235 ) , but missed reduce I think.

https://github.com/functionless/functionless/blob/main/test/vtl.test.ts#L864
https://github.com/functionless/functionless/blob/main/test/vtl.test.ts#L898

Technically the support is in VTL, APIg may just need to be tested at runtime.

And missing a test for map/reduce.

thantos:
also #302

sam-goodwin:
> also #302

Ah shit. I tried to find it but didn't search "destructured". Perhaps we should start requiring keywords in issues to help with searching?

Issue #302, opened 2022-06-26
Support for deconstruction


thantos:
* Resource
  * [x] Lambda/Function - Native support
  * [ ] App Sync - in #299 
  * [ ] Step Functions
  * [ ] Event Bridge
  * [ ] API Gateway
* Locations 
   * [ ] Variable Declarations `const {a,b} = c`
   * [ ] Parameters `({ input }) =>`
   * [ ] For Loops `for( {a,b} in list )`
* Forms
   * Array `const [1,2] = list`
   * Object `const {a, b} = obj`
* Features
  * Default values `const { a = 1 } = obj`
  * Renamed `const

thantos:
StepFunctions and AppSync are done. Closing in favor of resource specific tickets: #377 #378

PR #366, opened 2022-07-28
feat(step-functions, appsync): binding pattern/destruction

### LINKED ISSUE #301: Support default expressions in bindings
```ts
const { a = {} } = b;
```

thantos:
Partial #68 Partial #364 Closes #301 Partial #302

* [x] Step Function
   * [x] Parameter Deconstruction
   * [x] Variable Deconstruction
   * [x] ForOf
   * [x] Map/ForEach
   * [x] $SFN Methods (Map, ForEach)
   * [x] Validate new "unsupported `{...rest}` error code
* [x] App Sync (continue from #299)
   * [x] Reduce
   * [x] Map
* [x] Tests
* [x] Deprecate Related Error Codes

Will follow this with Event Bridge to close the partial tickets above.

## Unsupported: `{...rest}` in StepFunctions will not be supported. 

It is not possible to correctly replicate the "everything-but" logic in an object binding `rest` logic from JS with ASL. ASL doesn't support the ability to delete or enumerate keys of a JSON object. Unlike VTL/AppSync where we can simply enumerate all keys BUT the keys we had bound into a new object. 😢


src/asl.ts
```diff
@@ -514,16 +516,24 @@ export class ASL {
...
+    const inputParam = decl.parameters[0];
 
     const states = this.evalStmt(this.decl.body);
 
+    // only evaluate the input parameter if its a binding parameter.
+    // We'll set input during context initialization
+    const inputParameterBinding =
+      inputParam && isBindingPattern(inputParam.name)
+        ? this.evalDecl(inputParam, { jsonPath: "$" })
+        : undefined;
```
sam-goodwin, line 528: nice

src/asl.ts
```diff
@@ -715,19 +733,57 @@ export class ASL {
...
+              ? stmt.initializer.name
+              : isVariableDecl(stmt.initializer) &&
+                isIdentifier(stmt.initializer.name)
+              ? stmt.initializer.name.name
+              : undefined;
+
+            if (initializerName === undefined) {
+              throw new SynthError(
+                ErrorCodes.Unexpected_Error,
+                "The left-hand side of a 'for...in' statement cannot be a destructuring pattern."
```
sam-goodwin, line 752: Is this because we don't yet support it for the $$.Map.Index.Value logic?

src/asl.ts
```diff
@@ -715,19 +733,57 @@ export class ASL {
...
-            "Destructured parameter declarations are not yet supported by Step Functions. https://github.com/functionless/functionless/issues/364"
-          );
-        }
+            return {
+              startState: "assignIndex",
+              node: stmt.initializer,
+              states: {
+                assignIndex: {
+                  Type: "Pass",
+                  InputPath: `${tempArrayPath}[0].index`,
```
sam-goodwin, line 762: What is `tempArrayPath`?

src/asl.ts
```diff
@@ -2764,35 +2753,62 @@ export class ASL {
...
-          if (ref === ref.parent.parameters[0]) {
-            return "@";
-          } else if (ref === ref.parent.parameters[1]) {
-            throw new Error(
-              "the 'index' parameter in a .filter expression is not supported"
-            );
-          } else {
+        }
+        return resolveRef(ref);
+        function resolveRef(ref: Decl): string {
```
sam-goodwin, line 2758: Is this function necessary?

src/asl.ts
```diff
@@ -2918,6 +2934,285 @@ export class ASL {
...
+        return this.evalAssignment(decl.name, initialValue);
+      } else if (decl.initializer === undefined) {
+        return undefined;
+      }
+
+      return this.evalExprToSubState(decl.initializer, (exprOutput) => {
+        return this.evalAssignment(decl.name, exprOutput);
+      });
+    };
+    const state = getState();
```
sam-goodwin, line 2952: Would prefer to just inline `(() => {})()` or a ternary. Const functions drive be nuts because code reads backwards

src/asl.ts
```diff
@@ -2918,6 +2934,285 @@ export class ASL {
...
+    } else {
+      const rest = pattern.bindings.find((binding) => binding?.rest);
+      if (rest && isObjectBinding(pattern)) {
+        // TODO: validator
+        throw new SynthError(
+          ErrorCodes.StepFunctions_does_not_support_destructuring_object_with_rest
+        );
+      }
+
+      // run each of the assignments as a sequence of states, they should not rely on each other.
```
sam-goodwin, line 3024: Can default values point to other bound names?

src/asl.ts
```diff
@@ -2918,6 +2934,285 @@ export class ASL {
...
+                );
+              }
+              return {
+                value: value.value[i],
+                containsJsonPath: value.containsJsonPath,
+              };
+            }
+          } else {
+            // when `name` is a bindingPattern, propertyName should always be present.
+            const propertyNameExpr = binding.propertyName ?? binding.name;
```
sam-goodwin, line 3052: I don't understand, can you provide syntax to explain?

What about this case?
```ts
const [{a}]
```

src/asl.ts
```diff
@@ -2918,6 +2934,285 @@ export class ASL {
...
+          }
+        })();
+
+        // when there is a default value
+        if (binding.initializer) {
+          // if there is a default value, update the output to reflect it
+          const valueStatesWithDefault = this.applyDefaultValue(
+            updatedValue,
+            binding.initializer
+          );
```
sam-goodwin, line 3091: Does this set the value with a Condition?

src/asl.ts
```diff
@@ -2918,6 +2934,285 @@ export class ASL {
...
+        return this.evalAssignment(binding.name, updatedValue);
+      });
+
+      // rest is only value for arrays
+      const restState = rest
+        ? (() => {
+            if (ASLGraph.isJsonPath(value)) {
+              return {
+                jsonPath: `${value.jsonPath}[${pattern.bindings.length - 1}:]`,
+              };
```
sam-goodwin, line 3116: super nice!

test/step-function.localstack.test.ts
```diff
@@ -1449,6 +1449,81 @@ localstackTestSuite("sfnStack", (testResource, _stack, _app) => {
...
+      new StepFunction(
+        parent,
+        "sfn",
+        async ({
+          a,
+          bb: { value: b },
+          c = "what",
+          arr: [d, , e, f = "sir", ...arrRest],
+          value,
+        }) => {
```
sam-goodwin, line 1465: Scanned through the tests, seems like we're missing initializers?

src/asl.ts
thantos: That is the error that typescript throws, playing with javascript, it does actually support destruction, but it will be an edge case as the left side is always a string.

image

src/asl.ts
thantos: We are doing tail recursion on the input array, which requires us to mutate an array variable. `tempArrayPath` is the copy of the original array we are mutating during the tail. (don't want to mutate the original array).

src/asl.ts
thantos: It calls itself recursively.

```ts
          } else if (isBindingElem(ref)) {
            if (isArrayBinding(ref.parent)) {
              return `${resolveRef( // <-- there.
                ref.parent.parent
              )}[${ref.parent.bindings.indexOf(ref)}]`;
            }

            const propName = ref.propertyName
              ? isIdentifier(ref.propertyName)
                ? ref.propertyName.name
                : isStringLiteralExpr(ref.propertyName)
                ? ref.propertyName.value
                : evalToConstant(ref.propertyName)?.constant
              : isIdentifier(ref.name)
              ? ref.name.name
              : undefined;

            if (!propName) {
              throw new SynthError(
                ErrorCodes.StepFunctions_property_names_must_be_constant
              );
            }

            return `${resolveRef(ref.parent.parent)}[${propName}]`; // <-- and there
          }
```

src/asl.ts
thantos: uhh, I guess they can... I didn't know that was possible.

```ts
const { a, b = a } = { a: "1" }
b // '1'
```

Well I guess I have more work to do :-P

Though in theory this should work because the resulting names will be public and in order

```ts
const { c = d, d } = { d: "1" }

// Uncaught ReferenceError: d is not defined
```

Will add tests

src/asl.ts
thantos: `ArrayPattern([BindingElm(name=ObjectPattern([BindingElm(name=Ident("a"))]))])`

But this comment is only relevant to ObjectPattern and not ArrayPattern.

```ts
{a:{b}}
// ObjectPattern([BindingElm(name=ObjectBinding([BindingElm(name=Ident("b"))]),propertyName=Ident("a"))])
{a:b}
// ObjectPattern([BindingElm(name=Ident("b"),propertyName=Ident("a"))])
```

Basically, a binding element either outputs a new name (`name=ident || name=iden && propName=ident`) or chains to a nested binding pattern using the property (`propertName=ident && name=pattern`)

src/asl.ts
thantos: yeah

src/asl.ts
thantos: right? haha.

test/step-function.localstack.test.ts
thantos: `f = "sir"`

sam-goodwin: Hurrr durrr

src/asl.ts
thantos: Worked!

src/asl.ts
sam-goodwin: why is `length` 1 there?

src/asl.ts
```diff
@@ -2774,37 +2772,64 @@ export class ASL {
...
-          } else {
+        }
+        return resolveRef(ref);
+        function resolveRef(ref: Decl): string {
+          if (isParameterDecl(ref)) {
+            if (ref.parent !== predicate) {
+              throw new Error(
+                "cannot reference a ParameterDecl other than those in .filter((item, index) =>) in a JSONPath filter expression"
+              );
+            }
```
sam-goodwin, line 2783: SynthError?

src/asl.ts
```diff
@@ -2774,37 +2772,64 @@ export class ASL {
...
+            if (ref === ref.parent.parameters[0]) {
+              return "@";
```
sam-goodwin, line 2785: what is the `@` for? quick comment?

src/asl.ts
```diff
@@ -2774,37 +2772,64 @@ export class ASL {
...
+            } else if (ref === ref.parent.parameters[1]) {
+              throw new Error(
+                "the 'index' parameter in a .filter expression is not supported"
+              );
+            } else {
+              throw new Error(
+                "the 'array' parameter in a .filter expression is not supported"
+              );
```
sam-goodwin, line 2793: `SynthError`?

src/asl.ts
```diff
@@ -2774,37 +2772,64 @@ export class ASL {
...
+            isVariableDecl(ref) ||
+            isFunctionDecl(ref) ||
+            isClassDecl(ref) ||
+            isClassMember(ref) ||
+            isGetAccessorDecl(ref) ||
+            isSetAccessorDecl(ref)
+          ) {
             throw new Error(
-              "the 'array' parameter in a .filter expression is not supported"
+              `cannot reference a ${ref.kind} within a JSONPath .filter expression`
```
sam-goodwin, line 2828: will have to change this to `ref.kindName` when you merge.

test/step-function.localstack.test.ts
```diff
@@ -1460,6 +1460,7 @@ localstackTestSuite("sfnStack", (testResource, _stack, _app) => {
           a,
           bb: { value: b, [`${"a"}${"b"}`]: r },
           c = "what",
+          m = c,
           arr: [d, , e, ...arrRest],
           arr2: [f = "sir"],
           value,
```
sam-goodwin, line 1466: do we need tests for numeric property names? Or are they impossible in ASL?

```ts
{1: a}
```

src/vtl.ts
```diff
@@ -735,6 +717,63 @@ export abstract class VTL {
...
+      const variablePrefix = isInTopLevelScope(decl) ? `$context.stash.` : `$`;
+      if (isBindingPattern(decl.name)) {
+        if (!(decl.initializer || initialValueVar)) {
+          throw new SynthError(
+            ErrorCodes.Unexpected_Error,
+            "Expected an initializer for a binding pattern assignment"
+          );
+        }
+        const right = initialValueVar ?? this.var(decl.initializer!);
+        this.evaluateBindingPattern(decl.name, right);
```
sam-goodwin, line 764: nit: inline `right`

src/validate.ts
```diff
@@ -247,6 +247,15 @@ export function validate(
             ),
           ];
         }
+      } else if (ts.isBindingElement(node)) {
+        if (node.dotDotDotToken && ts.isObjectBindingPattern(node.parent)) {
```
sam-goodwin, line 251: `node.parent` is so useful. SWC doesn't have it for memory optimization reasons, it's a bummer.

sam-goodwin (approved):

src/asl.ts
thantos: length is the length of the name of the index

indices are "0", "1", "2", so length is the string length of each of the numeric strings.

`"0".length`

src/asl.ts
thantos: Most of this was your code that I just wrapped in a new method, so I am not 100% sure how it works. I think its the `@` in the json path filter expression.

src/asl.ts
thantos: :+1: you may have changed it, we'll see how the merge goes (this was existing code moved a bit)

src/validate.ts
thantos: thats too bad, always need to do top down I guess

src/vtl.ts
thantos: :+1:

test/step-function.localstack.test.ts
thantos: Latest wins

image

src/asl.ts
```diff
@@ -2754,13 +2756,32 @@ export class ASL {
...
       );
     }
 
     const toFilterCondition = (expr: Expr): string => {
-      if (isBinaryExpr(expr)) {
+      const constant = evalToConstant(expr);
+      if (constant) {
+        if (typeof constant.constant === "string") {
+          return `'${constant.constant.replace(/'/g, "\\'")}'`;
+        }
```
sam-goodwin, line 2770: comment on why this magic is needed

src/asl.ts
```diff
@@ -2754,13 +2756,32 @@ export class ASL {
...
+        if (
+          typeof constant.constant === "object" &&
+          constant.constant !== null
```
sam-goodwin, line 2773: I've always wondered why `null` is considered an object

sam-goodwin: https://stackoverflow.com/questions/18808226/why-is-typeof-null-object

test/__snapshots__/serialize.test.ts.snap
```diff
@@ -16065,6 +16065,150 @@ function __f0() {
 "
 `;
 
+exports[`serialize sfn sfn 5`] = `
```
sam-goodwin, line 16068: What change added this?

src/asl.ts
thantos: I am not sure, actually, I copies it from the `isStringLiteralExpr` I removed (because evalToConstant covers the case).

test/__snapshots__/serialize.test.ts.snap
thantos: Good question.

thantos: Fixed.

test/test-files/step-function.ts
```diff
@@ -650,3 +650,33 @@ new StepFunction(stack, "fn", async (input: { arr: { name: string }[] }) => {
...
+
+new StepFunction(stack, "fn", async () => {
+  const obj = { 1: "value" };
+  const arr = [1];
+
+  // valid
+  obj["1"];
+  // valid
+  arr[0];
+});
```
sam-goodwin, line 682: What happens when the type is `any`?

thantos: This would be a very minor edge case as we don't support references

thantos: A quick test says it fails when the type is `any`, but the error is from an existing validation which checks if the value is a constant.

thantos: > test/test-files/step-function.ts:667:7 - error Functionless(10025): StepFunctions Invalid collection access

sam-goodwin (approved):

PR #370, opened 2022-07-29
feat: update NodeKind to be a numerical Enum and add support for s-expression encoding

sam-goodwin:
This change optimizes and simplifies our AST representation by replacing the `kind` property in Node to be a numerical enum. With it, we will be able to support a compact s-expression encoding of the AST with a simple tuple.

```ts
new Identifier("id");
// s-expr
[NodeKind.Identifier, "id"]
// more compactly
[37, "id"]
```

Also as part of this change, took the time to generaize `visitEachChild` and `Node.clone()`, and moved all validations into the Node classes. Also standardized our AST validation messages.


src/event-bridge/target-input.ts
```diff
@@ -228,7 +229,10 @@ export const synthesizeEventBridgeTargets = (
         if (isPropAssignExpr(expr)) {
           const name = isIdentifier(expr.name)
             ? expr.name.name
-            : assertString(evalToConstant(expr.name)?.constant, expr.name.kind);
+            : assertString(
+                evalToConstant(expr.name)?.constant,
+                getNodeKindName(expr.name.kind)
```
thantos, line 234: `.kindName`?

thantos (approved):

src/integration.ts
```diff
@@ -20,7 +20,7 @@ export const isIntegration = <I extends IntegrationInput<string, AnyFunction>>(
   i: any
 ): i is I => typeof i === "object" && "kind" in i;
 
-export type IntegrationCallExpr = Omit<CallExpr, "expr"> & {
+export type IntegrationCallExpr = CallExpr & {
   expr: ReferenceExpr<Integration>;
 };
```
thantos, line 25: If you don't omit, it will union the expr as `undefined | ReferenceExpr` or at least I thought.

thantos (approved):

src/integration.ts
sam-goodwin: It won't union, it will intersect, so `expr` can only be a `ReferenceExpr<Integration>`

PR #372, opened 2022-07-31
feat: emit s-expressions to compiled .js and emit corresponding function expr and arrow expressions

sam-goodwin:
- [x] transform functions into either a `FunctionExpr` or `ArrowFunctionExpr` depending on the syntax used instead of always creating a `FunctionDecl` node type.
- [x] update interpreters to detect the root function by `func.parent === undefined` instead of relying on `isFunctionDecl`.  
- [x] instead of importing `functionless` and generating code that instantiates FunctionlessNodes, inject a dependency-less `register` function into all modules and use it to decorate closures with an s-expression.
```ts
function register_8269d1a8(func, ast) {
  func[Symbol.for("functionless:AST")] = ast;
  return func;
}
```
- [x] emit a compact s-expression form of the FunctionlessAST instead of long, verbose new-expresions
```ts
register_8269d1a8(
  () => {
    2 + (1 + 2);
  },
  () => [
    3,
    [],
    [8, [[28, [6, [42, 2], "+", [47, [6, [42, 1], "+", [42, 2]]]]]]],
  ]
)
```
- [x] intercept all `<expr>.bind(..args)` patterns to capture the BoundThis, BoundArgs and TargetFunction and expose them as properties on the bound function
```ts
function bind_8269d1a8(func, self, ...args) {
  const tmp = func.bind(self, ...args);
  if (typeof func === "function") {
    func[Symbol.for("functionless:BoundThis")] = self;
    func[Symbol.for("functionless:BoundArgs")] = args;
    func[Symbol.for("functionless:TargetFunction")] = func;
  }
  return tmp;
}

// transform this
foo.bind(self, arg)
// into a call that intercepts the arguments
bind_8269d1a8(foo, self, arg)
```
- [x] update `reflect` to parse the s-expression form of a Function's AST into a FunctionlessNode instance and cache the result in a global weak map of closure -> Node.

BREAKING CHANGE: all functions are now wrapped in a call to decorate them with the AST form and transformed code now contains s-expression arrays instead of instantiated FunctionlessNode classes.


thantos:
>  update interpreters to detect the root function by func.parent === undefined instead of relying on isFunctionDecl.

Should we use `func.parent === this.decl` type logic instead of `func.parent === undefined`?

Once we start having multiple higher order functions, current function.parent !== resource function.

```
const myFunc = (input: string) => `${input}suffix`;
const myHandler = (input: { value: string} ) => `${input}suffix`;
const sfn new StepFunction(this, 'sfn', myHandler);
```

myHandler is the root, but myFunc will also have `func.parent === undefined`.

src/compile.ts
```diff
@@ -80,8 +74,139 @@ export function compile(
...
+          [param("func"), param("ast")],
+          undefined,
+          ts.factory.createBlock([
+            // func[Symbol.for("functionless:ast")] = ast;
+            setSymbol("func", ReflectionSymbolNames.AST, "ast"),
+
+            // return func;
+            ret("func"),
+          ])
+        ),
```
thantos, line 99: nit: could these be created in the parent closure like `functionless` to reduce the number of time it is run?

src/compile.ts
```diff
@@ -80,8 +74,139 @@ export function compile(
...
+            // const tmp = func.bind(self, ...args)
+            ts.factory.createVariableStatement(
+              undefined,
+              ts.factory.createVariableDeclarationList(
+                [
+                  ts.factory.createVariableDeclaration(
+                    "tmp",
+                    undefined,
+                    undefined,
+                    ts.factory.createCallExpression(
```
thantos, line 125: nit: these are large, create a helper method?

src/compile.ts
```diff
@@ -101,42 +226,103 @@ export function compile(
...
+          checker.isEventBusWhenFunction(node) ||
+          checker.isNewEventTransform(node) ||
+          checker.isNewFunctionlessFunction(node) ||
+          checker.isNewRule(node) ||
+          checker.isNewStepFunction(node) ||
+          checker.isReflectFunction(node) ||
+          checker.isRuleMapFunction(node)
+        ) {
+          // if this is a Functionless primitive, transform its arguments to wrap functions in their AST form
+          return transform(node);
```
thantos, line 242: What will it take to remove this? tsdoc annotations in the resource?

```ts
class StepFunction {
   /**
     *  @functionless-reflect decl - reflect the callback passed to or defined within the StepFunction resource.
     */ 
   constructor(decl:   )
}
```

src/compile.ts
```diff
@@ -101,42 +226,103 @@ export function compile(
...
+        } else if (
+          ts.isCallExpression(node) &&
+          ts.isPropertyAccessExpression(node.expression) &&
+          ts.isIdentifier(node.expression.name) &&
+          node.expression.name.text === "bind"
+        ) {
+          // potentially a func.bind(self, ..args) call
```
thantos, line 249: what happens if it isn't?

src/compile.ts
```diff
@@ -101,42 +226,103 @@ export function compile(
...
+          // wrap in the generated bind function wrapper
+          return ts.factory.createCallExpression(bindName, undefined, [
+            node.expression.expression,
+            ...node.arguments,
+          ]);
+        } else if (ts.isBlock(node)) {
```
thantos, line 255: What if the function decl is not in a block? like statements in a swtich block?

src/compile.ts
```diff
@@ -336,47 +366,49 @@ export function compile(
...
+          if (
+            !ts.isFunctionDeclaration(impl) &&
+            !ts.isArrowFunction(impl) &&
+            !ts.isFunctionExpression(impl) &&
+            !ts.isConstructorDeclaration(impl) &&
+            !ts.isMethodDeclaration(impl)
+          ) {
+            throw new Error(
+              `Functionless reflection only supports function parameters with bodies, no signature only declarations or references. Found ${impl.getText()}.`
+            );
```
thantos, line 379: SynthError?

src/compile.ts
```diff
@@ -336,47 +366,49 @@ export function compile(
...
-        const body = ts.isBlock(impl.body)
-          ? toExpr(impl.body, scope ?? impl)
-          : newExpr(NodeKind.BlockStmt, [
-              ts.factory.createArrayLiteralExpression([
-                newExpr(NodeKind.ReturnStmt, [
-                  toExpr(impl.body, scope ?? impl),
+          if (impl.body === undefined) {
+            throw new Error(
+              `cannot parse declaration-only function: ${impl.getText()}`
+            );
```
thantos, line 385: Synth Error?

src/event-bridge/event-pattern/synth.ts
```diff
@@ -125,16 +125,12 @@ export const synthesizeEventPattern = (
...
-  if (isErr(predicate)) {
-    throw predicate.error;
-  } else if (!isFunctionDecl(predicate)) {
-    throw Error(
-      "Expected parameter to synthesizeEventPattern to be compiled by functionless."
-    );
-  }
-  const [eventDecl = undefined] = (<FunctionDecl>predicate).parameters;
+  // @ts-ignore
+  const func = validateFunctionLike(predicate, "synthesizeEventPattern");
```
thantos, line 131: Instead of `ts-ignore`, should we just type the predicate as `unknown as FunctionLike`? That way the rest of the type checking will work.

test/reflect.test.ts
```diff
@@ -210,8 +224,5 @@ test("computed object name", () => {
 
 test("err", () => {
   const fn = () => {};
-  const result = assertNodeKind<Err>(reflect(fn), NodeKind.Err);
-  expect(result.error.message).toEqual(
-    "Functionless reflection only supports function parameters with bodies, no signature only declarations or references. Found fn."
-  );
+  expect(reflect(fn)).toBeUndefined();
```
thantos, line 227: Should this test continue to error?

src/reflect.ts
```diff
@@ -28,9 +32,85 @@ import { AnyAsyncFunction, AnyFunction } from "./util";
...
+  } else if (typeof a === "function") {
+    if (a.name.startsWith("bound")) {
+    }
+    const ast = reflect(a);
+    return validateFunctionlessNode(ast, functionLocation, validate);
+  } else {
+    throw new SynthError(
+      ErrorCodes.FunctionDecl_not_compiled_by_Functionless,
+      `Expected input function to ${functionLocation} to be compiled by Functionless. Make sure you have the Functionless compiler plugin configured correctly.`
+    );
```
thantos, line 114: Could you test for this?

src/reflect.ts
```diff
@@ -28,9 +32,85 @@ import { AnyAsyncFunction, AnyFunction } from "./util";
...
+): FunctionLike<F> | Err | undefined {
+  if (func.name.startsWith("bound ")) {
+    // native bound function
+    const targetFunc = (<any>func)[ReflectionSymbols.TargetFunction];
+    if (targetFunc) {
+      return reflect(targetFunc);
+    } else {
+      throw new Error(
+        `bound function not compiled with Functionless, cannot introspect on its form`
+      );
```
thantos, line 46: SynthError?

src/reflect.ts
```diff
@@ -28,9 +32,85 @@ import { AnyAsyncFunction, AnyFunction } from "./util";
...
+  if (typeof astCallback === "function") {
+    if (!reflectCache.has(astCallback)) {
+      reflectCache.set(
+        astCallback,
+        parseSExpr(astCallback()) as FunctionLike | Err
+      );
+    }
+    return reflectCache.get(astCallback) as FunctionLike<F> | Err | undefined;
+  }
+  return undefined;
```
thantos, line 60: What does it mean to return undefined here?

src/reflect.ts
```diff
@@ -28,9 +32,85 @@ import { AnyAsyncFunction, AnyFunction } from "./util";
...
+  functionLocation: string,
+  validate: (e: FunctionlessNode) => e is E
+): E {
+  if (validate(a)) {
+    return a;
+  } else if (isErr(a)) {
+    throw a.error;
+  } else if (typeof a === "function") {
+    if (a.name.startsWith("bound")) {
+    }
```
thantos, line 107: ????????

thantos: Are we missing tests for this code?

src/reflect.ts
```diff
@@ -28,9 +32,85 @@ import { AnyAsyncFunction, AnyFunction } from "./util";
...
+export const isFunctionLikeOrErr = anyOf(isFunctionLike, isErr);
+
+export function validateFunctionLike(
+  a: any,
+  functionLocation: string
+): FunctionLike {
+  return validateFunctionlessNode(a, functionLocation, isFunctionLike);
+}
+
+function validateFunctionlessNode<E extends FunctionlessNode>(
```
thantos, line 96: document

src/reflect.ts
```diff
@@ -28,9 +32,85 @@ import { AnyAsyncFunction, AnyFunction } from "./util";
...
+export const ReflectionSymbols = {
+  AST: Symbol.for(ReflectionSymbolNames.AST),
+  BoundThis: Symbol.for(ReflectionSymbolNames.BoundThis),
+  BoundArgs: Symbol.for(ReflectionSymbolNames.BoundArgs),
+  TargetFunction: Symbol.for(ReflectionSymbolNames.TargetFunction),
+} as const;
+
+export const isFunctionLikeOrErr = anyOf(isFunctionLike, isErr);
+
+export function validateFunctionLike(
```
thantos, line 89: document

src/reflect.ts
```diff
@@ -28,9 +32,85 @@ import { AnyAsyncFunction, AnyFunction } from "./util";
...
 
 // to prevent the closure serializer from trying to import all of functionless.
 export const deploymentOnlyModule = true;
+
+export const ReflectionSymbolNames = {
+  AST: "functionless:AST",
+  BoundThis: "functionless:BoundThis",
+  BoundArgs: "functionless:BoundArgs",
+  TargetFunction: "functionless:TargetFunction",
+} as const;
```
thantos, line 78: Also add `:Reflect` here?

test/eventbus.test.ts
```diff
@@ -43,7 +43,8 @@ test("new transform without map", () => {
   const rule = new Rule(stack, "rule", bus, (_event) => true);
   const transform = new EventTransform((event) => event.source, rule);
 
-  expect(transform.targetInput.bind(rule.resource)).toEqual({
+  const result = transform.targetInput.bind(rule.resource);
+  expect(result).toEqual({
     inputPath: "$.source",
   } as aws_events.RuleTargetInputProperties);
 });
```
thantos, line 50: will we not support this?

thantos (changes requested):

src/compile.ts
sam-goodwin: It will eventually be run on every closure

src/compile.ts
sam-goodwin: If `func` is not a function, then no properties are set on the underlying object.
```ts
function bind_1(func, self, ...args) {
  const tmp = func.bind(self, ...args);
  if (typeof func === "function") {
    func[Symbol.for("functionless:BoundThis")] = self;
    func[Symbol.for("functionless:BoundArgs")] = args;
    func[Symbol.for("functionless:TargetFunction")] = func;
  }
  return tmp;
}
```

src/compile.ts
sam-goodwin: A FunctionDeclaration is always in a block - it's a statement. If it is in an expression-position, the it will be a FunctionExpression.

sam-goodwin: This particular piece of code is detecting function declarations (that only exist in a block) and hoisting a register call to the top of the block.

Function and Arrow expressions are detected during an ordinary visit because they can be registered in-line.

src/event-bridge/event-pattern/synth.ts
sam-goodwin: This `ts-ignore` was not required

src/reflect.ts
sam-goodwin: It means the function was not registered and cannot be reflected. The caller should handle the error.

src/reflect.ts
sam-goodwin: the code isn't broken, i just decided to move the logic to `reflect` instead of here and forgot to clean up

sam-goodwin: will add a test specifically for bound function

src/compile.ts
sam-goodwin: I don't expect the user to do anything for this case - it's internal.

sam-goodwin: You did remind me to add a test-case though

sam-goodwin: It only happens when we hoist registrations for function declarations in a block and that code just ignores any declaration-only functions - this is a bit like assertNever in that it catches a case that should never happen.

test/reflect.test.ts
sam-goodwin: No, callers of reflect throw this error, not `reflect` - i.e. `Function`, `StepFunction`, etc. call call `validateFunctionLike`

test/eventbus.test.ts
thantos: Ohh, I meant the inline bind. Are inline binds not supported? why was this changed?

src/checker.ts
```diff
@@ -442,12 +443,43 @@ export function makeFunctionlessChecker(
...
+      return id.parent.initializer === id;
+    } else if (ts.isPropertyAccessExpression(id.parent)) {
+      // event.bus
+      // ^T    ^F
+      return id.parent.expression === id;
+    }
+
+    return true;
+  }
+
```
thantos, line 472: Why not use the flags? Those seems to be correct? 

* https://github.com/functionless/functionless/blob/main/src/checker.ts#L774

or

```ts
                symbol?.valueDeclaration &&
                // eslint-disable-next-line no-bitwise
                (symbol.flags & ts.SymbolFlags.BlockScopedVariable ||
                  // eslint-disable-next-line no-bitwise
                  symbol.flags & ts.SymbolFlags.FunctionScopedVariable)
```

If not...

This seems like the wrong approach, to check how it is being used, missing other cases like being used in a binary expression, switch statement, conditional statement, return statement, and more.

sam-goodwin: Those cases don't matter - this function all analyzes them correctly. Not sure why this is the wrong approach?

I was unaware of the flags. Let me check on that

sam-goodwin: The problem is we are treating all identifiers as references and they are not. Why is this logic incorrect? The parent defines the interpretation of the identifier

thantos: If we keep this logic, should document which cases it is for and maybe update the name of the method to reflect that.

`isIdentifierUsedAsAVariableInitializer`?

also this seems backwards?

```ts
// { a: b = c }
//   ^T ^T  ^F
```

and what about ?

// const a = x ? b : c

sam-goodwin:
```ts
const a = x ? b : c
```

this would return false for a and true for the rest

sam-goodwin: > also this seems backwards?

Typo I think. My bad

thantos: Ohh, I see, you are encoding all of the cases where ident is used as a name instead of a variable. (I am an idiot and missed the `return true`). I think the flags are going to be the most accurate, they seemed to work for me without any new logic.

test/eventbus.test.ts
sam-goodwin: inlines are supported, I changed it because I was testing. Changed it back

src/checker.ts
sam-goodwin: I don't think flags work as `c` in the below code is a BlockScopedVariable
```ts
const d = "";
const {b: c = d} = {b: ""};
```

thantos (approved):

src/checker.ts
```diff
@@ -465,6 +494,30 @@ export function makeFunctionlessChecker(
...
+      //       ^F
+
+      // function foo() {}
+      //          ^F
+
+      // interface foo {}
+      //           ^F
+
+      // type A
+      //      ^F
```
thantos, line 515: love the docs.

src/compile.ts
```diff
@@ -74,8 +131,14 @@ export function compile(
         ts.factory.createStringLiteral("functionless")
       );
 
-      const registerName = ts.factory.createUniqueName("register");
-      const bindName = ts.factory.createUniqueName("bind");
+      sf.statements.find((stmt) => {
+        if (ts.isVariableStatement(stmt)) {
+        }
+        return false;
+      });
```
sam-goodwin: Oops, dead code - removed.

sam-goodwin:
BREAKING CHANGE: all functions are now wrapped in a call to decorate them with the AST form and transformed code now contains s-expression arrays instead of instantiated FunctionlessNode classes.

Issue #374, opened 2022-08-02
Total AST decoration, closure serialization from FunctionlessAST and higher-order function capabilities


sam-goodwin:
Now that our AST is 1:1 with TS, we can replace the async closure serializer that uses the debugger API with one that simply serializes the FunctionlessAST -> TypeScript and runs the result through es-build to produce a final bundle.

- [x] https://github.com/functionless/functionless/issues/373
- [ ] ~~https://github.com/functionless/functionless/issues/239~~
- [x] remove the work that the compiler does to flatten references and move that logic into interpreters https://github.com/functionless/functionless/pull/376
- [x] remove `PromiseExpr` and `PromiseArrayExpr` 
- [x] missing AST information for function modifiers https://github.com/functionless/functionless/pull/382
- [x] missing AST information - `const`, `let` or `var` in VariableDeclList https://github.com/functionless/functionless/pull/389
- [x] missing AST information for `for await (const i of)` https://github.com/functionless/functionless/pull/389
- [x] missing AST information - while loops, for loops without BlockStmt, e.g. `while (cond) a = b`
- [x] missing AST information - the variable declarations in for-loops must use a VariableDeclList
- [ ] ~~unnecessary AST information - Argument node type has no purpose~~
- [x] ensure that swc-jest + our transformer can be run during tests
- [x] swap out the ts-transformer for swc and apply it as a require hook
- [ ] ~~update IAM policy inference to support exploring function references, recursion and higher-order functions, etc. https://github.com/functionless/functionless/issues/310~~
- [x] replace pulumi's closure serializer with one that translates FunctionlessAST into javascript instead of relying on debugger APIs and parsing function.toString() like how pulumi does it

sam-goodwin:
I'm running into a fundamental question: should integration discovery be bottom-up or top-down? We currently do it bottom-up by calling functions such as `isIntegrationCallPattern` on a node and expecting that to resolve to a single integration via `ReferenceExpr`. Now that the transformer logic is being removed, we have to resolve that `expr` to an Integration by walking backwards (as we've discussed). This is a bottom up algorithm that is usually ideal for type-checking algorithms. 

I'm not sure this is ideal for detecting all of the Integrations which is closer to an evaluation algorithm (explore all pathways of the code) - a top-down walker.

[...]

PR #376, opened 2022-08-02
feat: remove logic in compile.ts to flatten references to expressions

sam-goodwin:
Partial #374 

- [x] remove flattening logic from compile.ts
```ts
function foo() {
  $AWS.DynamoDB.GetItem(..)

  // before:
  CallExpr(ReferenceExpr(() => $AWS.DynamoDB.GetItem), ..)

  // now:
  CallExpr(PropAccessExpr(GetItem, PropAccessExpr(DynamoDB, ReferenceExpr(() => $AWS)))
}
```

- [x] update interpreters to resolve FucntionlessAST to ReferenceExpr as we can no longer rely on that work being done during tsc.

### DIFF
--- a/src/appsync.ts
+++ b/src/appsync.ts
@@ -466,14 +466,19 @@ function synthesizeFunctions(api: appsync.GraphqlApi, decl: FunctionLike) {
         // we find the range of nodes to hoist so that we avoid visiting the middle nodes.
         // The start node is the first node in the integration pattern (integ, await, or promise)
         // The end is always the integration.
+
         const end = getIntegrationExprFromIntegrationCallPattern(node);
 
-        const updatedChild = visitSpecificChildren(node, [end], (expr) =>
-          normalizeAST(expr, hoist)
-        );
-        // when we find an integration call,
-        // if it is nested, hoist it up (create variable, add above, replace expr with variable)
-        return hoist && doHoist(node) ? hoist(updatedChild) : updatedChild;
+        if (end) {
+          const updatedChild = visitSpecificChildren(node, [end], (expr) =>
+            normalizeAST(expr, hoist)
+          );
+
+          // when we find an integration call,
+          // if it is nested, hoist it up (create variable, add above, replace expr with variable)
+          return hoist && doHoist(node) ? hoist(updatedChild) : updatedChild;
+        }
+        return visitEachChild(node, normalizeAST);
       } else if (isVariableStmt(node) && node.declList.decls.length > 1) {
         /**
          * Flatten variable declarations into multiple variable statements.
--- a/src/appsync.ts
+++ b/src/appsync.ts
@@ -478,7 +478,6 @@ function synthesizeFunctions(api: appsync.GraphqlApi, decl: FunctionLike) {
           // if it is nested, hoist it up (create variable, add above, replace expr with variable)
           return hoist && doHoist(node) ? hoist(updatedChild) : updatedChild;
         }
-        return visitEachChild(node, normalizeAST);
       } else if (isVariableStmt(node) && node.declList.decls.length > 1) {
         /**
          * Flatten variable declarations into multiple variable statements.
--- a/src/integration.ts
+++ b/src/integration.ts
@@ -329,6 +329,18 @@ export function findDeepIntegrations(
   return nodes;
 }
 
+/**
+ * A bottom-up algorithm that determines the ONLY {@link Integration} value that the {@link node}
+ * will resolve to at runtime.
+ *
+ * If the {@link node} resolves to 0 or more than 1 {@link Integration} then `undefined` is returned.
+ *
+ * **Note**: This function is an intermediate helper until we migrate the interpreters to be more general
+ * (likely by migrating to top-down algorithms, see https://github.com/functionless/functionless/issues/374#issuecomment-1203313604)
+ *
+ * @param node the node to resolve the {@link Integration} of.
+ * @returns the ONLY {@link Integration} that {@link node} can resolve to, otherwise `undefined`.
+ */
 export function tryFindIntegration(
   node: FunctionlessNode
 ): Integration | undefined {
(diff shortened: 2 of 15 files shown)


src/integration.ts
```diff
@@ -301,20 +302,143 @@ export function makeIntegration<K extends string, F extends AnyFunction>(
...
+  function resolve(
+    node: FunctionlessNode | undefined,
+    defaultValue: FunctionlessNode | undefined
+  ): any[] {
+    if (node === undefined) {
+      if (defaultValue === undefined) {
+        return [];
+      } else {
+        return resolve(defaultValue, undefined);
+      }
```
thantos, line 407: This seems wrong? Default value can only be determined at runtime.

`const { a = func } = b ? { a: func2 } : { a: undefined }`

`a` is `func` if `b` is `falsey` (runtime) or `func2` is undefined (maybe runtime)

src/integration.ts
```diff
@@ -301,20 +302,143 @@ export function makeIntegration<K extends string, F extends AnyFunction>(
...
+          ? node.name.name
+          : isStringLiteralExpr(node.element) ||
+            isNumberLiteralExpr(node.element)
+          ? node.element.value
+          : undefined;
+        if (key !== undefined) {
+          return [(<any>expr)?.[key]];
+        }
+        return [];
+      });
```
thantos, line 438: `evalToConstant` to try to get things like template expressions and string concat?

src/integration.ts
```diff
@@ -301,20 +302,143 @@ export function makeIntegration<K extends string, F extends AnyFunction>(
...
+      }
+    } else if (isReferenceExpr(node) || isThisExpr(node)) {
+      return [node.ref()];
+    } else if (isIdentifier(node)) {
+      return resolve(node.lookup(), defaultValue);
+    } else if (isBindingElem(node)) {
+      return resolve(node.parent, node.initializer).flatMap((pattern) => {
+        if (isIdentifier(node.name)) {
+          return [pattern[node.name.name]];
+        } else {
```
thantos, line 416: what is pattern here?

src/util.ts
```diff
@@ -209,7 +209,14 @@ export function isConstant(x: any): x is Constant {
...
 ): Constant | undefined => {
-  if (
+  if (isIdentifier(expr)) {
+    // const decl = expr.lookup();
+    // if (decl) {
+    //   if (isVariableDecl(decl) && decl.initializer) {
+    //     return evalToConstant(decl.initializer);
+    //   }
+    // }
+  } else if (
```
thantos, line 219: delete?

src/node.ts
```diff
@@ -362,11 +374,9 @@ export abstract class BaseNode<
   /**
    * @returns a mapping of name to the node visible in this node's scope.
    */
-  public getLexicalScope(): Map<string, Decl> {
+  public getLexicalScope(): Map<string, Binding[1]> {
```
thantos, line 377: maybe create a new type and use it in binding? this is harder to follow

src/node.ts
```diff
@@ -105,6 +111,12 @@ export abstract class BaseNode<
     ];
   }
 
+  public fork<N extends this["parent"]>(node: N): N {
+    // @ts-ignore
+    node.parent = this;
+    return node;
+  }
+
```
thantos, line 119: doc?

src/integration.ts
```diff
@@ -301,20 +302,143 @@ export function makeIntegration<K extends string, F extends AnyFunction>(
...
+ *
+ * const tables = [table1, table2];
+ *
+ * const a = table1;
+ *    // ^ [table1]
+ *
+ * for (const a of tables) {
+ *   const b = a;
+ *          // ^ [table1, table2]
+ *
```
thantos, line 370: Is this implemented? I don't see any code that could return an array of more than 1

src/appsync.ts
```diff
@@ -466,14 +466,19 @@ function synthesizeFunctions(api: appsync.GraphqlApi, decl: FunctionLike) {
...
+        if (end) {
+          const updatedChild = visitSpecificChildren(node, [end], (expr) =>
+            normalizeAST(expr, hoist)
+          );
+
+          // when we find an integration call,
+          // if it is nested, hoist it up (create variable, add above, replace expr with variable)
+          return hoist && doHoist(node) ? hoist(updatedChild) : updatedChild;
+        }
+        return visitEachChild(node, normalizeAST);
```
thantos, line 481: You need to closure hoist in or nested nodes won't get hoisted. (Or just let the logic fall through because the one at the bottom already does this)

src/visit.ts
```diff
@@ -23,48 +23,38 @@ export function visitEachChild<T extends FunctionlessNode>(
...
-            return [];
-          } else if (isNode(transformed)) {
-            return [transformed];
-          } else {
-            // spread the nodes into the array
-            return transformed;
-          }
-        });
+  const args = node._arguments.map((argument) => {
+    if (!(typeof argument === "object" || Array.isArray(argument))) {
```
thantos, line 27: nit: the left will always eval to true when the right does

thantos (changes requested):

src/integration.ts
sam-goodwin: We care calling `resolve` on the `defaultValue` which can return 0, 1 or many possible values. In your case, we would have logic for the ConditionExpr that returns `[{a: func2}, {a: undefined}]` which then distributes to `[func2, undefined]` which then unions with `a = func` to create `[func, func2, undefined]` as the set of all possible values that `a` can be.

src/integration.ts
sam-goodwin: bad name - it's one of the possible values that `node.initializer` can be

src/integration.ts
sam-goodwin: Not yet, no. I implemented only the cases I needed for the tests to pass - incremental progress. I added the comments as I was thinking it through.

sam-goodwin: We would need to update the interpreter logic to support this case so it didn't matter.

src/appsync.ts
sam-goodwin: Good catch.

src/integration.ts
sam-goodwin: Let's expand on this in https://github.com/functionless/functionless/issues/310 ?

thantos (approved):

Issue #369, opened 2022-07-28
Step Functions: shadow variable names overwrites state


thantos:
```ts
new StepFunction(async()=>{
   const a = "a";
   for(const a of ["b"]) {
      ....
   }
   return a; // returns "b", should return "a".
});
```

PR #379, opened 2022-08-02
feat(step-functions): more flexible filter and map in step functions

### LINKED ISSUE #57: Step Functions filter using a Map Task
When filtering lists, e.g. `list.filter(..)`, we currently only support cases where the filter expression can be converted to JSON Path. We need to add support for using a Map Task which can call APIs, make Choices, etc.

Ex.
```ts
list.filter(item => {
  if(task(item)) {
    return false;
  } else {
    return true;
  }
})
```

### LINKED ISSUE #84: Step Functions: chained map, filter, etc.

### LINKED ISSUE #318: Step Functions: Support mutable variables outside of For and Map scopes.
* [x] For Loops
* [ ] Nested Closures (Maps, ForEach, etc)

```ts
new StepFunction(stack, 'sfn', async () => {
   let a = "";
   for(const x in [1,2,3]) {
      a = `${a}${x}`
   }
   return a;
});
```

thantos:
Closes #84 Closes #57 Closes #318 

* [x] Filter
   * [x] without restrictions
   * [x] Filter index argument
   * [x] Filter array argument
   * [x] Optimize: attempt json path first
   * [x] Update validation
* [x] Map and Foreach
   * [x] Support mutating outside variable
   * [x] Stop using map state
* [x] Undo changes to message board

REGRESSION: `.map` in StepFunctions will overwrite shadowed variable names. This is a known issue #369 and will be resolved together with for loops and other scopes.

Now supported:

## Filter

Previously multi-line functions, string templates, mutating external variables, and arbitrary commands like `waitFor` were not supported in the json path.

```ts
new StepFunction<
  {
    arr: { value: number; key?: string }[];
    key: number;
    key2: string;
  },
  { value: number; key?: string }[]
>(stack, "sfn2", async ({ arr, key }) => {
  return arr
    .filter(({ value }) => value <= key)
    .filter((item) => {
      const { key: itemKey } = item;
      $SFN.waitFor(1);
      return itemKey === `hi${key}`;
    });
});
```

## Map and ForEach

```ts
new StepFunction<
  {
    arr: { value: number; key?: string }[];
    key: number;
    key2: string;
  },
  { value: number; key?: string }[]
>(stack, "sfn2", async ({ arr, key }) => {
  let a = "";
  return arr
    .map((item) => {
      a = `${a}${item.value}`;
      return item.key;
    });
});
```


src/asl.ts
```diff
@@ -2775,12 +2805,8 @@ export class ASL {
...
           constant.constant !== null
         ) {
-          throw new SynthError(
-            ErrorCodes.StepFunction_invalid_filter_syntax,
-            `Filter expressions do not support object or array literals, found: ${JSON.stringify(
-              constant.constant
-            )}`
-          );
+          // will be caught, try to compile with ASLGraph instead
+          throw new FilterJsonPathUnsupported("Use ASLGraph");
```
sam-goodwin, line 2809: Is it possible to have a functional API that doesn't depend on throwing this to signal that ASLGraph should be used? return `undefined`?

src/asl.ts
```diff
@@ -2850,58 +2866,154 @@ export class ASL {
...
+        // parse the arrStr back to json and assign over the filter result value.
+        end: {
+          startState: "format",
+          states: {
+            format: {
+              Type: "Pass",
+              // filterResult = { result: JSON.parse(filterResult.arrStr + "]") }
+              Parameters: {
+                // cannot use intrinsic functions in InputPath or Result
+                "result.$": `States.StringToJson(States.Format('{}]', ${filterResult}.arrStr))`,
```
sam-goodwin, line 2999: Where does the leading `[` come from?

thantos: the `[null` above. Set while also setting the temp array.

src/asl.ts
thantos: Not in a clean way, I'd also prefer that, but this method is always embedded in string templates `something.${filterToJsonPath()}`. Would need to re-rewrite them all to call the method and check for null and then chain it down. This was the lowest touch and the error is always caught.

thantos: I'll play with it a bit.

thantos: just made it monadic with `string | undefined`. accident `undefined` in string templates scares me, but I think I got them all.

test/test-files/step-function.ts
```diff
@@ -610,41 +610,36 @@ new StepFunction(stack, "sfn", async (input: { arr: string[] }) => {
...
 
 const someFunctionReference = () => {
   return true;
 };
 
+new StepFunction(stack, "fn", async () => {
+  [1, 2, 3].filter(someFunctionReference);
+  [1, 2, 3].map(someFunctionReference);
+  [1, 2, 3].forEach(someFunctionReference);
+});
```
sam-goodwin, line 627: This can likely be supported now that we have enhanced parsing of all closures

sam-goodwin (approved):

test/test-files/step-function.ts
thantos: yeah, totally, needs to be tested

test/step-function.localstack.test.ts
```diff
@@ -515,12 +512,17 @@ localstackTestSuite("sfnStack", (testResource, _stack, _app) => {
...
       return new StepFunction(parent, "sfn2", async (input) => {
+        let a = "";
         const l = (await func()).map((x) => `n${x}`);
         const l2 = input.arr.map((x) => `n${x}`);
-        return `${l[0]}${l[1]}${l[2]}${l2[0]}${l2[1]}${l2[2]}`;
+        input.arr.map((x) => {
+          a = `${a}a${x}`;
+          return a;
+        });
+        return `${l[0]}${l[1]}${l[2]}${l2[0]}${l2[1]}${l2[2]}${a}`;
```
sam-goodwin, line 522: Nice to see this supported!

src/asl.ts
```diff
@@ -2482,6 +2407,15 @@ export class ASL {
...
       mapOrParallelClosure === undefined ||
+      // TODO: this implementation is specific to the current state. Try/Catch/Finally needs to be generalize based on the generated ASL, not the AST.
+      // https://github.com/functionless/functionless/issues/385
+      (isArgument(mapOrParallelClosure.parent) &&
+        isCallExpr(mapOrParallelClosure.parent.parent) &&
+        isPropAccessExpr(mapOrParallelClosure.parent.parent.expr) &&
+        (mapOrParallelClosure.parent.parent.expr.name.name === "map" ||
+          mapOrParallelClosure.parent.parent.expr.name.name === "filter" ||
+          mapOrParallelClosure.parent.parent.expr.name.name === "forEach") &&
+        !isReferenceExpr(mapOrParallelClosure.parent.parent.expr.expr)) ||
```
sam-goodwin, line 2418: Is this fragile to ParnesExpr?

src/asl.ts
```diff
@@ -2837,60 +2792,494 @@ export class ASL {
...
+    const [itemParameter, indexParameter, arrayParameter] =
+      predicate.parameters;
+
+    const assignParameters = ASLGraph.joinSubStates(
+      predicate,
+      itemParameter
+        ? this.evalDecl(itemParameter, {
+            jsonPath: indexParameter
+              ? `${filterResult}.arr[0].item`
+              : `${filterResult}.arr[0]`,
```
sam-goodwin, line 2852: It's not clear to me why we are appending these to the json paths.

src/asl.ts
```diff
@@ -2837,60 +2792,494 @@ export class ASL {
...
+                ResultPath: "$",
+                Next: ASLGraph.DeferNext,
+              }),
+              ResultSelector: {
+                "arr.$": "$",
+                arrStr: "[null",
+              },
+              ResultPath: filterResult,
+              Next: "check",
+            }
```
sam-goodwin, line 2897: So if you use index, then you can't mutate a variable outside the scope?

src/asl.ts
```diff
@@ -2837,60 +2792,494 @@ export class ASL {
...
+                jsonPath: listPath,
+              })
+            : undefined,
+          this.evalStmt(callbackfn.body, {
+            End: undefined,
+            // write over the input value to avoid creating a new heap value
+            ResultPath: `${mapResultVariable}.arr[0]`,
+            // unique return name
+            Next: uniqueReturnName,
+          })
```
sam-goodwin, line 3050: Is this redundant? I've seen this code elsewhere, or at least similar code.

src/asl.ts
```diff
@@ -2837,60 +2792,494 @@ export class ASL {
...
+                    ResultPath: "$",
+                    Next: ASLGraph.DeferNext,
+                  }),
+                  ResultSelector: {
+                    "arr.$": "$",
+                    arrStr: "[null",
+                  },
+                  ResultPath: mapResultVariable,
+                  Next: "check",
+                }
```
sam-goodwin, line 3076: Again, copied code?

src/asl.ts
thantos: No. That map just zips up the index and the item. If you use index you get more state transitions.

src/asl.ts
thantos: wat?

src/asl.ts
thantos: Will document. 

Summary: in order to get the index, we zip the array items with their index. `[{ index: 0, item: item1], [index: 1, item : item2}, ...]`

When we zip, we need to reference `[].item` or `[].index` instead of just the array position. 

But we only do this when needed.

src/asl.ts
thantos: Uhh, yeah, map, filter, and forEach are similar. I decided to keep them separate to improve the readability, I'll see where I can merge them. 

Filter:

1. `init`
   1. copy array to tail
   2. initialize result string - map and filter, foreach doesn't have a result
2. `check` for an empty array - if so -> end
3. Setup `parameters`:
   1. item - evalDecl (catches binds)
   2. index - zip indices, evalDecl (bind isn't needed for a number, but just in case)
   3. array - evalDecl (catches bind)
4. run `body`
   3. return to heap value - map and filter
   4. return null - forEach
5. `tail`
   6. always - add body result to result string - map
   7. based on the predicate - add input item to the result string - filter
   8. tail array
6. go to `check`
7. end
   11. clear scratch space - forEach
   12. format result array - map and filter
   13. return result array - map and filter

src/asl.ts
thantos: and yeah, this is fragile, (thus the TODO item). any other ideas?

sam-goodwin: > wat?

I meant that it looks like it would fail if the argument was wrapped in parenthesized expression.

src/asl.ts
thantos: Alright, created a helper that abstracts most of the duplication away. Customization:
1. Whether to create a new variable for the callback output or write over the original item (in temp array)
2. how to handler the result of the callback and tail
3. what to do at the end (format array string vs nothing).

thantos:
uhh, fixing issue with `zipArray`, needs to use `this.aslGraphToStates` to make use of the globally unique state names.

sam-goodwin (approved):

PR #381, opened 2022-08-03
feat: remove PromiseExpr and PromiseArrayExpr

sam-goodwin:
Partial #374

This change removes `PromiseExpr` and `PromiseArrayExpr` from our tree as these nodes cannot be determined without type information.

BREAKING CHANGE: synth will no longer validate correct use of Promises - we rely purely on type-checking.


src/asl.ts
```diff
@@ -1569,32 +1563,7 @@ export class ASL {
...
-      if (
-        isArgument(expr.parent) &&
-        isCallExpr(expr.parent.parent) &&
-        isPromiseAll(expr.parent.parent)
-      ) {
-        return this.eval(expr.expr);
-      }
-      throw new SynthError(
-        ErrorCodes.Arrays_of_Integration_must_be_immediately_wrapped_in_Promise_all
-      );
```
sam-goodwin, line 1596: @thantos - i'm curious if there is another way to validate these, like using the AST representation of `Promise.all`?

thantos: Could we add tags to our AST and include `PROMISE` as a tag?

sam-goodwin: Wouldn't that require type information?

sam-goodwin: That would require type information so it doesn't solve the problem. We will just rely on static validation for now and not have it inside synth.

thantos (approved):

PR #382, opened 2022-08-03
feat: add isAsync and isAsterisk to function expressions and isOptional to ElementAccessExpr

sam-goodwin:
Partial #374

We were missing the following information from our tree:

```ts
async function foo() { }
function *foo() {}
async function *foo() {}

class Foo {
  async foo() { }
 *foo() {}
  async *foo() {}
}

object?.[element]
        // ^ 
```


src/asl.ts
```diff
@@ -4794,7 +4793,7 @@ function nodeToString(
     return `[${nodeToString(expr.expr)}]`;
   } else if (isElementAccessExpr(expr)) {
     return `${nodeToString(expr.expr)}[${nodeToString(expr.element)}]`;
-  } else if (isFunctionExpr(expr) || isArrowFunctionExpr(expr)) {
+  } else if (isFunctionLike(expr) || isArrowFunctionExpr(expr)) {
```
thantos, line 4796: Doesn't FunctionLike cover Arrow?

thantos: same above

thantos (approved):

src/asl.ts
sam-goodwin: Yes it does. My bad.

PR #386, opened 2022-08-05
feat: replace ts-transformer with SWC transformer

### LINKED ISSUE #374: Total AST decoration, closure serialization from FunctionlessAST and higher-order function capabilities

sam-goodwin:
Fixes #374 

- [x] depends on publishing the SWC WASM NPM package transformer to NPM https://github.com/functionless/swc-closure/pull/1/files
- [ ] should be synchronized with https://github.com/functionless/functionless-projen/pull/14
- [x] replace website/docs section about ts-patch with SWC 

This is change replaces our ts-transformer with a faster one implemented in Rust with [swc.rs](https://swc.rs)
1. It is significantly faster than the ts-transformer which makes running tests more delightful.
2. More importantly, it is so fast that we can run it in a require hook to decorate functions inside `node_modules` so reflection can be performed on arbitrary closures - not just those compiled by our transformer.
3. Finally, by running it in a require-hook instead of transforming the emitted JS, it has no impact on any other files being compiled, e.g. bundling a traditional Lambda handler `index.ts` file. Before, we would bloat that file which has negative impact on the bundle size, but now, the transformation is transient.

Below is our `.swcrc` JSON configuration file that the SWC plugin.

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "dynamicImport": false,
      "decorators": false,
      "hidden": {
        "jest": true
      }
    },
    "transform": null,
    "target": "es2021",
    "loose": false,
    "externalHelpers": false,
    "experimental": {
      "plugins": [["swc-closure", {}]]
    }
  },
  "minify": false,
  "sourceMaps": "inline",
  "module": {
    "type": "commonjs"
  }
}
```

To integrate into `jest`, we use the `@swc/jest` transformer and add the `@swc/register` import to configure the require-hook.
```ts
import type { Config } from "@jest/types";

// configure the @swc/register require hook to apply the transformer to all node_modules
import "@swc/register";

export default async (): Promise<Config.InitialOptions> => {
  return {
    ...config, // REDACTED
    transform: {
      "^.+\\.(t|j)sx?$": ["@swc/jest", {}],
    },
  };
};
```


.projenrc.js
```diff
@@ -112,7 +112,16 @@ const project = new CustomTypescriptProject({
      */
     // for serializer testing
     "uuid",
+    "@swc/cli",
+    "@swc/core",
+    "@swc/register",
+    "@swc/jest",
+    "swc-closure@file:../swc-closure",
```
sam-goodwin, line 119: TODO: consume the WASM from NPM

@thantos we need to decide on what to call the package, not sure how i feel about swc-closure

thantos: `swc-functionless`? isn't it bound to functionless AST right now?

sam-goodwin: Functionless ast is general though?

sam-goodwin: @functionless/swc is probably fine?

src/compile.ts
```diff
@@ -851,7 +861,14 @@ export function compile(
...
-      function ref(node: ts.Expression) {
+      function ref(node: ts.Identifier) {
+        const symbol = checker.getSymbolAtLocation(node);
+
+        const id = symbol?.valueDeclaration
+          ? getUniqueId(symbol.valueDeclaration)
+          : symbol?.declarations?.[0]
+          ? getUniqueId(symbol.declarations[0])
+          : 0;
         return newExpr(NodeKind.ReferenceExpr, [
```
thantos, line 872: I think all of the times it matters, this will have valueDeclaration?

src/function.ts
```diff
@@ -709,12 +709,12 @@ export class Function<
               `While serializing ${_resource.node.path}:\n\n${e.message}`
             );
           } else if (e instanceof Error) {
-            throw Error(
-              `While serializing ${_resource.node.path}:\n\n${e.message}`
-            );
+            throw e;
+            // throw Error(
+            //   `While serializing ${_resource.node.path}:\n\n${e.message}`
+            // );
```
thantos, line 715: delete?

thantos: I added this because it was very difficult to tell which function failed when the failure was async and there are many functions...

src/util.ts
```diff
@@ -209,7 +209,14 @@ export function isConstant(x: any): x is Constant {
...
   expr: Expr | QuasiString
 ): Constant | undefined => {
-  if (
+  if (isIdentifier(expr)) {
+    // const decl = expr.lookup();
+    // if (decl) {
+    //   if (isVariableDecl(decl) && decl.initializer) {
+    //     return evalToConstant(decl.initializer);
+    //   }
+    // }
```
thantos, line 218: delete?

thantos (approved):

.projenrc.js
sam-goodwin: Went with @functionless/ast-reflection

src/function.ts
sam-goodwin: This was a mistake, i was debugging

thantos:
Need to merge and see the errors?

thantos (approved):

thantos:
Also, update the website getting started and contributing markdown?

https://github.com/functionless/functionless/blob/main/website/docs/getting-started.md#L46
https://github.com/functionless/functionless/blob/main/CONTRIBUTING.md#L226

thantos (approved):

PR #387, opened 2022-08-05
feat(event-bridge): Add `trace-header` field to `Event` type

bakerlee:
Relevant aws-sdk-js commit: https://github.com/aws/aws-sdk-js/commit/438670cd3d6687c355e79d71ba3444f08a3cf5fc#diff-e39ababf70a50051cafcb815f0f7bb71fbb9f0ee616713dc2730885fbff9d7a7R1295-R1298

### DIFF
--- a/src/event-bridge/types.ts
+++ b/src/event-bridge/types.ts
@@ -13,6 +13,7 @@ export interface Event<
   time: string;
   region: string;
   resources: string[];
+  "trace-header": string;
 }
 
 /**
--- a/src/event-bridge/types.ts
+++ b/src/event-bridge/types.ts
@@ -13,7 +13,6 @@ export interface Event<
   time: string;
   region: string;
   resources: string[];
-  "trace-header": string;
 }
 
 /**
--- a/test/test-files/event-bus.ts
+++ b/test/test-files/event-bus.ts
@@ -11,31 +11,11 @@ const func = new Function<undefined, { source: string }>(
 );
 
 /**
- * Invalid integrations in transformers (minimal payload)
+ * Invalid integrations in transformers
  */
 bus.all().map(() => {
   // eslint-disable-next-line @typescript-eslint/no-floating-promises
   bus.putEvents({ "detail-type": "detail", detail: {}, source: "" });
-  return {};
-});
-
-/**
- * Invalid integrations in transformers (maximal payload)
- */
-bus.all().map(() => {
-  // eslint-disable-next-line @typescript-eslint/no-floating-promises
-  bus.putEvents({
-    "detail-type": "detail",
-    detail: {},
-    source: "",
-    version: "",
-    id: "",
-    account: "",
-    time: "",
-    region: "",
-    resources: [""],
-    "trace-header": "",
-  });
   return {};
 });
 
--- a/test/test-files/event-bus.ts
+++ b/test/test-files/event-bus.ts
@@ -11,11 +11,31 @@ const func = new Function<undefined, { source: string }>(
 );
 
 /**
- * Invalid integrations in transformers
+ * Invalid integrations in transformers (minimal payload)
  */
 bus.all().map(() => {
   // eslint-disable-next-line @typescript-eslint/no-floating-promises
   bus.putEvents({ "detail-type": "detail", detail: {}, source: "" });
+  return {};
+});
+
+/**
+ * Invalid integrations in transformers (maximal payload)
+ */
+bus.all().map(() => {
+  // eslint-disable-next-line @typescript-eslint/no-floating-promises
+  bus.putEvents({
+    "detail-type": "detail",
+    detail: {},
+    source: "",
+    version: "",
+    id: "",
+    account: "",
+    time: "",
+    region: "",
+    resources: [""],
+    "trace-header": "",
+  });
   return {};
 });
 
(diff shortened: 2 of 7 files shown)


thantos:
Thanks for the contribution!

src/event-bridge/types.ts
```diff
@@ -13,6 +13,7 @@ export interface Event<
   time: string;
   region: string;
   resources: string[];
+  "trace-header": string;
```
thantos, line 16: Do you know if trace-header can be used in the pattern match and input transform?

test/test-files/event-bus.ts
```diff
@@ -11,14 +11,34 @@ const func = new Function<undefined, { source: string }>(
...
+    id: "",
+    account: "",
+    time: "",
+    region: "",
+    resources: [""],
+    "trace-header": "",
+  });
+  return {};
+});
+
```
thantos, line 41: You don't need to add this here, this just tests our validations and your change doesn't change the validations.

However, could you add your field here:

https://github.com/functionless/functionless/blob/main/test/function.localstack.test.ts#L433

That will test the new field transform in lambda

and here: https://github.com/functionless/functionless/blob/main/test/step-function.test.ts#L818

To test the step function transform?

thantos (changes requested):

src/event-bridge/types.ts
bakerlee: It does not appear so. From the [docs](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-related-service-xray.html):

> ⚠️ Important
The trace header is not available on the event that's delivered to the invocation target.

I also tested it manually, using this event pattern:
```
{
  "trace-header": [{
    "exists": true
  }]
}
```

This caused all events to be ignored, which leads me to believe that `trace-header` is dropped before _any_ processing, not just upon delivery, per the docs.

bakerlee:
> Thanks for the contribution!

Thanks for the work you've done on the library, as well as the quick response and helpful suggestions!

test/test-files/event-bus.ts
bakerlee: Will do, commit incoming shortly.

sam-goodwin: Do we need to validate that trace header is not used in the event pattern?

bakerlee: That's a very good point. If not, it could cause a frustrating developer experience.

src/event-bridge/types.ts
thantos: In this case, I think we should just put the trace-header on the event that is used in the `putEvents()` integration method (and not the `Event` object used by Pattern and Input Transform).

https://github.com/functionless/functionless/blob/main/src/event-bridge/event-bus.ts#L562

```ts
export type PutEventInput<Evnt extends Event> = Partial<Evnt> &
  Pick<Evnt, "detail" | "source" | "detail-type"> & {
    "trace-header": string;
  };
```

test/test-files/event-bus.ts
thantos: See my comment above about which type to put it on. If we don't add it to `Event` and instead add it to `PutEventInput`, it shouldn't create confusion.

sam-goodwin (approved):

thantos (approved):

sam-goodwin:
Thanks for the fix! I'm curious, can you share any details on your use-case for functionless?

bakerlee:
> Thanks for the fix! I'm curious, can you share any details on your use-case for functionless?

@sam-goodwin Glad to!

I'm leading the effort to build a greenfield data processing PaaS. I'm pretty much an FP and event-sourcing zealot, so every operation is decoupled. Some of the categories of events we deal with are:

- User management operations (sales leads generated, new companies registered, subscriptions enabled/expired, users added/removed, permissions added/removed, etc.)
- Data processing operations (new datasets uploaded, automated processing events, user job submissions/events, etc.)
- Support operations (user requests, investigations, resolutions, admin overrides, etc.)

We have a skeleton crew, so I try to choose as many AWS managed services as possible (and of course, full CDK). For the event bus, as much as I love Kafka, I didn't want to risk the human overhead that I've seen, so EventBridge was a logical choice. (I also have no experience with MSK.) Similarly, as much as I love Scala, it's hard to hire for, and the learning curve can be steep. Plus, using Typescript for CDK and React (web and apps), it was a logical choice for backend code as well.

Despite religiously typing events, there was still way too much boilerplate written and time spent debugging issues caused by EventBridge's stringly-typed events. So I looked for a better way, and that's how I found Functionless!

sam-goodwin:
@bakerlee thanks for the response - super interesting. Those sound like great use-cases we would love to support with Functionless. We are working on some foundational pieces right now (like closure serializer and IAM Policy inferrence) which we are excited to use to build type-safe SQS, SNS, Kinesis and Firehose constructs with `map`, `flatMap`, `reduce`, `collect` etc. operations for functionally piping data around.

Straw-man experience would look something like the following
```ts
const bus = new EventBus<T>(stack, "bus");

const queue = bus.all().toSQSQueue(stack, "queue");

const firehose = queue.flatMap(event => tranform(event)).toKinesisFirehose({
  // s3
  // elastic search
  // redshift
  // snowflake??
  // etc.
  ..
})
```

Would love to connect and chat more if you're interested!

sam-goodwin:
One thing to note re: Kafka vs Event Bridge is ordering. Event bridge has no order guaranteees which are helpful for stream processing. 

On top of the SQS, SNS, Kinesis and Firehose abstractions, we could also explore a MSK abstraction. Type safe topics !

PR #394, opened 2022-08-06
fix(step-functions): shadow variable declarations should not erase

### LINKED ISSUE #369: Step Functions: shadow variable names overwrites state

thantos:
Fixes #369 

* [x] Shadow Variable Decl
* [x] Shadow Parameter Decl
* [x] Map/ForEach/Filter
* [x] $SFN
* [x] More tests
* Fixed a bug  in `.step()` which cause variable statements to be skipped when naming blocks by their first statement. 
* Random bugs found in `lexicalScope` relating to what should be visible after a scope like `For*Stmt`

Strategy:

1. Cache occurrence of variable name based on the uniqueness of the `Decl` in memory.
2. Replace variable names on declaration with a unique name.
3. Replace identifier on use with unique name.
4. Ensure all places use consist path for accessing variable names.


src/asl.ts
```diff
@@ -3345,6 +3362,51 @@ export class ASL {
     );
   }
 
+  /**
+   * In some cases, variable can be batch declared as state parameters.
+   * This method is an alternative to {@link evalDecl} which can evaluate multiple {@link ParameterDecl}s
+   * with initial values into a parameter object and any state to be run after (which setup any bound variables.)
+   *
+   * @returns a tuple tuple
```
sam-goodwin, line 3370: > a tuple tuple

not helpful

src/asl.ts
```diff
@@ -3345,6 +3362,51 @@ export class ASL {
...
+   *          1. a {@link Parameters} object intended to be used in the Parameters or ResultSelector of a state.
+   *             Contains the initialized parameter names. Could be empty.
+   *          2. a state or sub-state used to generate the bound names. May be undefined.
+   */
+  public evalParameterDeclForStateParameter(
+    node: FunctionlessNode,
+    ...parameters: [ParameterDecl | undefined, ASLGraph.JsonPath][]
```
sam-goodwin, line 3377: nit: add labels to the tuples to help with documentation?

```ts
...parameters: [parameter: ParameterDecl | undefined, path: ASLGraph.JsonPath]
```

src/asl.ts
```diff
@@ -3345,6 +3362,51 @@ export class ASL {
...
+              },
+              states,
+            ]
+          : [params, [...states, this.evalAssignment(decl.name, jsonPath)]];
+      },
+      [{}, []]
+    );
+
+    return [params, ASLGraph.joinSubStates(node, ...states)];
+  }
```
sam-goodwin, line 3408: Consider chunking this function up into a `map`/`flatMap` prior to the `reduce` - I find it hard to tease apart spreads from the evaluation logic.

src/node.ts
```diff
@@ -385,46 +387,58 @@ export abstract class BaseNode<
...
   public getLexicalScope(): Map<string, BindingDecl> {
-    return new Map(getLexicalScope(this as unknown as FunctionlessNode));
+    return new Map(
+      getLexicalScope(this as unknown as FunctionlessNode, "scope")
+    );
 
-    function getLexicalScope(node: FunctionlessNode | undefined): Binding[] {
+    function getLexicalScope(
+      node: FunctionlessNode | undefined,
+      kind: "scope" | "sibling"
```
sam-goodwin, line 396: Document this parameter, what is a `sibling`?

src/node.ts
```diff
@@ -385,46 +387,58 @@ export abstract class BaseNode<
...
-        return node.parameters.flatMap((param) =>
-          isIdentifier(param.name)
-            ? [[param.name.name, param]]
-            : getNames(param.name)
-        );
-      } else if (isForInStmt(node) || isForOfStmt(node)) {
-        return getNames(node.initializer);
-      } else if (isForStmt(node)) {
-        return getNames(node.initializer);
+        if (kind === "sibling") return [];
```
sam-goodwin, line 434: Why? I don't understand what `sibling` is modeling

src/step-function.ts
```diff
@@ -364,32 +362,20 @@ export namespace $SFN {
...
-              ? "$$.Map.Item.Index"
-              : arrayPath,
-          ],
-          states: isBindingPattern(param.name)
-            ? // if the param is a binding pattern, the value will be placed on the heap and then bound in the body
-              context.evalDecl(param, { jsonPath: `$.${paramName}` })
-            : undefined,
-        };
-      });
+      const [itemParam, indexParam, arrayParam] = callbackfn.parameters;
```
sam-goodwin, line 365: I was recently looking at this code, do we properly handle the case where all of these parameteres could be `undefined`? Their types are all `ParameterDecl` instead of `ParameterDecl | undefined` - which is wrong.

https://stackoverflow.com/questions/50647399/typescript-accessing-an-array-element-does-not-account-for-the-possibility-of-u

sam-goodwin (changes requested):

src/asl.ts
```diff
@@ -499,7 +500,7 @@ export class ASL {
     decl: FunctionLike
   ) {
     const self = this;
-    this.decl = visitEachChild(decl, function normalizeAST(node):
+    this.decl = decl = visitEachChild(decl, function normalizeAST(node):
```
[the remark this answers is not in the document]
thantos: Well clone breaks the value comparison I am doing in `getDeclarationName` and this constructor was using a mix of the post clone and pre-clone `decl`, so I update the original decl here to not make the mistake of using the wrong one. I don't know why we'd want two copies after updating the one.

src/asl.ts
```diff
@@ -525,27 +526,19 @@ export class ASL {
...
-        : undefined;
-
-    // only evaluate the context parameter if its a binding parameter.
-    // We'll inject the context variable if not deconstructed
-    const contextParameterBinding =
-      this.decl.parameters[1] && isBindingPattern(this.decl.parameters[1].name)
-        ? this.evalDecl(this.decl.parameters[1], { jsonPath: "$$" })
-        : undefined;
+    const [paramInitializer, paramStates] =
+      this.evalParameterDeclForStateParameter(
```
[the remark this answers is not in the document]
thantos:
```ts
{
    Type: "Pass",
    Parameters: {
          ...paramInitializer // < --- here
    }
}
```

Wasn't quite sure what to name this function, but I mean the `Parameters` field of a State (Pass, Task, Map, Parallel) object.

src/asl.ts
thantos: I think I can add another short pass to partition by `isIdentifier` now to clean this up.

src/node.ts
thantos: ahh, yeah, meant to take another look at this. Its kind of ugly.

src/node.ts
thantos: This code was returning names from scopes like function and for even after their scope was complete.

This is an attempt to inform the nodes types when to output their scoped names.

```
1 for(const a in []) { // scope
2  const b = ""; // sibling
3   for(const a in []){ ... } //sibling
4   a // node.getLexicalScope() -> { a: 1 }
5 }
```

I'll do another pass at this, but was happy to get it to behave as I intended.

src/step-function.ts
thantos: The functions that use them have been updated to handle undefined (these cases are well tested at least), but I'd like to try that flag and see how much of a mess it makes. Also don't like this implicit behavior by default.

```ts
  public evalParameterDeclForStateParameter(
    node: FunctionlessNode,
    ...parameters: [ParameterDecl | undefined, ASLGraph.JsonPath][]
```

sam-goodwin: > Also don't like this implicit behavior by default.

Totally agree with this - it's semantically wrong.

thantos: https://github.com/functionless/functionless/pull/398

src/asl.ts
```diff
@@ -594,6 +585,34 @@ export class ASL {
...
 
+  private getDeclarationName(binding: BindingDecl & { name: Identifier }) {
+    if (!this.variableNamesMap.get(binding)) {
+      const name = binding.name.name;
+      const nameCount = this.variableNamesCount.get(name);
+      this.variableNamesMap.set(
+        binding,
+        nameCount ? `${name}__${nameCount}` : name
+      );
+      this.variableNamesCount.set(name, (nameCount ?? 0) + 1);
```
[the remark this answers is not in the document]
thantos: nice call, it didn't consider the case that the name was already taken.

src/asl.ts
```diff
@@ -594,6 +585,34 @@ export class ASL {
...
+  }
+
+  private getIdentifierName(identifier: Identifier) {
+    const ref = identifier.lookup();
+
+    if (
+      (isBindingElem(ref) || isParameterDecl(ref) || isVariableDecl(ref)) &&
+      isIdentifier(ref.name)
+    ) {
+      return this.getDeclarationName(ref as any);
```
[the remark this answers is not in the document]
thantos: TS wouldn't narrow to `BindingDecl & { name: Identifier }`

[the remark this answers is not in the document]
thantos: updated to `as BindingDecl & { name: Identifier }`

src/asl.ts
thantos: The useful part is next. did not mean `tuple tuple` though. Updated.

image

src/node.ts
```diff
@@ -393,46 +395,75 @@ export abstract class BaseNode<
...
+     *             ```ts
+     *             for(const i in []) { // scope - emits i=self
+     *                const a = ""; // sibling - emits a=self
+     *                for(const a of []) {} // sibling emits []
+     *                a // requesting node
+     *             }
+     *             ```
+     *             some nodes only emit names to their `scope` (ex: for) and
+     *             other nodes emit names to all of their `sibling`s (ex: variableStmt)
+     */
```
sam-goodwin, line 415: Strange indentation

sam-goodwin (changes requested):

src/node.ts
thantos: hmm, I lined it up with the `the` after `kind`, but it works to line it up with the `kind` instead.

thantos: image

sam-goodwin: Maybe move the doc to the actual parameter declaration? That indentation hurts

sam-goodwin (approved):

PR #398, opened 2022-08-08
chore: enforce strict array index in code base 

thantos:


src/api.ts
```diff
@@ -738,15 +738,16 @@ export class APIGatewayVTL extends VTL {
       } else if (isReferenceExpr(expr.expr) || isThisExpr(expr.expr)) {
         const ref = expr.expr.ref();
         if (ref === Number) {
-          return this.exprToJson(expr.args[0]);
+          // Number() = 0
+          return expr.args[0] ? this.exprToJson(expr.args[0]) : "0";
```
sam-goodwin, line 742: why `"0"` instead of `0`?

src/api.ts
```diff
@@ -738,15 +738,16 @@ export class APIGatewayVTL extends VTL {
...
         } else {
           throw new SynthError(
             ErrorCodes.Unexpected_Error,
             "Called references are expected to be an integration."
           );
         }
       } else if (isIdentifier(expr.expr) && expr.expr.name === "Number") {
-        return this.exprToJson(expr.args[0]);
+        return expr.args[0] ? this.exprToJson(expr.args[0]) : "0";
```
sam-goodwin, line 750: Looks like we have two ways of dealing with Number()? Do we need both?

src/api.ts
```diff
@@ -761,11 +762,11 @@ export class APIGatewayVTL extends VTL {
             ref.parent.parameters.findIndex((param) => param === ref) === 0
           ) {
             // the first argument of the FunctionDecl is the `$input`, regardless of what it is named
-            if (expr.args.length === 0 || expr.args[0].expr === undefined) {
+            if (!expr.args[0] || expr.args[0].expr === undefined) {
```
sam-goodwin, line 765: Can you simplify to
```ts
if (expr.args[0]?.expr === undefined)
```

src/api.ts
```diff
@@ -761,11 +762,11 @@ export class APIGatewayVTL extends VTL {
...
               const key = this.newLocalVarName();
               return `{#foreach(${key} in $input.params().keySet())"${key}": "$input.params("${key}")"#if($foreach.hasNext),#end#end}`;
-            } else if (expr.args.length === 1) {
-              const argName = expr.args[0].expr!;
+            } else {
+              const argName = expr.args[0].expr;
```
sam-goodwin, line 769: Nice, did it narrow properly?

src/asl.ts
```diff
@@ -4058,7 +4058,7 @@ export namespace ASLGraph {
       .filter(ASLGraph.isStateOrSubState);
     return realStates.length === 0
       ? undefined
-      : realStates.length === 1
+      : realStates.length === 1 && realStates[0]
       ? { node, ...realStates[0] }
```
sam-goodwin, line 4062: I think it might be better in this case to do `...readStates[0]!` since our contract is such that the items in the array are never `undefiend`?

src/checker.ts
```diff
@@ -431,7 +431,7 @@ export function makeFunctionlessChecker(
         // explicit return false. We always want to check the parent of the declaration or signature.
         return false;
       }
-    } else if (symbol.declarations && symbol.declarations.length > 0) {
+    } else if (symbol.declarations && symbol.declarations[0]) {
```
sam-goodwin, line 434: `else if (symbol.declarations?.[0])`

src/event-bridge/utils.ts
```diff
@@ -184,7 +185,7 @@ export const flattenExpression = (expr: Expr, scope: EventScope): Expr => {
       throw Error("Object access must be a string.");
     } else if (isArrayLiteralExpr(parent)) {
       if (typeof key === "number") {
-        return parent.items[key];
+        return parent.items[key] ?? new UndefinedLiteralExpr();
```
sam-goodwin, line 188: should use `parent.fork(new UndefinedLiteralExpr())` or else the undefined literal will be an orphaned node.

src/vtl.ts
```diff
@@ -468,9 +469,10 @@ export abstract class VTL {
           const [previousValue, currentValue, currentIndex, array] =
             fn.parameters;
 
-          const currentValueName = isIdentifier(currentValue.name)
-            ? `$${currentValue.name.name}`
-            : this.newLocalVarName();
+          const currentValueName =
+            currentValue && isIdentifier(currentValue.name)
+              ? `$${currentValue.name.name}`
+              : this.newLocalVarName();
```
sam-goodwin, line 475: Could this be simplified as `isIdentifier(currentValue?.name)`

src/vtl.ts
```diff
@@ -495,7 +497,7 @@ export abstract class VTL {
             !!array
           );
 
-          if (isBindingPattern(currentValue.name)) {
+          if (currentValue && isBindingPattern(currentValue.name)) {
```
sam-goodwin, line 500: `isBindingPattern(currentValiue?.name)`

src/vtl.ts
```diff
@@ -686,7 +691,7 @@ export abstract class VTL {
     } else if (isVariableStmt(node)) {
       // variable statements with multiple declarations are turned into multiple variable statements.
       const decl = node.declList.decls[0];
-      return this.evalDecl(decl);
+      return decl ? this.evalDecl(decl) : undefined;
```
sam-goodwin, line 694: What is the result of returning `undefined`?

src/vtl.ts
```diff
@@ -1065,7 +1071,7 @@ export abstract class VTL {
         !!array
       );
 
-      if (isBindingPattern(value.name)) {
+      if (value && isBindingPattern(value.name)) {
```
sam-goodwin, line 1074: `value?.name`

sam-goodwin (changes requested):

src/api.ts
thantos: its returning a string? Above, returning a numeric literal looks like `number.toString(10)`

src/api.ts
thantos: I don't think so? Actually not sure which one, you change may have changed it back to not be a ref.

src/api.ts
thantos: it did!

src/api.ts
thantos: Apparently `Number` is still a ref. removed the second one.

src/vtl.ts
thantos: :/
image

src/vtl.ts
thantos: The contract says that `Stmt`s return void. A variable stmt shouldn't return a value.

src/vtl.ts
thantos: Oddly, even though the guard asserts a type, typescript doesn't narrow the value.

src/vtl.ts
sam-goodwin: Strange, it doesn't narrow it to defined?

sam-goodwin (approved):

src/vtl.ts
thantos: nope...

Issue #355, opened 2022-07-24
Improve Step Function State Names


thantos:
Currently step function state names are created by serializing Functionless AST into strings.

A state's name will be based first on the parent state and then either use the serialized AST of it's node or the local state name pre-pended onto the parent's state name.

This strategy was based on the previous design of step functions where node names had to be predictable. This is no longer the case. State names are computed locally and a mapping of local name to the state names are passed down a local state names are resolved.

```ts
try {
   task()
} catch {
   return null;
}
```

Here, task()'s first outputed state will be called `try` as that is the parent node.
Likewise, the return null will be called `catch`.

We should try to encode as much information as possible into the state names or comments to make the graphs readible.

thantos:
A few more ideas thrown out by @sam-goodwin and @mathisobadia.

### Using JS Labels

```ts
if(!auth) {
   AuthorizeUser:
      ....
}
```

Where we'd use the nearest label to decide the state name.

> Con: labels are not well known 

### Using comments

```ts
if(!auth) {
   // @fnls_state_name Authorize User
   ...
}
```

thantos:
### Using a Wrapper

At least one state generated by the body would have the given state name.

```ts
$SFN.withName("myName", () => {
   ....
});
```

tvanhens:
The current user experience is confusing because it is likely that multiple states will need to be generated for one statement within the closure. While I think allowing a user to label a statement would be useful, it still would be muddied by the extra states that need to be created in order to properly munge the data. Additionally, state transitions are expensive ($0.000025 per state transition) so this is not user-friendly from a cost perspective either.

The mapping between statements and step function states should be obvious and users should have an intuitive way of controlling the states that get created.

I would propose that, for the standard `StepFunction` construct (not including Express), we should create a secondary Express Step Function which shadows the top-level Standard Step Function. Any time we introduce extra states for the purposes of data munging and scope management, we should delegate those steps to this second step function such that the number of states is as close to 1:1 as is possible.

This would give the user a higher-degree of control over what states get created for visibility purposes in the top-level Step Function and save the user costly transitions (invoking a new Express Step Function costs $0.000001 or 1/25 the price of a transition).

Example 1:

```typescript
const x = {
   a, b, c
} // statement 1

func({ a, b, c }) // statement 2

func(func(), func(), func(func())) // statement 3
```

[...]

PR #399, opened 2022-08-08
feat(step-functions): improve support for binary operators

### LINKED ISSUE #393: Step Functions: Support `||` and `&&` in non-condition situations. Add Boolean coercion.
* `Boolean()` - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean/Boolean
* `||` - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR
   * Returns left if truthy, else right
   * `||` returns right when the left is falsy. `??` returns right only when the left is `undefined`.
* `&&` 
   * > More generally, the operator returns the value of the first [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) operand encountered when evaluating from left to right, or the value of the last operand if they are all [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy).
* Test
  * Operator Precedence - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR#operator_precedence

thantos:
Closes #393 

* [x] `&&` and `||` should return non boolean values. (see #393)
* [x] `Boolean()` coerce to boolean
* [x] `&&=`
* [x] `||=`
* [x] `??=`
* [x] Bug: `??` does not short circuit.
* [x] tests
* [x] Optimize chained choice states


src/asl.ts
```diff
@@ -430,19 +430,63 @@ export const isWaitState = (state: State): state is Wait => {
...
+   * Evaluates a single expression and returns the {@link ASLGraph.JsonPath} or {@link ASLGraph.LiteralValue}.
+   *
+   * Any require states will be merged in with the output.
+   *
+   * If the output was a {@link ASLGraph.ConditionOutput}, a new {@link Choice} state will turn the conditional into a boolean
+   * and return a {@link ASLGraph.JsonPath}.
+   *
+   * This method is the same as {@link ASL.evalExpr}, but can be used multiple times within a single {@link ASL.evalContext}.
+   */
+  evalExprToJsonPathOrLiteral: (
```
sam-goodwin, line 486: How are these two different?

src/asl.ts
```diff
@@ -978,62 +1035,92 @@ export class ASL {
...
+
+      const ifStates: Record<string, ASLGraph.SubState> = Object.fromEntries(
+        ifs.map((_if, i) => {
+          const stateName = i === 0 ? "if" : `if_${i}`;
+          const next =
+            i + 1 < ifs.length
+              ? `if_${i + 1}`
+              : els
+              ? "else"
+              : ASLGraph.DeferNext;
```
sam-goodwin, line 1060: Some descriptive comments would be helpful

src/asl.ts
```diff
@@ -978,62 +1035,92 @@ export class ASL {
...
+          const condition = this.eval(_if.when);
+          const conditionOutput = ASLGraph.getAslStateOutput(condition);
+          const stmtStates = this.evalStmt(_if.then, returnPass);
+          return [
+            stateName,
+            {
+              startState: "condition",
+              states: {
+                // run any states required for the condition and then evaluate the output (short circuit)
+                condition: ASLGraph.joinSubStates(_if.when, condition, {
```
sam-goodwin, line 1070: I can't grok `joinSubStates` - what do the first arguments mean? Are they other state graphs?

src/asl.ts
```diff
@@ -1446,6 +1525,100 @@ export class ASL {
     ) as any;
   }
 
+  /**
+   * Recursively evaluate a single expression, building a single {@link ASLGraph.NodeResults} object.
+   * All SubStates generated during evaluation will be merged into into a {@link ASLGraph.OutputSubState} along with the output
+   * of the handler callback.
+   *
+   * @param expr - Expression to evaluate.
+   * @param contextNode - Optional node to associate with the output state. This node may be used to name the resulting state.
```
sam-goodwin, line 1534: Does this doc comment properly apply to the overloads?

src/asl.ts
```diff
@@ -1446,6 +1525,100 @@ export class ASL {
...
+        ? normalizedState.output
+        : normalizedState;
+
+      return handler(newOutput, context);
+    });
+  }
+
+  /**
+   * Recursively evaluate a single expression, building a single {@link ASLGraph.NodeResults} object.
+   * All SubStates generated during evaluation will be merged into into a {@link ASLGraph.OutputSubState} along with the output
```
sam-goodwin, line 1576: nit:

> into into

src/asl.ts
```diff
@@ -1495,37 +1668,58 @@ export class ASL {
...
     return [handlerOutput, states];
   }
 
+  /**
+   * Returns the states required to normalize any {@link ASLGraph.Output} into a {@link ASLGraph.JsonPath}.
+   *
+   * * If the output was a {@link ASLGraph.LiteralValue}, a new state will be added that turns the literal into a json path.
+   * * If the output was a {@link ASLGraph.JsonPath}, the output is returned.
+   * * If the output was a {@link ASLGraph.ConditionOutput}, a new {@link Choice} state will turn the conditional into a boolean
+   *   and return a {@link ASLGraph.JsonPath}.
```
sam-goodwin, line 1684: Very clear - thankyou.

src/asl.ts
```diff
@@ -1495,37 +1668,58 @@ export class ASL {
...
+    | (ASLGraph.OutputSubState & { output: ASLGraph.JsonPath })
+    | ASLGraph.JsonPath
+    | ASLGraph.LiteralValue {
+    if (ASLGraph.isJsonPath(output) || ASLGraph.isLiteralValue(output)) {
+      return output;
+    } else {
+      return this.conditionState(node, output.condition);
+    }
+  }
+
```
sam-goodwin, line 1722: This seems to handle more than `ConditionToJsonPath`.

src/asl.ts
```diff
@@ -1591,25 +1779,52 @@ export class ASL {
...
-    };
-    const evalCondition = (expr: Expr) => {
-      const [cond, subState] = this.toCondition(expr);
-      subState && states.push(subState);
-      return cond;
+    const context: EvalContextContext = {
+      evalExpr: (expr: Expr, allowUndefined?: boolean) => {
+        const state = this.eval(expr, allowUndefined);
+        const output = ASLGraph.getAslStateOutput(state);
+        ASLGraph.isOutputStateOrSubState(state) && states.push(state);
```
sam-goodwin, line 1789: use `if` instead of `&&` for control-flow

src/asl.ts
```diff
@@ -1591,25 +1779,52 @@ export class ASL {
...
+        const output = ASLGraph.getAslStateOutput(state);
+        const normalizedStates = this.normalizeOutputToJsonPath(output, expr);
+        const normalizedOutput = ASLGraph.isOutputStateOrSubState(
+          normalizedStates
+        )
+          ? normalizedStates.output
+          : normalizedStates;
+        ASLGraph.isOutputStateOrSubState(state) && states.push(state);
+        ASLGraph.isOutputStateOrSubState(normalizedStates) &&
+          states.push(normalizedStates);
```
sam-goodwin, line 1806: use `if` instead of `&&` for control-flow

src/asl.ts
```diff
@@ -1591,25 +1779,52 @@ export class ASL {
...
+          expr
+        );
+        const normalizedOutput = ASLGraph.isOutputStateOrSubState(
+          normalizedStates
+        )
+          ? normalizedStates.output
+          : normalizedStates;
+        ASLGraph.isOutputStateOrSubState(state) && states.push(state);
+        ASLGraph.isOutputStateOrSubState(normalizedStates) &&
+          states.push(normalizedStates);
```
sam-goodwin, line 1823: use `if` instead of `&&` for control-flow

src/asl.ts
```diff
@@ -1807,31 +2019,48 @@ export class ASL {
...
 
+          return {
+            Type: "Pass",
+            Parameters: {
+              // intrinsic functions cannot be used in InputPath and some other json path locations.
+              // We compute the value and place it on the heap.
+              "string.$":
+                isPropAccessExpr(expr.expr) &&
+                isIdentifier(expr.expr.name) &&
+                expr.expr.name.name === "stringify"
```
sam-goodwin, line 2033: Maybe pull this out into a helper, `isJsonParse` and `isJsonStringify`. They're used earlier too and it's a bit verbose

src/asl.ts
```diff
@@ -2067,31 +2250,39 @@ export class ASL {
...
-            ),
-            output: left,
-          };
+          return this.assignValue(expr, right, left.jsonPath);
+        });
+      } else if (expr.op === "in") {
+        return this.evalContext(expr, ({ evalExpr }) => {
+          const left = evalExpr(expr.left);
+          const right = evalExpr(expr.right);
+          const elm = this.assertElementAccessConstant(left);
```
sam-goodwin, line 2272: nit: Does this return a different value than `left`? Should it be an `asserts` function instead?

src/asl.ts
```diff
@@ -2067,31 +2250,39 @@ export class ASL {
...
+          return ASLGraph.isLiteralValue(right)
+            ? right.value &&
+              typeof right.value === "object" &&
+              elm in right.value
+              ? { value: true, containsJsonPath: false }
+              : { value: false, containsJsonPath: false }
```
sam-goodwin, line 2278: You could simplify this to:
```ts
? {
  value: right.value && typeof right.value === "object" && elm in right.value,
  containsJsonPath: false
}
```

src/asl.ts
```diff
@@ -2129,33 +2319,152 @@ export class ASL {
...
-          return {
-            value: typeof exprOutput.value,
-            containsJsonPath: false,
-          };
-        }
-
-        const tempHeap = this.newHeapVariable();
+      } else if (expr.op === "&&" || expr.op === "||" || expr.op === "??") {
+        return this.evalExpr(expr.left, (leftOutput) => {
+          const right = this.eval(expr.right);
```
sam-goodwin, line 2329: So, if i undersetand correctly, eval(left) may reeturns a JsonPath, Condition or LiteralValue. If it returns a LiteralValue, does that mean no extra state was added to the machine?

src/asl.ts
```diff
@@ -2129,33 +2319,152 @@ export class ASL {
...
+            /**
+             * If left is a literal, evaluate the truthiness and return left or right.
+             * &&: when truthy, return right
+             * ||: when falsy, return right
+             * ??: when undefined or null, return right
+             */
+            if (expr.op === "&&") {
+              return !leftOutput.value ? leftOutput : right;
+            } else if (expr.op === "||") {
+              return !!leftOutput.value ? leftOutput : right;
```
sam-goodwin, line 2355: I find it a bit challengin to grok `!!`, any chance you can change this into a more explicit comparison?

src/asl.ts
```diff
@@ -2129,33 +2319,152 @@ export class ASL {
...
+              return {
+                condition:
+                  expr.op === "&&"
+                    ? ASL.and(leftOutput.condition, right.condition)
+                    : expr.op === "||"
+                    ? ASL.or(leftOutput.condition, right.condition)
+                    : // ??: a boolean cannot be undefined or null, return the left side.
+                      leftOutput.condition,
+              };
+            }
```
sam-goodwin, line 2377: Nice reduction!

src/asl.ts
```diff
@@ -2129,33 +2319,152 @@ export class ASL {
...
+          }
+
+          const condition =
+            /**
+             * Left is a condition
+             * ??: checks for null or undefined, a condition will always be true or false
```
sam-goodwin, line 2383: Why is this true?

src/asl.ts
```diff
@@ -2129,33 +2319,152 @@ export class ASL {
...
+          return this.conditionState(
+            expr,
+            condition,
+            left,
+            right,
+            left.jsonPath
+          );
+        });
+      }
+      assertNever(expr.op);
```
sam-goodwin, line 2447: nit: I prefer to do `return assertNever(expr.op)`. I think the dangling `assertNever` is awkward.

src/asl.ts
```diff
@@ -2129,33 +2319,152 @@ export class ASL {
...
+    } else if (isAwaitExpr(expr)) {
+      return this.eval(expr.expr);
+    } else if (isTypeOfExpr(expr)) {
+      return this.evalExpr(expr.expr, (exprOutput) => {
+        if (ASLGraph.isLiteralValue(exprOutput)) {
+          return {
+            value: typeof exprOutput.value,
+            containsJsonPath: false,
+          };
```
sam-goodwin, line 2456: aside: do we support `Array.isArray`?

src/asl.ts
```diff
@@ -2414,28 +2627,48 @@ export class ASL {
...
+   * @param trueState - states to execute when the condition is true. Default: returns true
+   * @param falseState - states to execute when the condition is false. Default: returns false
+   *
+   * @returns a {@link ASLGraph.OutputSubState} that executes true when the value is true and false when the value is false.
+   */
+  public conditionState(
+    node: FunctionlessNode | undefined,
+    cond: Condition,
+    trueState?: ASLGraph.NodeResults,
+    falseState?: ASLGraph.NodeResults,
```
sam-goodwin, line 2643: Why would these ever be `undefined`?

src/asl.ts
```diff
@@ -2605,152 +2841,136 @@ export class ASL {
...
+                    ASL.isPresent(`${arrayPath}[0]`),
+                    ASL.not(ASL.isPresent(resultVariable))
+                  ),
+                  Next: "initValue",
+                },
+                // not initialized, but the array is empty
+                {
+                  ...ASL.and(
+                    ASL.not(ASL.isPresent(`${arrayPath}[0]`)),
+                    ASL.not(ASL.isPresent(resultVariable))
```
sam-goodwin, line 2918: This means empty?

src/asl.ts
```diff
@@ -3020,72 +3235,69 @@ export class ASL {
...
+        true,
+        {
+          arrStr: "[null",
+        },
+        ({
+          tailJsonPath,
+          tailTarget,
+          iterationResult,
+          workingSpaceJsonPath,
+        }) => {
```
sam-goodwin, line 3262: I wonder if this would be easier to read if instead of passing many arguments, we passed a single object with properties for each of the arguments? That way meaning is given to them?

src/asl.ts
```diff
@@ -3735,8 +3949,40 @@ export class ASL {
     }
   }
 
+  private assignValue(
```
sam-goodwin, line 3952: docs

src/asl.ts
```diff
@@ -4739,40 +4789,138 @@ export namespace ASLGraph {
...
-        ...state,
-        Catch: state.Catch?.map((_catch) => ({
-          ..._catch,
-          Next: _catch.Next ? cb(_catch.Next) : _catch.Next,
-        })),
-        Next: state.Next ? cb(state.Next) : state.Next,
+  const joinChainedChoices = (
+    startState: string,
+    stateEntries: [string, State][]
+  ) => {
```
sam-goodwin, line 4826: Please use `function declarations` instead of `const arrow function`.

src/asl.ts
```diff
@@ -4871,13 +5016,331 @@ export namespace ASLGraph {
...
     };
   };
+
+  /**
+   * Compare any two {@link ASLGraph.Output} values.
+   */
+  export const compareOutputs = (
+    leftOutput: ASLGraph.Output,
+    rightOutput: ASLGraph.Output,
+    operator: ASL.ValueComparisonOperators | "!=" | "!=="
```
sam-goodwin, line 5026: I don't like these adhoc types. Why is `!=` and `!==` not a value comparison operator?

src/asl.ts
```diff
@@ -4871,13 +5016,331 @@ export namespace ASLGraph {
...
+      return ((operator === "==" || operator === "===") &&
+        leftOutput.value === rightOutput.value) ||
+        ((operator === "!=" || operator === "!==") &&
+          leftOutput.value !== rightOutput.value) ||
+        (leftOutput.value !== null &&
+          rightOutput.value !== null &&
+          ((operator === ">" && leftOutput.value > rightOutput.value) ||
+            (operator === "<" && leftOutput.value < rightOutput.value) ||
+            (operator === "<=" && leftOutput.value <= rightOutput.value) ||
+            (operator === ">=" && leftOutput.value >= rightOutput.value)))
```
sam-goodwin, line 5041: Should we introduce guards instead of these string literal comparisons to match the node guards and to also start developing composed helpers?

src/asl.ts
```diff
@@ -4871,13 +5016,331 @@ export namespace ASLGraph {
...
+    const op = leftOutput === left ? operator : invertBinaryOperator(operator);
+
+    return ASLGraph.compare(left, right, op as any);
+  };
+
+  export const compare = (
+    left: ASLGraph.JsonPath | ASLGraph.ConditionOutput,
+    right: ASLGraph.Output,
+    operator: ASL.ValueComparisonOperators | "!=" | "!=="
+  ): Condition => {
```
sam-goodwin, line 5066: no `const arrow function` please. It reads backwards.

sam-goodwin (changes requested):

src/asl.ts
```diff
@@ -430,19 +430,63 @@ export const isWaitState = (state: State): state is Wait => {
...
    * Callback provided to inject additional states into the graph.
    * The state will be joined (@see ASLGraph.joinSubStates ) with the previous and next states in the order received.
    */
   addState: (state: ASLGraph.NodeState | ASLGraph.SubState) => void;
+  /**
+   * Evaluates a single expression and returns the {@link ASLGraph.Output}.
+   *
+   * Any require states will be merged in with the output.
+   *
+   * This method is the same as {@link ASL.evalExpr}, but can be used multiple times within a single {@link ASL.evalContext}.
```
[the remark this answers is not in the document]
thantos: it can, but the outputs would need to be merged together.

```ts
return this.evalExpr(expr1, (output1) => {
   const states = this.evalExpr(expr2, (output2) => {
       return ... some states...
   })

   return ASLGraph.joinSubStates(expr,states, ... other states ... );
})
```

vs

```ts
return this.evalContext(expr, ({ evalExpr }) => {
   const output1 = evalExpr(expr1);
   const output2 = evalExpr(expr2);

   ... do something with the outputs ...
})
```

`evalExpr` is a simpler contract that evaluates a single expression, accumulates any states behind the scenes, gives the output to callback, and then merges the accumulated state with the state returned by the callback.

`evalContext` doesn't immediately evaluate an expression, but provides methods that can be used to evaluate any expression as needed. While in the context (callback) any states generated by the helper methods are accumulated and merged at the end with the states returned from the callback.

src/asl.ts
thantos: first one normalizes any output to a `JsonPath`, second one normalizes any output to a JsonPath or Literal value (aka, turns a condition into a JsonPath by adding a conditional state `heap = (cond) ? true : false`).

src/asl.ts
thantos: The first argument is just a node to associate with the result to give the result a name.

I don't like it, but also don't like the idea of needing to do 

```ts
{
   ... ASLGraph.joinSubStates(...),
   node
}
```

everywhere...

or having to remove the varArgs from the function to move `node` to the end of the arguments.

So far now it is

1. a node to associate with the merged state.
2. one or more states to merge together into a sequence.

src/asl.ts
thantos: Not the `contextNode` one... I guess I can duplicate the comments.

src/asl.ts
thantos: nice, thats been there for a least a month in the one it was copied from.

src/asl.ts
thantos: Updated `normalizeOutputToJsonPathOrLiteralValue`.

src/asl.ts
thantos: it does return a different value, changed to `getElementAccessConstant`

src/asl.ts
thantos: Not necessarily, `(await Task(), true)` would add some states and return a `LiteralValue`. We shouldn't make assumption about the states based on the `ASLGraph.Output` type. Always assume that `eval` can return states or just outputs.

However, if `ASLGraph.isAslGraphOutput(this.eval(expr))` is `true`, then you know no states were added. I use this fact here: https://github.com/functionless/functionless/blob/sfn_improve_binary_op/src/asl.ts#L2404.

`right` here isn't evaluated with `evalExpr` it uses `eval` which returns the entire node and output, so I know if `right` is a LiteralValue then there is no other work to do (and no short circuiting risks).

src/asl.ts
thantos: `!!` and `Boolean(...)` are the most succinct ways to handle truthy in JS. I wouldn't want to re-implement truthy logic (and we actually want truthy here).

src/asl.ts
thantos: `??` looks for null or undefined. `true` and `false` are always defined and non-null. Refactored a bit.

src/asl.ts
thantos: Nope, and neither does ASL... we can do isPresent on `$.arr[0]` (I assume this won't error), but a zero sized array would also fail. As far as I know, there are no properties on an array we could check for...

src/asl.ts
thantos: Defaults to return true and false constants, which was the original contract of this function.

src/asl.ts
thantos: yeah, haha. Or rather, it means that there is not a valid item at index 0. It could also mean the `arrayPath` doesn't exist. In out case we know the array path exists because we created it and we don't allow undefined.

src/asl.ts
thantos: Wow, a little formatting change wrapped a ton of unchanged code into this PR... the only part that changed here was the parameters in the arrow function and line 3252.

I also considered using an object instead. I liked that the arrow functions were in the order they were executed. I guess we could change the first 3-4 parameters to an object as they are not in order.

src/asl.ts
thantos: All of the functions in the namespaces where const arrow functions. I'll change them all to be functions.

src/asl.ts
thantos: OK, I just added them to the table and set all of the values to undnefined.

src/asl.ts
thantos: These are in the namespace and exported, so it really doesn't matter, but I did change all of the const arrow functions in the namesapces to be functions.

src/asl.ts
thantos: Not something that changed, I just moved them (new rule: anything that accepts ASLGraph.Output should be in ASLGraph instead of the ASL namespace. ASL are for raw ASL wrappers and output ASL, not ASLGraph).

Not sure if the guards would help too much, seems like they would get quite verbose to cover all operators.

src/asl.ts
sam-goodwin: Why not implement them?

src/asl.ts
sam-goodwin: Well, I would eventually want to change these to a numeric enum, but OK.

src/asl.ts
thantos: ASL doesn't have a StringNotEquals (the table is a map between operator and the ASL name (StringEquals, NumberLessThan, etc).

`!==` is `Not(*Equals)`.

src/asl.ts
thantos: I see. Would the guard accept a string operator or would you manually create a guard for all  44(+) operators?

sam-goodwin: All 44.

sam-goodwin: Later is fine :)

src/asl.ts
sam-goodwin: Why not implement them as functions? There are a bunch of functions referencing this map as the authoritative source of supported operators. It's difficult to follow.

src/asl.ts
sam-goodwin: Maybe rename the function `join` and take an object as input to help give meaning to the args?

src/asl.ts
sam-goodwin: What does !!` do?

src/asl.ts
sam-goodwin: Could we Stringify it as JSON and check that somehow? `isPresent(a[0]) || ToJson(a) === "[]"`

src/asl.ts
sam-goodwin: I think you can detect an empty array by checking the `ToJson(arr) === "[]"`

src/asl.ts
sam-goodwin: No big deal, i'm sure it reads better when you spend more time in the code base. Didn't realize this code was unchanged.

src/asl.ts
thantos: Was trying to generalize the handling of the matrix of operator, type, and path or constant based. I'll take a pass at this in another PR.

src/asl.ts
thantos: My plan was to rename this `sequence` or `chain` and get rid of `node` when figuring out better node naming.

sam-goodwin: Sounds good - can be done later.

src/asl.ts
thantos: `!` not truthy, `!!` is truthy.

> It is possible to use a couple of NOT operators in series to explicitly force the conversion of any value to the corresponding [boolean primitive](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#boolean_type). The conversion is based on the "[truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy)ness" or "[falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)ness" of the value (see truthy and falsy).

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_NOT#double_not_!!

> Do not use a Boolean object to convert a non-boolean value to a boolean value. To perform this task, instead, use Boolean as a function, or a [double NOT operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_NOT#double_not_!!):

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean

src/asl.ts
thantos: that may work. won't be pretty because we'll need a state to stingily and another to check the output.

Base case: is array of size 1 - 1 state (choice with condition)
Worst case: is not array or is empty - (choice: check for [1], pass: string format, choice: check for "[]")

We can optimize by checking for all other types in the first choice which would mean only objects would hit the worst case.

src/asl.ts
thantos: Hmm, good idea. Would take 2 states at least to check that, but it would work I think. (don't need it here, again can assume array)

src/asl.ts
sam-goodwin: That is unnecessary on this case

src/asl.ts
thantos: #355 ^

sam-goodwin (approved):

src/asl.ts
thantos: Yeah, I think you are right, the ternary should also enforce truthiness. I'll change it in the next PR.

PR #403, opened 2022-08-09
fix: enforce YieldExpr.expr is defined for yield*

sam-goodwin:
When `yield*` is used (i.e. when `delegate` is `true` in a `YieldExpr`), then the `expr` must be defined, otherwise it can be `undefined`.

```ts
yield* <expr> // valid
yield*; // invalid
```

### DIFF
--- a/src/expression.ts
+++ b/src/expression.ts
@@ -669,20 +669,29 @@ export class ImportKeyword extends BaseNode<NodeKind.ImportKeyword> {
   }
 }
 
-export class YieldExpr extends BaseExpr<NodeKind.YieldExpr> {
+export class YieldExpr<
+  Delegate extends boolean = boolean
+> extends BaseExpr<NodeKind.YieldExpr> {
   constructor(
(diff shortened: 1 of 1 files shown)


thantos (approved):

src/expression.ts
```diff
@@ -669,20 +669,29 @@ export class ImportKeyword extends BaseNode<NodeKind.ImportKeyword> {
...
+> extends BaseExpr<NodeKind.YieldExpr> {
   constructor(
     /**
      * The expression to yield (or delegate) to.
+     *
+     * If {@link delegate} is `true`, then {@link expr} must be defined, because
+     * `yield*` defers to another Generator, which `undefined` is not.
      */
-    readonly expr: Expr | undefined,
+    readonly expr: Expr | Delegate extends true ? undefined : never,
```
sam-goodwin, line 682: ughhh, that type is wrong lol, it should be `never : undefiend`

PR #404, opened 2022-08-09
feat: add Span information to F-AST

### LINKED ISSUE #392: Store span information in Functionless AST
## Keywords

span, source map, ast, location

## Description

When creating the FunctionlessAST, we lose the span information for each node and therefore don't know the location of a node. This information will be useful in many places:
1. when generating TypeScript we can generate a source map that points back to the original source code
2. add the source code location into SynthError so errors that can only be caught at runtime can be shown in context of the offending source code
3. place the information in to a Step Function's comment field to build a source-map like feature for ASL
4. generate a source map from VTL to TypeScript

To achieve this, we may want to consider changing flags such as `isAwait`, `isAsync` and other node modifiers to be a `Token` node type instead of primitives JS types so that we can associate. span with them.

sam-goodwin:
Fixes #392 

This change adds a Node's Span information to each node.

```ts
/**
 * A {@link Span} is a range of text within a source file, represented
 * by two numbers, a `start` and `end` character position.
 */
export type Span = [
  /**
   * Character position where this span starts.
   */
  start: number,
  /**
   * Character position where this span ends.
   */
  end: number
];
```

## Positional Convention

Because a span 2-tuple looks identical to a Node's s-expression `[kind: number, ...args]`, we adopt the convention of the span always being the element directly following the `kind`, making the definition of a s-expression now:

```ts
export type SExpr = [kind: number, span: Span, ...args: any[]]
```

### DIFF
--- a/.projenrc.js
+++ b/.projenrc.js
@@ -116,7 +116,7 @@ const project = new CustomTypescriptProject({
     "@swc/core",
     "@swc/register",
     "@swc/jest",
-    "@functionless/ast-reflection@file:../ast-reflection",
+    "@functionless/ast-reflection@^0.1.0",
   ],
   jestOptions: {
     jestConfig: {
--- a/.projenrc.js
+++ b/.projenrc.js
@@ -116,7 +116,7 @@ const project = new CustomTypescriptProject({
     "@swc/core",
     "@swc/register",
     "@swc/jest",
-    "@functionless/ast-reflection",
+    "@functionless/ast-reflection@file:../ast-reflection",
   ],
   jestOptions: {
     jestConfig: {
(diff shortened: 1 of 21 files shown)


.projenrc.js
```diff
@@ -116,7 +116,7 @@ const project = new CustomTypescriptProject({
     "@swc/core",
     "@swc/register",
     "@swc/jest",
-    "@functionless/ast-reflection",
+    "@functionless/ast-reflection@^0.1.0",
```
thantos, line 119: As we spoke about offline, lets make this a dependency to pin the version.

thantos (approved):

PR #406, opened 2022-08-09
fix: configure SWC and ast-reflection as dependencies

sam-goodwin:
SWC and ast-reflection versions are now decided by functionless


.projenrc.js
```diff
@@ -85,6 +85,11 @@ const project = new CustomTypescriptProject({
     "fs-extra",
     "minimatch",
     "@functionless/nodejs-closure-serializer",
+    "@functionless/ast-reflection@0.0.2",
+    "@swc/cli",
+    "@swc/core@1.2.218",
+    "@swc/register",
+    "@swc/jest",
```
sam-goodwin, line 92: I wonder if this should be up to the user to decide?

thantos: Should these be peers of the ast-reflection package and functionless? How dependent are we on the version of the `@swc` packages?

sam-goodwin:
Related to https://github.com/functionless/functionless-projen/pull/15

package.json
```diff
@@ -86,7 +81,12 @@
     "@aws-cdk/aws-appsync-alpha": "*"
   },
   "dependencies": {
+    "@functionless/ast-reflection": "0.0.2",
```
thantos, line 84: Should we start to use the second position for breaking changes and let this be `~0.0.2`?

.projenrc.js
sam-goodwin: `@swc/core` is fragile (hence the pinned version) as I've experienced ABI errors. We need to make sure ast-reflection is compiled against the same SWC core library veresions as @swc/core.

Good point on peer dependencies in ast-reflection.

Also, I think @swc/jest should not be in here. It's not guarnateed that everyone uses jest.

sam-goodwin: I added a peerDep on @swc/core to ast-reflection https://github.com/functionless/ast-reflection/pull/9. The only strict dependency for ast-reflection is `@swc/core` so I left out the others. Functionless can define the versions of the other dependencies.

thantos: yea, agree, only things that impact the consumer, `/jest` can be a devDep.

`@core` what happens if someone is trying to use SWC for other things?

package.json
sam-goodwin: Does `~` semantics apply to `0.x.y`?

.projenrc.js
sam-goodwin: They'll have to use a compatible version. We should just continue to build against latest.

thantos (approved):

PR #408, opened 2022-08-10
chore: migrate to TS .projenrc.ts

sam-goodwin:
Was getting annoyed with the lack of intellisense.


.projenrc.ts
```diff
@@ -85,7 +84,7 @@ const project = new CustomTypescriptProject({
     "fs-extra",
     "minimatch",
     "@functionless/nodejs-closure-serializer",
-    "@functionless/ast-reflection@^0.1.1",
+    "@functionless/ast-reflection@file:../ast-reflection",
```
thantos, line 87: mean to keep this?

thantos (approved):

.projenrc.ts
sam-goodwin: Nope, never do. Wonder if we can have an eslint rule lol

sam-goodwin:
Fixed, this change was a mess lo. Sorry about that.

thantos (approved):

PR #410, opened 2022-08-10
feat(ast): update ClassDecl.name and ClassExpr.name to be an Identifier

sam-goodwin:
For consistency, we will use `Identifier` instead of `string` for all names in the AST. This change updates ClassDecl and ClassExpr. They are not used so the change is not impactful on any existing code.

### DIFF
--- a/src/declaration.ts
+++ b/src/declaration.ts
@@ -59,7 +59,7 @@ export class ClassDecl<C extends AnyClass = AnyClass> extends BaseDecl<
   ) {
     super(NodeKind.ClassDecl, span, arguments);
     this.ensure(name, "name", [NodeKind.Identifier]);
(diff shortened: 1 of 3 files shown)


thantos (approved):

src/declaration.ts
```diff
@@ -58,7 +58,8 @@ export class ClassDecl<C extends AnyClass = AnyClass> extends BaseDecl<
     readonly filename?: string
   ) {
     super(NodeKind.ClassDecl, span, arguments);
-    this.ensure(name, "name", ["string"]);
+    this.ensure(name, "name", [NodeKind.Identifier]);
+    this.ensure(heritage, "name", ["undefined", NodeKind.Identifier]);
```
sam-goodwin, line 62: This is wrong, it should be ensured as an `Expr.

PR #413, opened 2022-08-11
fix(step-functions): assignment during literal can use the wrong values

thantos:
* [x] safe-variable heap assignment approach
* ~~in-order evaluation approach~~

### DIFF
--- a/src/asl.ts
+++ b/src/asl.ts
@@ -2185,57 +2185,82 @@ export class ASL {
       }
       assertNever(expr);
     } else if (isObjectLiteralExpr(expr)) {
-      return this.evalContext(expr, ({ evalExprToJsonPathOrLiteral }) => {
-        return expr.properties.reduce(
-          (obj: ASLGraph.LiteralValue, prop) => {
-            if (!isPropAssignExpr(prop)) {
-              throw new Error(
-                `${prop.kindName} is not supported in Amazon States Language`
-              );
-            }
-            if (
-              (isComputedPropertyNameExpr(prop.name) &&
-                isStringLiteralExpr(prop.name.expr)) ||
-              isIdentifier(prop.name) ||
-              isStringLiteralExpr(prop.name)
-            ) {
-              const name = isIdentifier(prop.name)
-                ? prop.name.name
-                : isStringLiteralExpr(prop.name)
-                ? prop.name.value
-                : isStringLiteralExpr(prop.name.expr)
-                ? prop.name.expr.value
-                : undefined;
-              if (!name) {
+      return this.evalContext(
+        expr,
+        ({ evalExprToJsonPathOrLiteral, addState }) => {
+          return expr.properties.reduce(
+            (obj: ASLGraph.LiteralValue, prop) => {
+              if (!isPropAssignExpr(prop)) {
+                throw new Error(
+                  `${prop.kindName} is not supported in Amazon States Language`
+                );
+              }
+              if (
+                (isComputedPropertyNameExpr(prop.name) &&
+                  isStringLiteralExpr(prop.name.expr)) ||
+                isIdentifier(prop.name) ||
+                isStringLiteralExpr(prop.name)
+              ) {
+                const name = isIdentifier(prop.name)
+                  ? prop.name.name
+                  : isStringLiteralExpr(prop.name)
+                  ? prop.name.value
+                  : isStringLiteralExpr(prop.name.expr)
+                  ? prop.name.expr.value
+                  : undefined;
+                if (!name) {
+                  throw new SynthError(
+                    ErrorCodes.StepFunctions_property_names_must_be_constant
+                  );
+                }
+
+                const valueOutput = evalExprToJsonPathOrLiteral(prop.expr);
+
+                // if the value is a json path, assign to a new heap value that will not change
+                // we do this to prevent the edge case where a variable reference changes
+                // between evaluation and assignment. we'll optimize this out later.
+                const assign = ASLGraph.isJsonPath(valueOutput)
+                  ? this.stateWithHeapOutput({
+                      ...ASLGraph.passWithInput(
+                        {
+                          Type: "Pass",
+                          Next: ASLGraph.DeferNext,
+                        },
+                        valueOutput
+                      ),
+                    })
+                  : undefined;
+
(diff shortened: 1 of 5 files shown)


src/asl.ts
```diff
@@ -1955,19 +1953,45 @@ export class ASL {
...
+             * instead of whatever json path they return. Do the same for conditions, they need have a boolean assigned
+             * to a heap variable anyways. We can handle jsonPath and conditions in the same way.
+             *
+             *
+             * ```ts
+             * let x = "1";
+             * ${x} ${(x = "2")}` // should output `1 2`
+             * ```
+             *
+             * If the heap assignment isn't necessary, we will optimize out the extra assignment later.
```
sam-goodwin, line 1977: Does this "later" code exist already? Can you link to where this will be done?

thantos: it does not exist yet, wanted to get everything lined up how I wanted it and then write the optimization. Doing to re-write to execute in order now and will write the optimization after.

thantos: https://github.com/functionless/functionless/issues/415

thantos:
Attempting another strategy where "dynamic" properties are assigned in order. 

However the optimization of assigning constant values is not as effective as I'd originally hoped. Using this strategy, any json path, conditional, or object literal which uses dynamic values could contain an incorrect value.

ex:
```ts
        let z = "";
        const g = {
          a: z,
          b: (z = "a"),
          c: ((z = "b"), z),
          z,
          t: z === "b",
          o: { z }
        };
```

`a: z` is now correct, being assigned the original value of `z`, but z, t and o are all wrong, being assigned in the initial object.

We could also treat any json path, condition, or literal value with embedded json path as dynamic values, but that would really limit the objects that benifit from this  strategy.

In the name of time time and correctness, will keep the aggressively assigned heap values and look to optimization (#415) to reduce the unnecessary states. 

code: 
```ts
} else if (isObjectLiteralExpr(expr)) {
      return this.evalContext(expr, ({ addState }) => {
        const evaluatedProperties = expr.properties.map((prop) => {
          if (!isPropAssignExpr(prop)) {
            throw new Error(
              `${prop.kindName} is not supported in Amazon States Language`
            );
          }
          if (
            (isComputedPropertyNameExpr(prop.name) &&
              isStringLiteralExpr(prop.name.expr)) ||
            isIdentifier(prop.name) ||
            isStringLiteralExpr(prop.name)
          ) {
            const name = isIdentifier(prop.name)
              ? prop.name.name
              : isStringLiteralExpr(prop.name)
              ? prop.name.value
              : isStringLiteralExpr(prop.name.expr)
              ? prop.name.expr.value
              : undefined;
            if (!name) {
              throw new SynthError(
                ErrorCodes.StepFunctions_property_names_must_be_constant
              );
            }

            return [name, this.eval(prop.expr)] as const;
          } else {
            throw new SynthError(
              ErrorCodes.StepFunctions_property_names_must_be_constant
            );
          }
        });

        const constantProperties = evaluatedProperties.filter(
          (entry): entry is [string, ASLGraph.Output] =>
            ASLGraph.isAslGraphOutput(entry[1])
        );

        const dynamicProperties = evaluatedProperties.filter(
          (
            entry
          ): entry is [
            string,
            ASLGraph.OutputState | ASLGraph.OutputSubState
          ] => !ASLGraph.isAslGraphOutput(entry[1])
        );

        // create an object out of the literal properties
        const constantPayload = constantProperties.reduce(
          (obj, [name, out]) => {
            // if this output is a Condition, need to normalize it.
            const updated = this.normalizeOutputToJsonPathOrLiteralValue(out);
            if (ASLGraph.isStateOrSubState(updated)) {
              addState(updated);
            }
            const updatedOut = ASLGraph.isStateOrSubState(updated)
              ? updated.output
              : updated;
            return {
              value: {
                ...(obj.value as Record<string, any>),
                ...ASLGraph.jsonAssignment(name, updatedOut),
              },
              containsJsonPath:
                obj.containsJsonPath ||
                ASLGraph.isJsonPath(updatedOut) ||
                updatedOut.containsJsonPath,
            };
          },
          { value: {}, containsJsonPath: false }
        );

        // if there are any elements that need to run additional states,
        // first assign the object literal with constant values (if there are any)
        // and then run each of the object assignments in order.
        // Finally, return the heap variable that contains the new object.
        if (dynamicProperties.length > 0) {
          // location where the object will be created
          const objOutput = this.newHeapVariable();

          const dynamicPropertiesStates = ASLGraph.joinSubStates(
            expr,
            constantProperties.length > 0
              ? this.assignValue(expr, constantPayload, objOutput)
              : undefined,
            // run each of the property assignment in order
            ...dynamicProperties.map(([name, states]) => {
              return ASLGraph.joinSubStates(
                states.node,
                states,
                this.assignValue(
                  states.node,
                  states.output,
                  `${objOutput}['${name}']`
                )
              )!;
            })
          )!;

          return {
            ...dynamicPropertiesStates,
            output: { jsonPath: objOutput },
          };
        }

        return constantPayload;
      });
    } 
```

src/asl.ts
sam-goodwin: Hmm, can you explain how inefficient the generated ASL will be if we merge this?

thantos: > Hmm, can you explain how inefficient the generated ASL will be if we merge this?

It will work correctly for all cases. 

Each dynamic assignment in an object, array, or template literal will create an extra state and variable.

(This pr is still draft so I can take a crack at the optimization).

sam-goodwin: How may states doe this require?

```ts
{
  constant: "value",
  dynamic: x
}
```

test/step-function.test.ts
```diff
@@ -883,7 +883,7 @@ test("put an event bus event", () => {
     {
       Parameters: { Entries: [{ EventBusName: bus.eventBusArn }] },
     },
-    "bus.putEvents"
+    "1__await bus.putEvents"
```
sam-goodwin, line 886: What changed here

sam-goodwin (approved):

test/step-function.test.ts
thantos: More states, needed to be more specific with the state names.

PR #414, opened 2022-08-12
feat(step-functions): support all AWS SDK operations.

### LINKED ISSUE #405: Support for typesafe SDK integrations
It would be great to have general SDK integrations (beyond just DynamoDB and EventBridge) similar to [ts2asl](https://github.com/Stedi/ts2asl/blob/main/examples/sdk-states.md).

Maybe something like

```ts
import { S3 } from "@aws-sdk/client-s3"; // use the v2 compatible to infer types

const workflow = new StepFunction(
  this,
  "Workflow",
  async (input: Input) => {
    await $AWS.SDK(S3).getObject({
      Bucket: input.bucket,
      Key: input.key
    });
  }
);
```

the [v2 compatible style](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/index.html#v2-compatible-style) has all functions directly on the class which would allow to infer the type

matthias-pichler-warrify:
Fixes #405


src/step-function.ts
```diff
@@ -466,6 +468,110 @@ export namespace $SFN {
...
+          return makeIntegration<"$AWS.SDK", (input: any) => Promise<any>>({
+            kind: "$AWS.SDK",
+            asl: (call, context) => {
+              const input = call.args[0]?.expr;
+              if (!isObjectLiteralExpr(input)) {
+                throw new SynthError(
+                  ErrorCodes.Expected_an_object_literal,
+                  `First argument ('input') into $AWS.SDK must be an object.`
+                );
+              }
```
sam-goodwin, line 509: This limitation is unnecessary. @thantos can confirm, but I believe we only needed that constraint before because we were extracting the Table from the ObjectLiteral. Step Functions can evaluate arbitrary inputs so it doesn't matter in the case when passing a string.

src/step-function.ts
```diff
@@ -466,6 +468,110 @@ export namespace $SFN {
...
+  } = new Proxy({} as any, {
+    get(_, serviceName: ServiceKeys) {
+      const client = new AWS[serviceName]();
+      const sdkIntegrationServiceName = mapSDKServiceName(serviceName);
+
+      return new Proxy(client as any, {
+        get: (_, prop: string) => {
+          return makeIntegration<"$AWS.SDK", (input: any) => Promise<any>>({
+            kind: "$AWS.SDK",
+            asl: (call, context) => {
```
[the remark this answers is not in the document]
matthias-pichler-warrify: I think that would be great 👍 Do you have a pointer for me on how to do this?

src/step-function.ts
```diff
@@ -466,6 +468,114 @@ export namespace $SFN {
...
+   * We define this to omit any non-funciton types from service
+   * instances. e.g. `apiVersions` and `config`
+   */
+  type ServiceInstance<T extends AWSServiceClass> = {
+    [K in FunctionKeys<InstanceType<T>>]: InstanceType<T>[K];
+  };
+
+  export const SDK: {
+    [serviceName in ServiceKeys]: ServiceInstance<typeof AWS[serviceName]>;
+  } = new Proxy({} as any, {
```
sam-goodwin, line 493: We should move to this beauty https://github.com/functionless/functionless/issues/405#issuecomment-1213036271

sam-goodwin (changes requested):

src/step-function.ts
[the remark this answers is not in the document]
sam-goodwin: Sorry I'm on phone. There is an interface where you can instantiate the client and store in the prewarm cache. 

```ts
native = {
      /**
       * Wire up permissions for this function to be called by the calling function
       */
      bind: (context: Function<any, any>) => {
        this.resource.grantInvoke(context.resource);
      },
      /**
       * Code that runs once per lambda invocation
       * The pre-warm client supports singleton invocation of clients (or other logic) across all integrations in the caller function.
       */
      preWarm: (preWarmContext: NativePreWarmContext) => {
        preWarmContext.getOrInit(PrewarmClients.LAMBDA);
      },
      /**
       * This method is called from the calling runtime lambda code (context) to invoke this lambda function.
       */
      // @ts-ignore - Typescript fails when comparing Promise<O> with ReturnType<ConditionalFunction<P, O>> though they should be the same.
      call: async (args, prewarmContext) => {
        const [payload] = args;
        const lambdaClient = prewarmContext.getOrInit<AWS.Lambda>(
          PrewarmClients.LAMBDA
        );
        const response = (
          await lambdaClient
            .invoke({
              FunctionName: functionName,
              ...(payload ? { Payload: JSON.stringify(payload) } : undefined),
            })
            .promise()
        ).Payload?.toString();
        return (response ? JSON.parse(response) : undefined) as Out;
      },
    };

```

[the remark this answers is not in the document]
matthias-pichler-warrify: Alright I added native support now (I hope 🙈 ) and also moved it to the `$AWS` namespace. I am very unsure whether I wired up the iam permissions correctly.

src/aws.ts
```diff
@@ -504,6 +507,237 @@ export namespace $AWS {
...
+                        new AWS[serviceName](
+                          props?.clientConfigRetriever?.(key)
+                        ),
+                    });
+
+                    return client[methodName](args[0]).promise();
+                  },
+                },
+                asl: (call, context) => {
+                  const iamOptions = call.args[1] as unknown as IamOptions;
```
thantos, line 573: This is actually the Argument, not an `IamOptions`, any value in typing this as `IamOptions` here?

src/aws.ts
```diff
@@ -504,6 +507,237 @@ export namespace $AWS {
...
+                      `Second argument ('iamOptions') into a SDK call must be an object`
+                    );
+                  }
+
+                  context.role.addToPrincipalPolicy(
+                    new aws_iam.PolicyStatement({
+                      effect: aws_iam.Effect.ALLOW,
+                      actions: iamOptions.iamActions ?? defaultIamActions,
+                      resources: iamOptions.iamResources,
+                      conditions: iamOptions.iamConditions,
```
thantos, line 587: if iamOptions is an `Argument`, getting a hold of the real value would look something like `call.args[1].expr.getProperty()...`

Also, the properties themselves could be any expression

```ts
new StepFunction(stack, 'sfn', async (input.actions) => {
    $AWS.SDK.CloudWatch.describeAlarms({}, { iamAction: input.actions })
})
```

=> 

`ObjectLiteral({ iamAction: PropAccess("actions", Identifier(input)) })`

You are probably looking to enforce that the entire object is a constant, in which you could do.

```ts
// toConstant returns undefined if the value cannot be converted to a compile/synth time constant.
const constantIamArgs = call.args[1]?.expr ? toConstant(call.args[1].expr) : undefined;

if(!constantIamArgs) {
   throw new SynthError(...);
} else if(typeof contantIamArg.constant !== "object") {
   // error
}
```

src/aws.ts
```diff
@@ -504,6 +507,237 @@ export namespace $AWS {
       },
     });
   }
+
+  export const SDK: StepFunctionSDK = new Proxy<any>(
```
thantos, line 511: nit: If we support lambda, `StepFunctionSDK` is probably not the best name.

src/aws.ts
```diff
@@ -504,6 +507,237 @@ export namespace $AWS {
...
+
+                  const sdkIntegrationServiceName =
+                    mapAslSdkServiceName(serviceName);
+                  const input = call.args[0]?.expr;
+
+                  if (!input) {
+                    throw (
+                      (new SynthError(ErrorCodes.Invalid_Input),
+                      "SDK integrations need a single input")
+                    );
```
thantos, line 599: This doesn't look right, I imagine you meant

```ts
throw new SynthError(ErrorCodes.Invalid_Input, "SDK integrations need a single input");
```

src/aws.ts
```diff
@@ -504,6 +507,237 @@ export namespace $AWS {
...
+                        "Expected an object literal as the first parameter."
+                      );
+                    }
+
+                    return context.stateWithHeapOutput({
+                      Type: "Task",
+                      Resource: `arn:aws:states:::aws-sdk:${sdkIntegrationServiceName}:${methodName}`,
+                      Parameters: output.value,
+                      Next: ASLGraph.DeferNext,
+                    });
```
thantos, line 619: if you want to support more than just a LiteralValue, you can do

```ts
// normalized any output to a jsonPath or literal
return context.evalExprToJsonPathOrLiteral(input, (output) => {
    return context.stateWithHeapOutput(
        // can add LiteralValue or JsonPath as the parameter to a task.
        ASLGraph.taskWithInput({
            Type: "Task",
            Resource: `arn:aws:states:::aws-sdk:${sdkIntegrationServiceName}:${methodName}`,
            Next: ASLGraph.DeferNext,
       }, output);
   });
);
```

src/aws.ts
```diff
@@ -504,6 +507,237 @@ export namespace $AWS {
...
+    ? (input: Input, iamOptions: IamOptions) => Promise<Output>
+    : never
+  : never;
+
+function mapAslSdkServiceName(serviceName: string): string {
+  // source: https://docs.aws.amazon.com/step-functions/latest/dg/supported-services-awssdk.html
+  switch (serviceName) {
+    case "Discovery":
+      return "applicationdiscovery";
+    // TODO: should I simply remove any trailing `Service` or leave the explicit mapping
```
thantos, line 709: +1 to explicit mapping. 

Also would be nice if the user could override this mapping as an escape catch

```ts
$AWS.SDK.SomeService.operation(input, { iam: ..., serviceNameOverride: "the right service name" });
```

src/aws.ts
thantos, line 511: What will it take to support the v3 clients? Could we use the same interface or should we make this SDKv2 to support SDKv3 in the future?

src/aws.ts
```diff
@@ -504,6 +507,237 @@ export namespace $AWS {
...
+   *
+   * @example
+   * ```
+   * {
+   *   "StringEquals" : { "aws:username" : "johndoe" }
+   * }
+   * ```
+   * @see https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition.html
+   */
+  iamConditions: Record<string, any>;
```
thantos, line 693: nit: is the `iam` prefix required if the object is already called `Iam*`?

src/aws.ts
```diff
@@ -504,6 +507,237 @@ export namespace $AWS {
...
+
+  export const SDK: StepFunctionSDK = new Proxy<any>(
+    {},
+    {
+      get(_, serviceName: ServiceKeys) {
+        return new Proxy<any>(
+          {},
+          {
+            get: (_, methodName: string) => {
+              const defaultServicePrefix = serviceName.toLowerCase();
```
thantos, line 519: wouldn't this require similar logic to the `mapAslSdkServiceName` method?

src/step-function.ts
thantos: Left a comment on the newest version, should be able to support any input object.

src/aws.ts
```diff
@@ -504,6 +507,237 @@ export namespace $AWS {
...
+              ];
+
+              return makeIntegration<
+                `$AWS.SDK.${ServiceKeys}`,
+                (input: any) => Promise<any>
+              >({
+                kind: `$AWS.SDK.${serviceName}`,
+                native: {
+                  bind(context, args) {
+                    const iamOptions = args[1] as unknown as IamOptions;
```
thantos, line 533: Same as below, `args[1]` will be an `Argument` with an expression for a value

src/aws.ts
```diff
@@ -504,6 +507,237 @@ export namespace $AWS {
...
+                  },
+                  preWarm(preWarmContext) {
+                    preWarmContext.getOrInit({
+                      key: `$AWS.SDK.${serviceName}`,
+                      init: (key, props) =>
+                        new AWS[serviceName](
+                          props?.clientConfigRetriever?.(key)
+                        ),
+                    });
+                  },
```
thantos, line 559: nice!

src/aws.ts
```diff
@@ -504,6 +507,237 @@ export namespace $AWS {
       },
     });
   }
+
+  export const SDK: StepFunctionSDK = new Proxy<any>(
+    {},
+    {
+      get(_, serviceName: ServiceKeys) {
+        return new Proxy<any>(
```
thantos, line 515: @sam-goodwin I am a little worried that the closure serializer will catch the proxy and fail. Need to test this at least.

thantos (changes requested):
Thank you for this change and getting into PR so quickly.

We'll need some transpile and runtime tests in the `step-function.test` (just runs the transpiler to ASL), `step-function.localstack`,  and `function.localstack` files. And because I am a bit worried about the proxy in the serialized lambdas functions, a test in `serialize.test`.

There are instructions on setting up localstack to run the localstack tests in the [`CONTRIBUTING.MD`](https://github.com/functionless/functionless/blob/main/CONTRIBUTING.md#testing)

src/aws.ts
sam-goodwin: Yeah I thought about that but it should only be serializing the returned data from the `get`. I think it'll work but of course need a test.

src/aws.ts
sam-goodwin: Yeah, I was wondering if we should update $AWS be this proxy and then have separate handlers for type safe ones?

src/aws.ts
thantos: It matters what ends up being serialized. Its pretty easy with the current serializer for it to try to serialize everything.

src/aws.ts
thantos: I'd be ok with that. for example putting them as static methods on their corresponding resource objects.

`Table.aws.putItem(...)`
`StepFunction.aws.startExecution()`

They would be more discoverable as well.

Then we'd have 3 forms:

1. Generalized API calls inferred from the SDK or generated using smithy (`$AWS.DynamoDb.putItem({ TableName: myTable.tableName, ... })`)
   1. Brute force access to the SDK for coverage
3. Special call that accepts a construct at the resource level, implemented per resource (`Table.aws.putItem()` or `Table.putItem({ Table: myTable, ... })`)
   1. gives full access to the SDK, most accurate at permission inference and specific integrations. created by hand (? for now?)
5. Specific resource which has invokable methods or is invokable (`myTable.putItem`)
   1. does not give full access to the SDK input or response, cleanest, created by hand

src/step-function.ts
```diff
@@ -536,6 +535,45 @@ export namespace $SFN {
   });
 }
 
+function mapSDKServiceName(serviceName: string): string {
+  // source: https://docs.aws.amazon.com/step-functions/latest/dg/supported-services-awssdk.html
+  switch (serviceName) {
+    case "Discovery":
+      return "applicationdiscovery";
```
sam-goodwin: Should we use a constant object literal instead of a switch?

src/aws.ts
matthias-pichler-warrify: @thantos good catch, I renamed it to just `SDK` now 👍 

If I understand you two correctly you are thinking about making `$AWS` what is currently `$AWS.SDK` and move all manually implemented calls somewhere else?

src/aws.ts
matthias-pichler-warrify: I wasn't sure how to get to the real options, thanks for the pointer to `evalToConstant` though 🙏

src/aws.ts
sam-goodwin: Yeah it might be the right approach - first go broad across all services, then build up from that with targeted type-safe (data aware) resource implementations. `$AWS` should map 1:1 with the aws-sdk as the name is inspired by this pattern:

```ts
import AWS from "aws-sdk"
```

+1 to @thantos's approach of moving the type-safe APIs to the Resource's Class as static methods, e.g. `Table.BatchGetItem( .. )`.

sam-goodwin: Should we do it as a part of this change, or merge it in with a different name and then follow-up with a migration?

src/aws.ts
matthias-pichler-warrify: Unfortunately I get `Property 'evalExprToJsonPathOrLiteral' does not exist on type 'ASL'.ts(2339)` ... what is the correct method here?

src/aws.ts
matthias-pichler-warrify: I would prefer getting this in first under `$AWS.SDK` and then move everything as I am not too sure I can do the migration with my limited knowledge of the project.

src/aws.ts
matthias-pichler-warrify: I renamed the parameter to `options` and the `IamOptions` type to `SdkCallOptions`. I wanted this second parameter to be the place where all additional stuff (e.g. your idea of the `serviceNameOverride`) would go. That is why I would leave the `iam*` prefix in the new `SdkCallOptions` type

package.json
```diff
@@ -89,7 +89,8 @@
     "@swc/register": "^0.1.10",
     "@types/aws-lambda": "^8.10.101",
     "fs-extra": "^10.1.0",
-    "minimatch": "^5.1.0"
+    "minimatch": "^5.1.0",
+    "utility-types": "^3.10.0"
```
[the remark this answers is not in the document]
matthias-pichler-warrify: good catch, not using it anymore.

src/step-function.ts
matthias-pichler-warrify: I was also thinking about that and given that we now also need a mapping for iam prefixes I would do a object literal 👍 

two questions: 
1. where should this be defined?
2. in a .ts or a .json file?

sam-goodwin: Good question, we use `.ts` for everything but I just added an imported `.json` file in another PR haha. JSON file seems to be more appropriate for pure data and we don't lose any type safety over the data. But .ts would be more consistent with our current code. What do you think?

src/aws.ts
sam-goodwin: Makes sense! I agree.

sam-goodwin: This is already an impactful change worth merging, we can easily make a breaking change later in a refactor. Thanks again for putting this together.

matthias-pichler-warrify: The good thing is that the sdk v2 and v3 are both currently supported and updated by AWS. v3 is a bit tricky as every client is split into a separate package. So getting all the types would mean installing all the packages (and some way to find them as new ones are added) and then combining all types from the packages into one.

Another thing that makes this hard is the fact that the new sdk uses the format

```ts
import { S3Client } from "@aws-sdk/client-s3";
const s3 = new S3Client({});
await s3.send(new GetObjectCommand({/*input here*/}));
```

it also has a v2 compatible style


```ts
import { S3 } from "@aws-sdk/client-s3";
const s3 = new S3({});
await s3.getObject({/*input here*/});
```

so it seems that we would need to go for the v2 compatible types anyway.

src/step-function.ts
matthias-pichler-warrify: I think I'll put it into the `aws.ts` file for now

src/aws.ts
thantos: Woops, forgot to mention they were new, haha. That change added full support for conditions anywhere in the source.

src/aws.ts
thantos: 100% agree.

src/aws.ts
matthias-pichler-warrify: I added an `IAM_SERVICE_PREFIX` mapping in `iam.ts` and kept the `.toLowerCase()` as a fallback. 👍  

The StepFunctions service name mapping only had a handful (~15) of exceptions to the `.toLowerCase()` "rule". IAM prefixes on the other hand are a lot messier. Out of over 310 service classes in the sdk only about 190 follow this rule. So currently I made the whole mapping explicit, but if you want to we can get rid of the 190 lines that follow this rule.

.gitpod.yml
```diff
@@ -0,0 +1,15 @@
+tasks:
```
matthias-pichler-warrify, line 1: Is it okay if I add a `.gitpod.yml` file to make the development in [gitpod.io](https://gitpod.io) (which I use) a bit more pleasant?

thantos: Yeah, looks good. Thanks for adding this.

test/step-function.localstack.test.ts
```diff
@@ -370,6 +370,40 @@ localstackTestSuite("sfnStack", (testResource, _stack, _app) => {
...
+      return new StepFunction<{}, string | undefined>(
+        parent,
+        "fn",
+        async () => {
+          const { MetricAlarms } = await $AWS.SDK.CloudWatch.describeAlarms(
+            {
+              AlarmNames: [alarm.alarmName],
+            },
+            { iamResources: [alarm.alarmArn] }
+          );
```
sam-goodwin, line 394: Maybe this is isn't important, but I find this to not be aesthetically pleasing - 2 object literals. Would it be too risky to move these parameters into the payload itself?

sam-goodwin: Or maybe:
```ts
$AWS.SDK.CloudWatch.withIAM({
  resources: [alarm.alarmArn]
}).describeAlarms({
  AlarmNames: [alarm.alarmName]
});
```

src/aws-sdk.ts
```diff
@@ -0,0 +1,98 @@
...
+  /**
+   * The resources for the IAM statement that will be added to the state machine
+   * role's policy to allow the state machine to make the API call.
+   *
+   * Use `["*"]` to grant access to all resources (discouraged).
+   *
+   * @example ["arn:aws:s3:::my_bucket"]
+   * @see https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_resource.html
+   */
+  iamResources: string[];
```
sam-goodwin, line 40: I wonder if we should re-use the `PolicyDocument` type from the CDK? Or `PolicyStatement`?

src/util.ts
```diff
@@ -408,3 +408,31 @@ export const invertBinaryOperator = (op: BinaryOp): BinaryOp => {
       return op;
   }
 };
+
+type _OverloadUnion<TOverload, TPartialOverload = unknown> = TPartialOverload &
```
sam-goodwin, line 412: Maybe link/credit where we found this type?

src/iam.ts
```diff
@@ -0,0 +1,318 @@
...
+  DAX: "dax",
+  Detective: "detective",
+  DeviceFarm: "devicefarm",
+  DevOpsGuru: "devops-guru",
+  DirectConnect: "directconnect",
+  DirectoryService: "ds",
+  Discovery: "discovery",
+  DLM: "dlm",
+  DMS: "dms",
+  DocDB: "rds", // source: https://docs.aws.amazon.com/documentdb/latest/developerguide/UsingWithRDS.IAM.ResourcePermissions.html
```
sam-goodwin, line 95: weird, haha

src/iam.ts
```diff
@@ -0,0 +1,318 @@
+import type { ServiceKeys } from "./aws-sdk";
+
+export const IAM_SERVICE_PREFIX: Record<ServiceKeys, string> = {
```
sam-goodwin, line 3: Great job enumerating these.

test/util.ts
```diff
@@ -288,6 +288,7 @@ export const normalizeCDKJson = (json: object) => {
   return JSON.parse(
     JSON.stringify(json)
       .replace(/\$\{Token\[[a-zA-Z0-9.-_]*\]\}/g, "__REPLACED_TOKEN")
-      .replace(/\"arn:[^\"]*\"/g, `"__REPLACED_ARN"`)
+      // do not replace arns that are for aws sdk service integrations
+      .replace(/\"arn:(?!aws:states:::aws-sdk)[^\"]*\"/g, `"__REPLACED_ARN"`)
```
thantos, line 292: nice!

src/aws.ts
```diff
@@ -504,6 +508,142 @@ export namespace $AWS {
...
+                        "Second argument ('options') into a SDK call must be an object"
+                      );
+                    }
+
+                    context.resource.addToRolePolicy(
+                      new aws_iam.PolicyStatement({
+                        effect: aws_iam.Effect.ALLOW,
+                        actions: options.iamActions ?? defaultIamActions,
+                        resources: options.iamResources,
+                        conditions: options.iamConditions,
```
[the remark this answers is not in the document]
thantos: This is going to grant the step function the ability to invoke any resource with this policy.

Ideally we'd be able to do what CDK does and create a statement for the resource being invoked.

https://github.com/aws/aws-cdk/blob/v1.168.0/packages/@aws-cdk/aws-stepfunctions-tasks/lib/lambda/invoke.ts#L121-L126

I don't think we should create such a permissive policy by default. We either need to require at least one resource be provided (`iamResources: [string, ...string[]]`), have the developer provide their own statement outside of the call scope, or create a way to extract the arn from the call (which would require the payload be an object literal and not a reference).

To @sam-goodwin 's comment below, I wonder if we should change the IAM part to a grant method. And decouple it from the call. We could error or warn when no policy is created.

```ts
// infer the permission from the api.
AWS.SDK.CloudWatch.describeAlarms.grant({ resources: [alarm.alarmName] });
/// explicit
AWS.SDK.CloudWatch.describeAlarms.grant({ actions: ..., resources: [alarm.alarmName] });

AWS.SDK.CloudWatch.describeAlarms({ AlarmNames: [alarm.alarmName] });
```

thantos (changes requested):

src/aws.ts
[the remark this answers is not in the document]
thantos: (cross post from discord)

Another issue I have with the approach of providing IAM permissions at request time is that the iam permissions exist outside of the scope of the call. 

All three of these blocks have the same output.

```ts
$AWS.SDK.CloudWatch.describeAlarms({
  AlarmNames: [alarm.alarmName]
}, {
  ..iam stuff
});
$AWS.SDK.CloudWatch.describeAlarms({
  AlarmNames: [alarm.alarmName]
});
```

```ts
$AWS.SDK.CloudWatch.describeAlarms({
  AlarmNames: [alarm.alarmName]
});
$AWS.SDK.CloudWatch.describeAlarms({
  AlarmNames: [alarm.alarmName]
}, {
  ..iam stuff
});
```

```ts
$AWS.SDK.CloudWatch.describeAlarms({
  AlarmNames: [alarm.alarmName]
});
$AWS.SDK.CloudWatch.describeAlarms({
  AlarmNames: [alarm.alarmName]
}, {
  ..iam stuff
});
$AWS.SDK.CloudWatch.describeAlarms({
  AlarmNames: [alarm.alarmName]
}, {
  ..iam stuff
});
$AWS.SDK.CloudWatch.describeAlarms({
  AlarmNames: [alarm.alarmName]
}, {
  ..iam stuff
});
$AWS.SDK.CloudWatch.describeAlarms({
  AlarmNames: [alarm.alarmName]
}, {
  ..iam stuff
});
$AWS.SDK.CloudWatch.describeAlarms({
  AlarmNames: [alarm.alarmName]
}, {
  ..iam stuff
});
$AWS.SDK.CloudWatch.describeAlarms({
  AlarmNames: [alarm.alarmName]
}, {
  ..iam stuff
});
```

and... if a user does make use of `resources` or `conditions`, any call to an api anywhere in the call stack will create a more permissive statement.

The condition below doesn't matter when myFunc is called. myFunc will grant the step function permission to call any resource with no condition.

```ts
 function myFunc() {
  $AWS.SDK.CloudWatch.describeAlarms({
     AlarmNames: [alarm.alarmName]
  });
}

StepFunction(stack, 'sfn', async () => {
      $AWS.SDK.CloudWatch.describeAlarms({
          AlarmNames: [alarm.alarmName]
      }, {
           ... some condition
      });
      myFunc();
});
```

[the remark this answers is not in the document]
sam-goodwin: Does this do anything on its own? Or does it return an object that represents that policy?

```ts
AWS.SDK.CloudWatch.describeAlarms.grant({ resources: [alarm.alarmName] });
```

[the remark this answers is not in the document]
thantos: Like the grant function in CDK. It adds the policy to the current step function.... Which I also don't like because it would imply positional, but would allow a higher order functions to assert permissions.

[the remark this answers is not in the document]
sam-goodwin: Maybe it helps to look at how this would be used to implement the methods on actual resources?

```ts
class Queue<T> {
  private readonly sqs;
  constructor(scope: Construct, id: string) {
    this.queue = new aws_sqs.Queue(this, "Queue");
    this.sqs = $AWS.SQS.grant(this.queue.grantRead)
  }

  public sendMessage(message: T): Promise<void> {
    await this.sqs.sendMessage({
      QueueArn: this.resource.arn,
      MessageBody: JSON.stringify(message)
    });
  }
}
```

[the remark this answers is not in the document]
thantos: ok, so this makes a scoped "resource" (`this.sqs`) that can grant any caller of `Queue.sendMessage`, the permission to call  sendMessage on `this.queue`.

What would this look like to give access to an array of queues? can it work with strings and not constructs? and do we need to understand all of the `grant*` methods in CDK?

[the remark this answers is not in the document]
thantos: What is the least we could do to ship this? Could we ship without any inferred IAM and only with either an explicit option or   using the CDK options. Both options with documentation.

src/aws-sdk.ts
matthias-pichler-warrify: The [PolicyStatement](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/aws-iam/lib/policy-statement.ts#L43) would be the correct class here. It would be passed to the call and we could check [`actions`](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/aws-iam/lib/policy-statement.ts#L539) field and add the default if not present 👍

matthias-pichler-warrify: @sam-goodwin would you like the following syntax better:

```ts
$AWS.SDK.CloudWatch.describeAlarms({
  AlarmNames: [alarm.alarmName]
}, new iam.PolicyStatement({resources: ["*"]}));
```

test/step-function.localstack.test.ts
matthias-pichler-warrify: I am definitely with you on this one. It just doesn't look good.

Messing with the sdk call input seems a bit risky as I wouldn't want to accidentally overwrite something in the input.

src/aws.ts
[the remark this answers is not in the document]
matthias-pichler-warrify: So to add my 2cents here:

What I found very helpful from the cdk stepfunctions modules is that they add iam permissions automatically. Across all the state machines we have, there were less than a handful where we had to add explicit permissions ourselves. I kind of wanted to go into a similar direction here as well. (Adding the IAM prefix map makes it even more seamless than the cdk stepfunctions module). The use of the `iamResources` came directly from the [`CallAwsService` task in the stepfunctions cdk](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_stepfunctions_tasks.CallAwsService.html#iamresources).

I do however also find the current approach unsatisfactory. But I guess what I would expect from a solution is the following:

1. infer as much as is reliably doable:
  That basically means the iam action as the resources are already infeasible to do (there are thousands) and most services don't use arns in their sdk methods but rather names so constructing the arn from the call is another difficult tasks.
2. no magic
  replacing the `AlarmName` with a `aws_cloudwatch.Alarm` class in the sdk call input and then inferring the required `grant...()` method from the sdk call is something that IMO should be reserved for the higher level integrations. As mentioned yesterday I think we should of this as a sort of L1 construct. To avoid duplicate IAM statements we could build a cache of added statements indexed by something like the object's hash. But I think that is an optimization that can be added later. Whether or not conditions come into effect or not is probably nothing we should solve as this goes where deep into IAM land and could lead to nasty edge cases.

3. make the requirement for iam explicit and allow full control if wanted
  Making the requirement for the iam resources explicit via a required type enforces users to actually specify them instead of experiencing opaque 403s during testing. An escape hatch for the inferred iam action is definitely needed and adding conditions allows even more control (that is something that the stepfunctions cdk is missing IMO)

With these things in mind I think what could make sense it one of the following:

## Adding a second parameter

```ts
$AWS.SDK.CloudWatch.describeAlarms({
  AlarmNames: [alarm.alarmName]
}, new iam.PolicyStatement({resources: [alarm.alarmArn]})); // or use the existing object literal but I like this better
```

## adding a chained call

```ts
$AWS.SDK.CloudWatch.withIAM({
  resources: [alarm.alarmArn]
}).describeAlarms({
  AlarmNames: [alarm.alarmName]
});

$AWS.SDK.CloudWatch.withPolicy(
  new iam.PolicyStatement({resources: [alarm.alarmArn]})
).describeAlarms({
  AlarmNames: [alarm.alarmName]
});

$AWS.SDK.CloudWatch.grant({
  resources: [alarm.alarmArn]
}).describeAlarms({
  AlarmNames: [alarm.alarmName]
});
```

## nesting the sdk parameters and other things all into the first argument (Similar to what the cdk does)

```ts
$AWS.SDK.CloudWatch.describeAlarms({
  params: {
    AlarmNames: [alarm.alarmName]
  },
  iamResources: [alarm.alarmArn],
});

// or use a PolicyStatement instead
$AWS.SDK.CloudWatch.describeAlarms({
  params: {
    AlarmNames: [alarm.alarmName]
  },
  policy:  new iam.PolicyStatement({resources: [alarm.alarmArn]}),
});
```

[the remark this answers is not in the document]
matthias-pichler-warrify: I think I like the last option the most

```ts
$AWS.SDK.CloudWatch.describeAlarms({
  params: {
    AlarmNames: [alarm.alarmName]
  },
  policy:  new iam.PolicyStatement({resources: [alarm.alarmArn]}),
});
```

test/__snapshots__/serialize.test.ts.snap
```diff
@@ -12503,6 +12503,7 @@ __fnls_AWS.Lambda = __fnls_AWS_Lambda;
 __f19.kind = \\"$AWS.EventBridge.putEvent\\";
 var __fnls_AWS_EventBridge = {putEvents: __f19};
 __fnls_AWS.EventBridge = __fnls_AWS_EventBridge;
+__fnls_AWS.SDK = undefined;
```
matthias-pichler-warrify, line 12506: I think me refactoring somehow broke the serialize tests. This used to be something like:

```
__fnls_AWS.SDK = {};
```

src/aws-sdk/index.ts
```diff
@@ -0,0 +1,241 @@
+import { aws_iam } from "aws-cdk-lib";
```
sam-goodwin, line 1: I'd prefer if we didn't have logic in any of the `index.ts` files and instead only use them to export

src/aws-sdk/index.ts
```diff
@@ -0,0 +1,241 @@
...
+            },
+            output
+          )
+        );
+      });
+    },
+  });
+}
+
+function policyStatementForSdkCall(
```
sam-goodwin, line 136: Let's add some docs.

src/aws-sdk/index.ts
```diff
@@ -0,0 +1,241 @@
...
+  const defaultMethod =
+    methodName.charAt(0).toUpperCase() + methodName.slice(1);
+  const defaultIamActions = [`${defaultServicePrefix}:${defaultMethod}`];
+
+  return new aws_iam.PolicyStatement({
+    effect: aws_iam.Effect.ALLOW,
+    actions: options.iamActions ?? defaultIamActions,
+    resources: options.iamResources,
+    conditions: options.iamConditions,
+  });
```
sam-goodwin, line 153: Should we add a `SID`? Anyway to avoid adding duplicate/redundant Policy Statements to a Policy Document?

src/aws-sdk/types.ts
```diff
@@ -0,0 +1,103 @@
...
+  /**
+   * The resources for the IAM statement that will be added to the state machine
+   * role's policy to allow the state machine to make the API call.
+   *
+   * Use `["*"]` to grant access to all resources (discouraged).
+   *
+   * @example ["arn:aws:s3:::my_bucket"]
+   * @see https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_resource.html
+   */
+  iamResources: [string, ...string[]];
```
sam-goodwin, line 45: I wonder if we could improve upon this by grouping all of the IAM details into a separate object?

Could the call end up looking like this?

```ts
$AWS.SDK.CloudWatch.describeAlarms({
  AlarmName: alarm.alarmName,
  iam: {
    resources: [alarm.alarmArn]
  }
});
```

By making it lower-case, we avoid any chance of collision without diverging from the AWS SDK's interface (we are strictly a super-set).

sam-goodwin: Or (potentially too fancy), use `$`.

```ts
$AWS.SDK.CloudWatch.describeAlarms({
  AlarmName: alarm.alarmName,
  $IAM: {
    resources: [alarm.alarmArn]
  }
});
```

sam-goodwin: Should IAM be an array by the way? Thinking of the case where an API needs access to different resources. For example, a S3 Bucket with KMS Encryption - you need access to both resources.

Also, are we losing anything by not using the grant methods? It should be possible to write your own IAM Policies and also use grant methods.

```ts
$AWS.SDK.CloudWatch.describeAlarm({
  AlarmName: alarm.alarmName,
  $IAM: [
    // option 1 - a callback
    (role) => alarm.grantDescribe(role),
    // option 2 - write a PolicyStatement
    {
       resources: [alarm.alarmArn],
       actions: [".."]
    }
  ]
});
```

src/aws-sdk/types.ts
```diff
@@ -0,0 +1,103 @@
...
+   *   "Type": "Task",
+   *   "Resource": "arn:aws:states:::aws-sdk:${sdkIntegrationServiceName}:${methodName}"
+   * }
+   * ```
+   *
+   * By default the service name is lowercased (e.g. `$AWS.SDK.CloudWatch.describeAlarms` => `"arn:aws:states:::aws-sdk:cloudwatch:describeAlarms"`).
+   * Some known special cases are also handled in {@link SDK_INTEGRATION_SERVICE_NAME}.
+   *
+   * To see which service name to use see the {@link https://docs.aws.amazon.com/step-functions/latest/dg/supported-services-awssdk.html#supported-services-awssdk-list | StepFunctions Developer Guide}
+   */
```
sam-goodwin, line 93: Great docs!

test/function.localstack.test.ts
```diff
@@ -430,6 +431,37 @@ localstackTestSuite("functionStack", (testResource, _stack, _app) => {
...
+      return new Function(
+        parent,
+        "function",
+        localstackClientConfig,
+        async () => {
+          const result = await $AWS.SDK.CloudWatch.describeAlarms({
+            params: {
+              AlarmNames: [alarm.alarmName],
+            },
+            iamResources: [alarm.alarmArn],
```
sam-goodwin, line 455: Hmm, how does it know which IAM Action to authorize against the Resource?

sam-goodwin (changes requested):

src/aws.ts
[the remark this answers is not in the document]
thantos: I am OK with following CDK for precedence, thanks for pointing that out and for such a detailed response.

I prefer option 1 with three changes:
1. don't use the policy statement, agree that re-use isn't helpful when the contract is different
2. Validate that a resource is provided or an explicit "*"
3. create an "options" object which contains more than just IAM in it for future configuration (ex: retry for sfn)
4. Require the iam object always be provided, even if that means duplication, we can relax this later or higher order functions will solve for this.

Because:
1. It separates the runtime and synth time configuration
2. It doesn't require a constant object literal be used for the runtime payload in order to extract the synth time values
3. users cannot shoot themselves in the foot by providing an empty resources array implicitly
5. the chained interface is a little cluncky too and doesn't add much value when the values are required.
6. We did something similar to option 3 for Sfn's invocation method and I don't like having to put the payload in an named object parameter every time.

src/aws-sdk/types.ts
thantos: While I understand the ergonomics of merging the payloads, mixing synth and runtime will force the payload to be a object literal only for ergonomic reasons.

This wouldn't be possible in the above

```ts
$AWS.SDK.CloudWatch.describeAlarms(input.describeAlarmPayload,
{
  iam: {
    resources: [alarm.alarmArn]
  }
});
```

thantos (changes requested):

src/aws-sdk/index.ts
```diff
@@ -0,0 +1,241 @@
...
+      },
+    });
+  }
+}
+
+function makeSdkIntegration(serviceName: ServiceKeys, methodName: string) {
+  return makeIntegration<
+    `$AWS.SDK.${ServiceKeys}`,
+    (input: any) => Promise<any>
+  >({
```
thantos, line 58: Note: you can also implement the `Integration` interface like `StepFunction`, `Function`, etc does if you'd prefer a class. Up to you.

The `makeIntegration` method is cleaner, but limited.

src/aws-sdk/index.ts
```diff
@@ -0,0 +1,241 @@
...
+    );
+  }
+
+  if (
+    !isPropAssignExpr(iamResources) ||
+    !isArrayLiteralExpr(iamResources.expr)
+  ) {
+    throw new SynthError(
+      ErrorCodes.Invalid_Input,
+      `Option 'iamResources' of a SDK call should be an array, found ${iamResources.kindName}`
```
thantos, line 194: nit: `array literal`

src/aws-sdk/index.ts
```diff
@@ -0,0 +1,241 @@
...
+    iamConditions &&
+    (!isPropAssignExpr(iamConditions) ||
+      // TODO: I don't think this is correct as the iamConditions most likely can be a variable as well, maybe a evalConstant here?
+      !isObjectLiteralExpr(iamConditions.expr))
+  ) {
+    throw new SynthError(
+      ErrorCodes.Expected_an_object_literal,
+      `Option 'iamConditions' of a SDK call should be an object, found ${iamConditions.kindName}`
+    );
+  }
```
thantos, line 229: For all of these, you should run `evalToConstant` and then fail when it returns undefined or an invalid runtime object type.

```ts
const iamResourcesValue = evalToConstant(iamResources.expr);
if(!iamResourcesValue || Array.isArray(iamResourcesValue) || iamResourcesValue.some(r => typeof r !== "string")) {
   ....
}
```

This will do a deep evaluation of the object to ensure all are literals or captured variables at synth time.

For example, this will work...

```ts
const arns = myConstructs.map(c => c.arn);
new StepFunction(this, 'sfn', sync () => {
   SDK.service.op(payload, { iamResources: arns });l
});
```

This works because `arns` will be a captured in a `ReferenceExpr` and during `evalToConstant` we get the captured value from the referenceExpr and return it (including applying properties/element access) to the object/arrays.

src/aws.ts
```diff
@@ -38,7 +41,7 @@ import {
   makeIntegration,
 } from "./integration";
 import { AnyTable, isTable, ITable } from "./table";
-import type { AnyAsyncFunction } from "./util";
+import { AnyAsyncFunction } from "./util";
```
thantos, line 44: why did this need to change?

test/__snapshots__/serialize.test.ts.snap
thantos: Based on the synth output in the new test below, I don't imagine the current state will actually work at runtime in lambda.  In theory undefined or void are fine until its actually used. 

1. Have you tried this in lambda, I see the tests are not actually running on the PR for whatever reason.
2. Is it the proxy or the importing that is breaking the serialization?
3. Is it different if you do something like 
```ts
const cloudWatchClient = $AWS.SDK.CloudWatch;
new Function(..., async () => {
    cloudWatchClient.describeAlarms(...)
})
```

thantos (changes requested):

.projenrc.ts
```diff
@@ -101,6 +101,7 @@ const project = new CustomTypescriptProject({
     "graphql-request",
     "ts-node",
     "ts-patch",
+    "flatted",
```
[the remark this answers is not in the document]
matthias-pichler-warrify: Since I wanted to thes the automatically generated role policy for step functions I had to update the `normalizeCDKJson` util to handle circular objects: https://github.com/functionless/functionless/pull/414/files#diff-8951d3f508f54ec05c5df49f04f821a2aec5c75bce858639461f388e498a398cR11

src/aws-sdk/index.ts
sam-goodwin: Why? This should never be evaluated during synth time right? The IAM resources shouldn't create states.

src/aws-sdk/types.ts
sam-goodwin: Ok fair enough. I also saw the other comment about how these are L1s and grant methods belong in the L2. Fair enough.

test/function.localstack.test.ts
matthias-pichler-warrify: That's what I use the iam service name mapping for: I get the service prefix from there and then just use the method name. E.g. 

- `$AWS.SDK.CloudWatch.describeAlarms => "cloudwatch:DescribeAlarms"`
- `$AWS.SDK.StepFunctions.describeExecution => "states:DescribeExecution"`

src/aws-sdk/index.ts
matthias-pichler-warrify: I am not sure adding a `sid` actually does any deduplication in the cdk

src/aws.ts
matthias-pichler-warrify: when the proxy still used to live in this file I had to use more utils here ... changed it back now 👍

matthias-pichler-warrify:
Somehow I messed up the serialization (or the AWS SDK did). I get the following error in the tests:

```
  ● recursion

    While serializing functionStack/parent16/function:

    Error serializing '()=>{ integrationPrewarms.forEach(re ...'

    '()=>{ integrationPrewarms.forEach(re ...': captured
      variable 'integrationPrewarms' which indirectly referenced
        function 'preWarm': which captured
          variable '_awsSdk' which indirectly referenced
            function 'extractData': which could not be serialized because
              Unexpected missing variable in closure environment: payload

    Function code:
      function extractData(resp) {
        Rest.extractData(resp);
      
        var req = resp.request;
        var operation = req.service.api.operations[req.operation];
```

test/__snapshots__/serialize.test.ts.snap
matthias-pichler-warrify: I tried adding a test where I use a referenced method outside the 

```ts
const {describeAlarms} = $AWS.SDK.CloudWatch;

serialize( () => {describeAlarms(...)})
```

And it said that either SDK or CloudWatch is undefined.

thantos:
> Somehow I messed up the serialization (or the AWS SDK did). I get the following error in the tests:
> 
> ```
>   ● recursion
> 
>     While serializing functionStack/parent16/function:
> 
>     Error serializing '()=>{ integrationPrewarms.forEach(re ...'
> 
>     '()=>{ integrationPrewarms.forEach(re ...': captured
>       variable 'integrationPrewarms' which indirectly referenced
>         function 'preWarm': which captured
>           variable '_awsSdk' which indirectly referenced
>             function 'extractData': which could not be serialized because
>               Unexpected missing variable in closure environment: payload
> 
>     Function code:
>       function extractData(resp) {
>         Rest.extractData(resp);
>       
>         var req = resp.request;
>         var operation = req.service.api.operations[req.operation];
> ```

I have an idea how to fix this and the case where serialization is undefined, let me see what I can do and get back... the closure serializer is... a pain in the ass (@sam-goodwin  is working on a new, experimental version #402)

thantos:
Alright, I've been playing around with this PR. 

Proxies will not work in the Lambda Closure Serializer for a few reasons, but the main reason is that proxies are not serializable (right now) and `$AWS.SDK.CloudWatch.describeAlarms` make it to SDK and returns `{}`.

 this works for me though

```ts
const da = $AWS.SDK.CloudWatch.describeAlarms;

new function(... async () => { da(...) })
```

Because the proxies are resolved before serialization.

----

Additionally, when using `AWS[serviceName]` in the `call` and `preWarm` function, the entire AWS SDK is serialized. Using `require('aws-sdk`)[serviceName]` resolves this.

matthias-pichler-warrify:
> Alright, I've been playing around with this PR.
> 
> Proxies will not work in the Lambda Closure Serializer for a few reasons, but the main reason is that proxies are not serializable (right now) and `$AWS.SDK.CloudWatch.describeAlarms` make it to SDK and returns `{}`.
> 
> this works for me though
> 
> ```ts
> const da = $AWS.SDK.CloudWatch.describeAlarms;
> 
> new function(... async () => { da(...) })
> ```
> 
> Because the proxies are resolved before serialization.
> 
> Additionally, when using `AWS[serviceName]` in the `call` and `preWarm` function, the entire AWS SDK is serialized. Using `require('aws-sdk`)[serviceName]` resolves this.

Thanks, I changed the `require("aws-sdk")` 👍 

Thanks for the investigation. What does this mean for this PR though? Should Lambda support be removed until the serializer supports it? Or will the new serializer be available in a next release?

thantos:
Are you seeing any issues with typescript from the types in this change? My IDE is bogged down and typescript keeps crashing on this change.

sam-goodwin:
We may have to move to a code generation approach. Use the ts compiler to parse the AWS sdk and generate the interface instead of the proxy

sam-goodwin:
If the types aren't a problem I agree with dropping lambda support to unblock this

src/aws-sdk/types.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export type SDK = {
+  [serviceName in ServiceKeys]: typeof AWS[serviceName] extends new (
+    ...args: any[]
+  ) => infer Client
+    ? {
+        [methodName in keyof Client]: SdkMethod<Client[methodName]>;
+      }
+    : never;
+};
```
sam-goodwin, line 28: @thantos we should try collapsing this to an interface. This may help with redundant type computation

sam-goodwin:
```
type Mixin<T> = T

interface SDK extends Mixin<{
  [serviceName in keyof typeof AWS]: ...
}
```

sam-goodwin: Hmm this might not help because it's methods that have the problem and they will still be lazily bound

thantos:
No idea if this is better, but what I ended up with

```ts
type AWSServiceClient<T> = T extends new () => infer Client ? Client : never;
type TAWS = typeof AWS;
export type Mixin<T> = T;

/**
 * First we have to extract the names of all Services in the v2 AWS namespace
 *
 * @returns "AccessAnalyzer" | "Account" | ... | "XRay"
 */
export type ServiceKeys = {
  [K in keyof TAWS]: AWSServiceClient<TAWS[K]> extends never ? never : K;
}[Exclude<keyof TAWS, "Service">];

export interface SDK
  extends Mixin<{
    [serviceName in ServiceKeys]: SDKClient<
      AWSServiceClient<TAWS[serviceName]>
    >;
  }> {}

export type SDKClient<Client> = {
  [methodName in keyof Client]: SdkMethod<Client[methodName]>;
};
```

thantos:
Another attempt which only computes the clients once.

```ts
type AWSServiceClient<T> = T extends new () => infer Client
  ? Client extends AWSService
    ? Client
    : never
  : never;
type TAWS = typeof AWS;
export type Mixin<T> = T;

interface SDKClients
  extends Mixin<{
    [K in Exclude<keyof TAWS, "Service">]: AWSServiceClient<TAWS[K]>;
  }> {}

/**
 * First we have to extract the names of all Services in the v2 AWS namespace
 *
 * @returns "AccessAnalyzer" | "Account" | ... | "XRay"
 */
export type ServiceKeys = {
  [K in keyof SDKClients]: SDKClients[K] extends never ? never : K;
}[keyof SDKClients];

export interface SDK
  extends Mixin<{
    [serviceName in ServiceKeys]: SDKClient<SDKClients[serviceName]>;
  }> {}

export type SDKClient<Client extends AWSService> = {
  [methodName in keyof Client]: SdkMethod<Client[methodName]>;
};
```

thantos:
https://github.com/functionless/functionless/tree/sussman/sdk

Here is a working branch (lambda tests still don't work for some reason):

1. Changed the types, which seem "better" but hard to say
2. Replace the snapshot tests on the role with the cdk asserts library
3. added "lazy integrations" which are resolved on discovery to reduce impact of creating an integration for every API and Method.
4. Replaced the SDK proxy with a map of server to method to lazy integrations. Proxies do not work in Functionless's Lambda's closure serializer (yet).

Exiting problems:
1. When the full `$AWS.SDK...` is used, the Lambda bundle is huge (~4.5MB) as it bundles in all of the service/method pairs and their lazy functions. 
2. types, maybe
3. lambda localstack tests are still not passing 
4. Synthesis takes a long time (minutes) to 1) build the SDK map at runtime. 2) bundle the large file.

I think the path forward is:
1. Ensure the types are not detrimental to the system or generate them statically
2. Ship SFN support only
3. focus on proxy support in the new closure serializer

sam-goodwin:
I think we should drop Lambda support for now so we can get this change in for Step Functions - that is already super useful and was the original reason behind the change.

Have we tested the types? If we're worried about that, I suggest we write a quick script to code generate the interface. My current thought is to use the ts compiler to parse all the type declarations in aws-sdk and spit out an `interface` with all the types. This + a Proxy solution will be ideal IMO because how lightweight the late-bound Proxy is over generated methods.

I've updated ast-reflection (https://github.com/functionless/ast-reflection/pull/22) to intercept Proxies so we can follow up this change with our new closure serializer that will have support for serializing Proxies.

The new closure serializer can also have bespoke heuristics implemented for cases we care deeply about, such as the `$AWS.SDK`.

sam-goodwin:
Working on the script to generate the types. PR is in draft as I haven't yet generated the interface, but it does successfully discover all of the API methods and request/response payloads. See: https://github.com/functionless/functionless/pull/430

matthias-pichler-warrify:
Alright I removed the `native` block from the integration to drop lambda support, updated the types to what @thantos suggested and also used his new snapshot tests for the step-functions 👍 

Is there anything missing to ship this?

sam-goodwin (approved):

sam-goodwin:
> Alright I removed the `native` block from the integration to drop lambda support, updated the types to what @thantos suggested and also used his new snapshot tests for the step-functions 👍 
> 
> 
> 
> Is there anything missing to ship this?

Only thing missing is mentioning this in website/docs

sam-goodwin:
@thantos @matthias-pichler-warrify  - https://github.com/functionless/functionless/pull/430 now successfully generates an interface for the entire AWS SDK in the form we need for this. So, if we conclude that the mapped type is too expensive and creates a terrible experience, we can simply use this generated form. Check it out and lemme know what you think.

sam-goodwin:
https://raw.githubusercontent.com/functionless/functionless/16ee5dafeacc1ac46161cdd9811558777cfc6cfd/src/sdk.generated.ts

sam-goodwin:
> https://raw.githubusercontent.com/functionless/functionless/78d18ba7531fd96250b2f62fd9ddc33350d8cc80/src/sdk.generated.ts

The sheer size of this file puts into perspective why the mapped type was choking. It might not be the recursive nature, it might just be the mass. @thantos since your change, are you still experiencing the slow down?

thantos:
Project before SDK

```
        Command being timed: "npx tsc"
        User time (seconds): 49.83
        Percent of CPU this job got: 163%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:31.77
        Maximum resident set size (kbytes): 2334792
```

Head of this PR

```
        Command being timed: "npx tsc"
        User time (seconds): 53.25
        Percent of CPU this job got: 160%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:34.56
        Maximum resident set size (kbytes): 2347792
```

Before my change

```
        Command being timed: "npx tsc"
        User time (seconds): 52.55
        Percent of CPU this job got: 160%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:34.07
        Maximum resident set size (kbytes): 2346012
```

This suggests my change made it worse :-)... let me play with it a bit

thantos:
Getting rid of the Mixin reduces the memory footprint a bit:         `Maximum resident set size (kbytes): 2339100`

sam-goodwin:
Have you tried https://github.com/functionless/functionless/pull/430

Running tsc wine test this very well. The problems will arise once you start doing interactive coding.

thantos:
Got rid of the OverloadUnion

```ts
export type SdkMethod<API> = API extends {
  (params: infer Input, cb: AnyFunction): AWS.Request<infer Output, any>;
  (cb: AnyFunction): any;
}
  ? (input: Input, options: SdkCallOptions) => Promise<Output>
  : never;
```

sam-goodwin:
Woah, TIL.

sam-goodwin:
> Got rid of the OverloadUnion
> 
> 
> 
> ```ts
> 
> export type SdkMethod<API> = API extends {
> 
>   (params: infer Input, cb: AnyFunction): AWS.Request<infer Output, any>;
> 
>   (cb: AnyFunction): any;
> 
> }
> 
>   ? (input: Input, options: SdkCallOptions) => Promise<Output>
> 
>   : never;
> 
> ```

Does it help with performance? The C sdk is so huge that i worry it'll still be slow

thantos:
The best experience by far is with #430. I say lets ship that PR and then we can merge this change on top and make any changes we need (I already started).

thantos:
@matthias-pichler-warrify 

Here is a working, building and test passing branch:
https://github.com/functionless/functionless/compare/main...matthias-pichler-warrify/main

* updated the code to use the generated typescript types from #430 (still a little worried about the performance, but should be enough to ship and we can play with it from there)
* merged from main
* Updated the localstack tests to use dynamo instead of alarms. There is a localstack bug cause alarms and some other resources to fail in Cfn. 
* updated and moved the generated types from how they were in #430.

thantos (approved):

thantos:
Thank you again for the inspiration and work here. Letting the tests run (needed approval to run) and will merge if they pass.

sam-goodwin:
Looks like we need to update the snapshot tests.
```
         },
         "TableName": "__REPLACED_TOKEN",
       },
-      "Resource": "__REPLACED_ARN",
+      "Resource": "arn:aws:states:::aws-sdk:dynamodb:getItem",
       "ResultPath": "$.heap28",
       "Type": "Task",
     },
     ```

sam-goodwin:
Build failed but not sure why. Re running them. 

@thantos we should look into why our tests have become fragile?

sam-goodwin:
Our build fails some time. I suspect because of local stack

Issue #418, opened 2022-08-17
User Authorization


sam-goodwin:
A staple of any cloud application is the ability to have users sign up, sign in and authenticate.

- [ ] AWS Cognito User Pools with type-safe triggers
- [ ] Non-AWS providers? Auth0, Custom, others?

Issue #134, opened 2022-05-17
Imported or wrapped constructs should use `from`


thantos:
All functionless resources which can wrap existing CDK constructs should use a `from` import pattern rather than accept the construct in the constructor.

Example - EventBus (already follows this pattern):

```ts
functionless.EventBus.fromBus(new aws_events.EventBus(stack, name, props));
new functionless.EventBus(stack, "eventBus", props);
``` 

Example of a invalid pattern:

```ts
new functionless.Table(new aws_dynamo.Table(stack, "table"))
```

Should be 

```ts
functionless.Table.fromTable(new aws_dynamo.Table(stack, "table", props));
new Table(stack, "table", props);
```

* [x] Event Bridge
* [ ] Lambda - Part of #103 
* [ ] Dynamo (table)
* [ ] Step Functions
* [x] AppSync - Does not wrap a construct.

PR #419, opened 2022-08-17
feat(user-pool): add a UserPool Construct with type-safe triggers

sam-goodwin:
Implements AWS Cognito User Pools for #418

Adds a new UserPool Construct that is a light wrapper around `aws_cogntito.UserPool`, adding typed  Triggers using Functionless's `Function`.

There are three ways to configure a Trigger:

1. pass a [Function](./function/index.md) to the `lambdaTriggers` property when instantiating the `UserPool`.

```ts
// option 1 - in-line the Function
const userPool = new UserPool(stack, "UserPool", {
  lambdaTriggers: {
    createAuthChallenge: new Function(
      stack,
      "CreateAuthChallenge",
      async (event) => {
        // implement logic for the CreateAuthChallenge lifecycle event
        return event;
      }
    ),
  },
});
```

2. Call `userPool.on`.

```ts
userPool.on(
  "createAuthChallenge",
  new Function(stack, "CreateAuthChallenge", async (event) => {
    // implement logic for the CreateAuthChallenge lifecycle event
    return event;
  })
);
```

3. Call the specific `userPool.onX` method:

```ts
userPool.onCreateAuthChallenge(
  new Function(stack, "CreateAuthChallenge", async (event) => {
    // implement logic for the CreateAuthChallenge lifecycle event
    return event;
  })
);
```


src/user-pool.ts
```diff
@@ -0,0 +1,233 @@
...
+  readonly customEmailSender?: Trigger<lambda.CustomEmailSenderTriggerEvent>;
+  /**
+   * Amazon Cognito invokes this trigger to send SMS notifications to users.
+   * @see https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-custom-sms-sender.html
+   * @default - no trigger configured
+   */
+  readonly customSmsSender?: Trigger<lambda.CustomSMSSenderTriggerEvent>;
+}
+
+// map a friendly trigger name to the {@link aws_cognito.UserPoolOperation}.
```
thantos, line 88: nit use block comment to get `@link`

src/user-pool.ts
```diff
@@ -0,0 +1,233 @@
...
+                ]
+              )
+            )
+          : undefined,
+      });
+    }
+  }
+
+  /**
+   * Configure a {@link Function} to handle AWS Cognito Triggers.
```
thantos, line 174: Nit: `Configure AWS Cognito to trigger a {@link Function}.` (The function is usable in theory).

src/user-pool.ts
```diff
@@ -0,0 +1,233 @@
...
+  readonly resource: aws_cognito.UserPool;
+
+  /**
+   * Create a new {@link UserPool} from an existing {@link aws_cognito.UserPool}.
+   *
+   *
+   * @param resource an existing UserPool Construct.
+   * @see {@link aws_cognito.UserPool}
+   */
+  constructor(resource: aws_cognito.UserPool);
```
thantos, line 136: All of the import cases should use a "from" method. #134

src/util.ts
```diff
@@ -27,6 +27,7 @@ import {
 } from "./guards";
 import { FunctionlessNode } from "./node";
 
+export type Mixin<T> = T;
```
thantos, line 30: What does this do?

thantos (changes requested):

src/user-pool.ts
sam-goodwin: I can add the `from` method but am I forced to remove the constructor? I don't need a `BaseUserPool`.

src/user-pool.ts
```diff
@@ -190,44 +333,240 @@ export class UserPool {
...
+   * ```
+   *
+   * @see https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-custom-message.html
+   * @default - no trigger configured
+   */
+  public onCustomMessage(
+    handler: Trigger<lambda.CustomMessageTriggerEvent>
+  ): void {
+    return this.on("customMessage", handler);
+  }
```
thantos, line 393: Why did you need to get rid of the dynamic ones? Just curious, fine with explicit.

src/user-pool.ts
thantos: I guess we have been removing them, shouldn't we be consistent? The harm would be that multiple constructor overrides would get messy or if there are non-import only methods, it would be harder to model that.

For example StepFunctions exposes `definition`, unless its an imported Sfn. 

Maybe its more of an issue for transpiled resources.

The argument for keeping the constructor is to avoid the base class and inheritance?

src/user-pool.ts
```diff
@@ -0,0 +1,572 @@
...
+  defineAuthChallenge: aws_cognito.UserPoolOperation.DEFINE_AUTH_CHALLENGE,
+  postAuthentication: aws_cognito.UserPoolOperation.POST_AUTHENTICATION,
+  postConfirmation: aws_cognito.UserPoolOperation.POST_CONFIRMATION,
+  preAuthentication: aws_cognito.UserPoolOperation.PRE_AUTHENTICATION,
+  preSignUp: aws_cognito.UserPoolOperation.PRE_SIGN_UP,
+  preTokenGeneration: aws_cognito.UserPoolOperation.PRE_TOKEN_GENERATION,
+  userMigration: aws_cognito.UserPoolOperation.USER_MIGRATION,
+  verifyAuthChallengeResponse:
+    aws_cognito.UserPoolOperation.VERIFY_AUTH_CHALLENGE_RESPONSE,
+};
```
thantos, line 236: Is there an escape hatch when a new trigger name is added? (idk if this happens).

Could we accept the actual value behind the CDK enum instead of remapping a string to the enum? That way a user could override and pass in any name if a new one is added.

src/user-pool.ts
sam-goodwin: I was worried it would confuse developers and we can't expand on doc and typedoc is probably terrible

sam-goodwin: Although I was particularly proud of that. It felt very much like a macro lol

src/user-pool.ts
```diff
@@ -0,0 +1,572 @@
...
+   * @returns a Functionless {@link UserPool}.
+   */
+  public static from(resource: aws_cognito.UserPool): UserPool {
+    return new UserPool(resource);
+  }
+
+  /**
+   * The underlying {@link aws_cognito.UserPool} instance.
+   */
+  readonly resource: aws_cognito.UserPool;
```
thantos, line 268: Was going to ask if this should be `IUserPool`, but I see that `addTrigger` is on `UserPool` only. Thats annoying. I bet imported user pools cannot have triggers. 

Do we have a case where we'd want a user pool to be wrapped even if imported? (I think so?)

thantos: https://github.com/aws/aws-cdk/blob/v1.169.0/packages/@aws-cdk/aws-cognito/lib/user-pool.ts#L708

src/user-pool.ts
sam-goodwin: Mostly to avoid an extra class but I am fine with changing it to be consistent. I usually hate Base* classes but I also get the point off it confusing developers. The first constructor they see is the import one so that could be a problem. I can roll in a fix for secrets into this change.

src/user-pool.ts
thantos:
```ts
functionless.UserPool.from(aws_cognito.UserPool.fromUserPoolArn(...))
```

or

```ts
functionless.UserPool.fromUserPoolArn(...)
```

thantos: That said... step function relies on `StepFunction` not `IStepFunction` because `IStepFunction` doesn't support `stateMachineType` and thus cannot be validated... maybe we need to fix that.

thantos (approved):

src/user-pool.ts
sam-goodwin: Ok, made it possible al ways around.

```ts
test("add a trigger to a user pool using on(aws_cognito.UserPoolOperation)", () => {
  const stack = new Stack();

  const pool = new UserPool(stack, "UserPool");

  pool.on(
    aws_cognito.UserPoolOperation.DEFINE_AUTH_CHALLENGE,
    new Function(
      stack,
      "CreateAuthChallenge",
      // must provide explicit event type
      async (event: CreateAuthChallengeTriggerEvent) => {
        return event;
      }
    )
  );
});

```

sam-goodwin: This has the drawback of not inferring the type of the event. The enum is a synthetic enum where the enum items are objects (they are all the same type and can't be used as keys). So the user must annotate the `event: CreateAuthChallengeTriggerEvent ` type.

src/user-pool.ts
thantos: Yeah, it was cool, but agree about explicit being a better experience.

thantos:
Looks good, just missing the change to Secret as a mentioned: https://github.com/functionless/functionless/pull/419#discussion_r949785586

PR #421, opened 2022-08-17
feat(secrets-manager): add Secret Resource

### LINKED ISSUE #192: Secrets Manager integration
How about something like this?

```ts
interface Credentials {
  username: string,
  password: string,
}

// an existing secret stored in Secrets Manager
const secret = new Secret<Credentials>('my-secret-arn-or-name');

// 1. Secret ARN is passed as env var
// 2. IAM permissions are given to Fn
// 3. Fn gets the secret string at runtime, parses it and caches it (outside of the closure, once per container)
new Function(scope, 'Fn', async () => {
  const username = await secret.getValue().username;
  console.log('The username is: ' + username`);
});
```

Currently using a `CachedSecret` in my Lambda runtimes, but it's only part of the solution:

<details>
<summary>Code for CachedSecret</summary>
<p>

```ts
import { SecretsManager } from 'aws-sdk';
import { tryParse } from './utils';

export class CachedSecret<T> {
  private static secretsManager = new SecretsManager({ apiVersion: '2017-10-17' });

  private static cache: Map<string, any> = new Map();

  private readonly secretId: string;

  constructor(secretId: string) {
    this.secretId = secretId;
  }

  public async getValue(): Promise<T> {
    const cached = CachedSecret.cache.get(this.secretId);
    if (cached) {
      return cached;
    }

    const getSecretValue = await CachedSecret.secretsManager.getSecretValue({
      SecretId: this.secretId,
    }).promise();

    if (!getSecretValue.SecretString) {
      throw new Error('No secret string');
    }

    const parsed = tryParse<T>(getSecretValue.SecretString);
    if (!parsed) {
      throw new Error('Secret string is not JSON parsable');
    }

    CachedSecret.cache.set(this.secretId, parsed);
    return parsed;
  }
}
```

</p>
</details>

sam-goodwin:
Fixes #192
# Secret

A `Secret` Resource stores sensitive data so that it can be securely distributed to your application code, e.g. in a Lambda Function.

There are three types of Secrets:

1. `TextSecret`
2. `BinarySecret`
3. `JsonSecret`

## TextSecret

A `TextSecret` stores the secret encoded as UTF-8 text.

```ts
const secret = new TextSecret(scope, id);
```

## Get Secret Value

The value of the Secret can be obtained by calling `getSecretValue`.

```ts
const apiKeySecret = new TextSecret(scope, "APIKey");

new Function(scope, "Foo", async () => {
  const apiKey = (await apiKey.getSecretValue()).SecretValue;

  // use within your application
});
```

## Put Secret Value

The value of the Secret can be updated by calling `putSecretValue`.

```ts
const apiKeySecret = new TextSecret(scope, "APIKey");

new Function(scope, "Foo", async () => {
  await apiKey.putSecretValue({
    SecretValue: "new api key",
  });
});
```

## BinarySecret

A `BinarySecret` stores the secret as raw Binary data using NodeJS's `Buffer` type.

```ts
const apiKeySecret = new BinarySecret(scope, "APIKey");
```

The interface is the same as TextSecret, except that the secret value is wrapped in a `Buffer`.

## JsonSecret

The `JsonSecret` provides a type-safe interface over a Secret and automatic serialization to and from JSON. It is recommended to use JsonSecret when storing structured Secret data.

```ts
interface UserPass {
  username: string;
  password: string;
}

const userPass = new JsonSecret<UserPass>(scope, "UserPass");

new Function(scope, "Foo", async () => {
  const { username, password } = (await userPass.getSecretValue()).SecretValue;

  await userPass.putSecretValue({
    // type-checked SecretValue
    SecretValue: {
      username,
      password,
    },
  });
});
```

## Initializing the Secret Value

:::warning
AWS will throw a ResourceNotFoundException if there is no secret value stored in the secret.

> An error occurred (ResourceNotFoundException) when calling the GetSecretValue operation: Secrets Manager can't find the specified secret value for staging label: AWSCURRENT

If you encounter this problem, be sure to upload an initial version of the Secret. This can be achieved with the CDK's [`SecretValue`](https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_core.SecretValue.html) class.

```ts
const secret = new TextSecret(scope, id, {
  secretStringValue: SecretValue.ssmSecure("<parameter name>"),
});
```

This applies to all Secret types - TextSecret, BinarySecret, JsonSecret
:::


src/secret.ts
```diff
@@ -0,0 +1,221 @@
...
+    async function getSecretValue(
+      [input]: [
+        input?: Omit<AWS.SecretsManager.GetSecretValueRequest, "SecretId">
+      ],
+      context: NativePreWarmContext
+    ) {
+      const client = context.getOrInit<AWS.SecretsManager>(
+        PrewarmClients.SECRETS_MANAGER
+      );
+      console.log(secretArn);
```
thantos, line 212: nit: remove

thantos (changes requested):

src/secret.ts
```diff
@@ -0,0 +1,221 @@
...
+   * ```
+   */
+  public readonly putSecretValue;
+
+  public readonly getJsonValue;
+
+  /**
+   * Get the
+   */
+  public readonly getValue;
```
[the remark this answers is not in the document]
sam-goodwin: Draft :)

src/secret.ts
```diff
@@ -0,0 +1,221 @@
...
+      ) => Promise<
+        AWS.SecretsManager.PutSecretValueResponse & {
+          SecretValue?: SecretData;
+        }
+      >
+    >({
+      kind: "AWS.SecretsManager.PutSecretValue",
+      native: {
+        bind: (context) => {
+          this.resource.grantRead(context.resource);
```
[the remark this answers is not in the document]
sam-goodwin: It doesn't even need read I don't think lol

src/secret.ts
sam-goodwin: I'm debugging why local stack can't find the secret

src/secret.ts
```diff
@@ -0,0 +1,221 @@
...
+    if (args.length === 1) {
+      this.resource = args[0];
+      this.storeAsJson = true;
+    } else {
+      const [scope, id, props] = args;
+      this.resource = new aws_secretsmanager.Secret(scope, id, props);
+      this.storeAsJson = props?.storeAsJson ?? true;
+    }
+
+    const storeAsJson = this.storeAsJson;
```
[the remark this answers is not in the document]
sam-goodwin: I don't think we can assume it's always json. Maybe we a JsonSecret class?

src/secret.ts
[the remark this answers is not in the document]
thantos: You requested review :-P.

src/secret.ts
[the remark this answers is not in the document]
thantos: but isn't `this.storeAsJson` always accurate?

I was wondering about a separate class too. If we can simplify the contract I'd say make a separate class. If it reduces the risk of calling `getJson` when it will fail too,

src/secret.ts
thantos: ohh, nice

src/secret.ts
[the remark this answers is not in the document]
sam-goodwin: It's not the same. If the Secret is just an API key, e.g. `ABCD1234567`, that's not valid JSON. It would have to be `"ABCD1234567"`. There's a difference between a JSON-encoded string and arbitrary text.

sam-goodwin:
Think we should add support for these in Step Functions? It seems dangerous to store them in state but might be fine in an Express Step Function? No different to in a Lambda's memory?

thantos:
> Think we should add support for these in Step Functions? It seems dangerous to store them in state but might be fine in an Express Step Function? No different to in a Lambda's memory?

I don't think we should be opinionated. If SFN supports it, we should too.

test/secret.localstack.test.ts
```diff
@@ -0,0 +1,80 @@
...
+
+      const func = new Function(
+        scope,
+        "Func",
+        {
+          ...localstackClientConfig,
+        },
+        async (input: "get" | UserPass) => {
+          if (input === "get") {
+            return secret.getJsonValue();
```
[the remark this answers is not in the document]
sam-goodwin: It throws an error. Customers can use `TextSecret` or `BinarySecret` otherwise.

sam-goodwin:
> > Think we should add support for these in Step Functions? It seems dangerous to store them in state but might be fine in an Express Step Function? No different to in a Lambda's memory?
> 
> I don't think we should be opinionated. If SFN supports it, we should too.

Does #414 cover this for now?

src/function-prewarm.ts
```diff
@@ -6,7 +6,9 @@ export interface PrewarmClientInitializer<T, O> {
   init: (key: string, props?: PrewarmProps) => O;
 }
 
+// TODO: use names that align with the AWS.SDK
 export type ClientName =
+  | "SECRETS_MANAGER"
```
thantos, line 11: 100%, do this or create issue?

src/secret.ts
```diff
@@ -0,0 +1,332 @@
...
+      this.resource = args[0] as aws_secretsmanager.ISecret;
+      props = args[1];
+    } else {
+      props = args[2] as SecretProps;
+      this.resource = new aws_secretsmanager.Secret(args[0], args[1], props);
+    }
+    const serialize = this.serialize;
+    const deserialize = this.deserialize;
+
+    const secretArn = this.resource.secretArn;
```
thantos, line 80: nit: comment on why these are here?

src/secret.ts
```diff
@@ -0,0 +1,332 @@
...
+  } else if (secretValue.SecretBinary instanceof Uint8Array) {
+    return Buffer.from(secretValue.SecretBinary).toString("utf8");
+  } else if (secretValue.SecretBinary instanceof Blob) {
+    return Buffer.from(await secretValue.SecretBinary.arrayBuffer()).toString(
+      "utf8"
+    );
+  } else {
+    throw new Error(
+      `unable to parse SecretValue, expected a string or Buffer, but received `
+    );
```
thantos, line 330: Synth error?

thantos (approved):

src/function-prewarm.ts
sam-goodwin: I can just do it if we think it's not worth being in its own PR? My only thought is that it will affect #405

src/secret.ts
sam-goodwin: It does not happen during synth though?

src/function-prewarm.ts
sam-goodwin: going ahead with this change.

src/secret.ts
sam-goodwin: Maybe we should rename SynthError to FunctionlessErr, or FunError lol

src/function-prewarm.ts
sam-goodwin: Done.

PR #425, opened 2022-08-19
feat(step-function): optimizer

thantos:
* Attempt to reduce unnecessary variables in a ASL state machine.
* Add options to turn on or off optimizations


src/util.ts
```diff
@@ -408,3 +408,153 @@ export const invertBinaryOperator = (op: BinaryOp): BinaryOp => {
...
+  return (
+    path1Segs.length === path2Segs.length &&
+    jsonPathSegmentStartsWith(path1Segs, path2Segs)
+  );
+}
+
+/**
+ * Formats a series of json path segments into a value json path.
+ *
+ * Attempts to determine of a segment can be represented using dot notation or element access notation.
```
sam-goodwin, line 481: Nit: "if"

test/step-function.test.ts
```diff
@@ -510,6 +510,88 @@ test("boolean logic", () => {
...
+    {
+      optimization: {
+        joinConsecutiveChoices: false,
+        optimizeVariableAssignments: false,
+        removeNoOpStates: false,
+        removeUnreachableStates: false,
+      },
+    } as ASLOptions,
+  ],
+])("%s optimizer", (_, opt) => {
```
sam-goodwin, line 526: Love this test. 

Does this mean that all optimizations are turned on by default?

test/step-function.test.ts
```diff
@@ -4019,6 +4132,37 @@ describe("binding", () => {
     expect(normalizeDefinition(definition)).toMatchSnapshot();
   });
 
+  // TODO: https://github.com/functionless/functionless/issues/420
+  // https://mathiasbynens.be/notes/javascript-properties
+  test.skip("unicode variable names", () => {
```
sam-goodwin, line 4137: This test is so awesome. Shame we skip it

test/step-function.test.ts
```diff
@@ -4019,6 +4132,37 @@ describe("binding", () => {
...
+    const { stack } = initStepFunctionApp();
+    const definition = new StepFunction(stack, "machine1", async () => {
+      var H̹̙̦̮͉̩̗̗ͧ̇̏̊̾Eͨ͆͒̆ͮ̃͏̷̮̣̫̤̣Cͯ̂͐͏̨̛͔̦̟͈̻O̜͎͍͙͚̬̝̣̽ͮ͐͗̀ͤ̍̀͢M̴̡̲̭͍͇̼̟̯̦̉̒͠Ḛ̛̙̞̪̗ͥͤͩ̾͑̔͐ͅṮ̴̷̷̗̼͍̿̿̓̽͐H̙̙̔̄͜ = 42;
+      const _ = "___";
+      const $ = "$$";
+      const ƒ = {
```
sam-goodwin, line 4143: Ooooooo look at this nice f

src/asl.ts
```diff
@@ -513,6 +518,15 @@ const FUNCTIONLESS_CONTEXT_NAME = "fnl_context";
  */
 const FUNCTIONLESS_CONTEXT_JSON_PATH = `$.${FUNCTIONLESS_CONTEXT_NAME}`;
 
+export interface ASLOptions {
+  /**
+   * Options for the ASL Optimizer.
+   *
+   * Default: Suggested
```
sam-goodwin, line 525: What does this mean

src/asl.ts
```diff
@@ -854,7 +870,7 @@ export class ASL {
 
         const initializer: ASLGraph.SubState | ASLGraph.NodeState = (() => {
           /**ForInStmt
-           * Assign the value to $.0__[variableName].
+           * Assign the value to $.fnls__0__[variableName].
```
sam-goodwin, line 873: Why change?

src/asl.ts
```diff
@@ -5526,28 +5253,1768 @@ export namespace ASLGraph {
...
-      isPresentAndNotNull(v),
-      or(
-        and(isString(v), not(stringEquals(v, ""))),
-        and(isNumeric(v), not(numericEquals(v, 0))),
-        and(isBoolean(v), booleanEquals(v, true))
-      )
+/**
+ * Operations that can help "optimize" an ASL state machine.
+ */
+namespace ASLOptimizer {
```
sam-goodwin, line 5259: Maybe this can be in its own file?

src/asl.ts
```diff
@@ -5526,28 +5253,1768 @@ export namespace ASLGraph {
...
+        return (
+          name !== startState &&
+          isPassState(state) &&
+          !!state.Next &&
+          !(
+            state.End ||
+            state.InputPath ||
+            state.OutputPath ||
+            state.Parameters ||
+            state.Result ||
```
sam-goodwin, line 5313: What if this was an empty string or 0 -ie a falsy value

src/asl.ts
```diff
@@ -5526,28 +5253,1768 @@ export namespace ASLGraph {
...
+   *       "string.$": "States.Array($.var1)"
+   *    },
+   *    ResultPath: "$.var2"
+   * }
+   * =>
+   * from: "$.var1"
+   * to: "$.var2"
+   */
+  interface VariableIntrinsicUsage {
+    type: "Intrinsic";
```
sam-goodwin, line 5356: Should we stick with our `kind` convention for consistency?

src/asl.ts
```diff
@@ -5526,28 +5253,1768 @@ export namespace ASLGraph {
...
+    usage: {
+      id: number;
+      index: number;
+      usage: StateVariableUsage;
+    }[];
+  }
+
+  function isVariableAssignment(variable: string, usage: VariableUsage) {
+    return usage.type === "Assignment" ||
+      usage.type === "PropAssignment" ||
```
sam-goodwin, line 5485: Learning from our past mistakes, add isX guard functions?

src/asl.ts
```diff
@@ -5526,28 +5253,1768 @@ export namespace ASLGraph {
...
+      ? jsonPathStartsWith(usage.from, variable)
+      : // stateoutput and literal assignments cannot use other variables
+      usage.type === "LiteralAssignment" ||
+        usage.type === "LiteralPropAssignment" ||
+        usage.type === "StateOutput"
+      ? false
+      : assertNever(usage);
+  }
+
+  function analyzeVariables(
```
sam-goodwin, line 5521: Need some docs for these core functions. "Analyze" is quite vague

src/asl.ts
```diff
@@ -5526,28 +5253,1768 @@ export namespace ASLGraph {
...
+      ),
+      usage: variableUsageWithId.filter(({ usage }) =>
+        isVariableUsage(v, usage)
+      ),
+    }));
+
+    return { variableUsages, variableNames: [...variableNames], stats };
+
+    function depthFirst(stateName: string) {
+      if (visited.has(stateName)) return;
```
sam-goodwin, line 5568: nit: always use braces

src/asl.ts
```diff
@@ -5526,28 +5253,1768 @@ export namespace ASLGraph {
...
+        isVariableAssignment(v, usage)
+      ),
+      usage: variableUsageWithId.filter(({ usage }) =>
+        isVariableUsage(v, usage)
+      ),
+    }));
+
+    return { variableUsages, variableNames: [...variableNames], stats };
+
+    function depthFirst(stateName: string) {
```
sam-goodwin, line 5567: Curious why this isn't a pure function?

src/asl.ts
```diff
@@ -5526,28 +5253,1768 @@ export namespace ASLGraph {
...
+      names.forEach((n) => variableNames.add(n));
+      visitTransition(state, depthFirst);
+    }
+  }
+
+  function getVariableUsage(state: State): VariableUsage[] {
+    const usages: VariableUsage[] = (() => {
+      if (isPassState(state)) {
+        if (state.ResultPath) {
+          if (state.Result !== undefined) {
```
sam-goodwin, line 5587: Yep, here we are checking Result is defined. Previously when finding empty Pass states you were checking Result is truthy. Make sure you treat them consistently.

src/asl.ts
```diff
@@ -5526,28 +5253,1768 @@ export namespace ASLGraph {
...
+              {
+                type: "LiteralAssignment",
+                to: state.ResultPath,
+                value: state.Result,
+              },
+            ];
+          } else if (state.InputPath) {
+            if (
+              state.InputPath.includes("?(") ||
+              state.InputPath.includes(":")
```
sam-goodwin, line 5598: What is special about these

src/asl.ts
```diff
@@ -5526,28 +5253,1768 @@ export namespace ASLGraph {
...
+        ...usages,
+        {
+          type: "ReturnUsage" as const,
+          from: state.ResultPath,
+        },
+      ];
+    }
+    return usages;
+
+    function choiceUsage(condition: Condition): string[] {
```
sam-goodwin, line 5714: What do you mean by choicrUsage?

src/asl.ts
```diff
@@ -5526,28 +5253,1768 @@ export namespace ASLGraph {
...
+      const vars = new Set<string>();
+      for (const key in condition) {
+        if (key === "Variable") {
+          vars.add(condition.Variable!);
```
sam-goodwin, line 5718: Update to pure function?

src/asl.ts
```diff
@@ -5526,28 +5253,1768 @@ export namespace ASLGraph {
...
+          return choiceUsage(condition.Not!);
+        } else if (key.endsWith("Path")) {
+          // @ts-ignore
+          vars.add(condition[key]!);
+        }
+      }
+      return [...vars];
+    }
+
+    function catchAssignment(_catch?: Catch[]): StateOutput[] {
```
sam-goodwin, line 5732: These function names confuse me. Is this getting something or doing something?

src/asl.ts
```diff
@@ -5526,28 +5253,1768 @@ export namespace ASLGraph {
...
+            type: "FilterPropAssignment",
+            from: x,
+            to: resultPath,
+            props,
+          }));
+        }
+      }
+      // if the value is json path, it will not be a boolean or string
+      return [
+        resultPath === "[[task]]" || resultPath === "[[map]]"
```
sam-goodwin, line 5822: What are these special values?

src/asl.ts
```diff
@@ -5526,28 +5253,1768 @@ export namespace ASLGraph {
...
+    }
+
+    /**
+     * Pull variable references out of intrinsic functions and json path filters.
+     *
+     * States.Array($.var) => $.var
+     * $.arr[?(['value'] == $.var)] => $.arr, $.var
+     */
+    function extractVariableReferences(jsonPath: string) {
+      return jsonPath.match(/(\$\.?[^,)]+)+/g) ?? [];
```
sam-goodwin, line 5845: Are there any risks using regexes here?

src/asl.ts
```diff
@@ -5526,28 +5253,1768 @@ export namespace ASLGraph {
...
+    })(state);
+
+    return names.filter((n): n is string => !!n);
+  }
+
+  export const DefaultOptimizeOptions: OptimizeOptions = {
+    optimizeVariableAssignments: true,
+    removeUnreachableStates: true,
+    joinConsecutiveChoices: true,
+    removeNoOpStates: true,
```
sam-goodwin, line 5880: All on by default?

src/asl.ts
```diff
@@ -5526,28 +5253,1768 @@ export namespace ASLGraph {
...
+  };
+
+  export interface OptimizeOptions {
+    /**
+     * Attempts to collapse assignments that provide no functional value.
```
sam-goodwin, line 5885: What do you mean "attempts"? Can we be more definitive ?

src/asl.ts
```diff
@@ -5526,28 +5253,1768 @@ export namespace ASLGraph {
...
+     * Collapses assignments created by the interpreter and the input source.
+     *
+     * ```ts
+     * const a = "a"
+     * const b = a;
+     * =>
+     * const b = "a"
+     * ```
+     */
+    optimizeVariableAssignments: boolean;
```
sam-goodwin, line 5896: This is a vague name. Perhaps:

> inlineRedundantVariableAssignments

src/asl.ts
```diff
@@ -5526,28 +5253,1768 @@ export namespace ASLGraph {
...
+    /**
+     * Traverses the graph transitions and removes any states that cannot be reached.
+     * Unreachable states will fail on deployment to AWS Step Functions.
+     *
+     * Note: If this optimization is turned off, it is likely the output ASL will fail.
+     *       Functionless creates some unreachable that this logic is expected to remove.
+     *       Only turn this off if you intend to preform similar logic or use the ASL for another
+     *       purpose than AWS Step Functions.
+     */
+    removeUnreachableStates: boolean;
```
sam-goodwin, line 5906: Why would we ever turn this off? Should it even be an option?

src/asl.ts
```diff
@@ -5526,28 +5253,1768 @@ export namespace ASLGraph {
...
+     * if(a && b) {}
+     * else(a) {}
+     * ```
+     *
+     * In ASL this may remove unnecessary state transitions without any impact on the logic.
+     *
+     * Note: This could make the ASL itself larger when the same Choice is merged into multiple calling choices.
+     *       Turn off at the cost of adding more state transitions.
+     */
+    joinConsecutiveChoices: boolean;
```
sam-goodwin, line 5929: Love it!

Do we have a test case where there are subsequent choices except the inner one has an else?

```ts
if(a) {
  if(b) {}
  else {}
}
```

src/asl.ts
```diff
@@ -5526,28 +5253,1768 @@ export namespace ASLGraph {
...
+    /**
+     * The Functionless transpiler and sometimes user can can create states that don't do anything.
```
sam-goodwin, line 5931: Examples?

src/asl.ts
```diff
@@ -5526,28 +5253,1768 @@ export namespace ASLGraph {
...
+    const reducedGraph = reduceGraph(startState, states, options);
+
+    // finally run remove unreachable states (when enabled) to ensure any isolated sub-graphs are removed.
+    return options.removeUnreachableStates
+      ? removeUnreachableStates(startState, reducedGraph)
+      : reducedGraph;
+  }
+
+  /**
+   * Run optimizations that reduce the number of states in the graph
```
sam-goodwin, line 5961: Isn't that all of them?

src/asl.ts
```diff
@@ -5526,28 +5253,1768 @@ export namespace ASLGraph {
...
+      !(
+        sourceAssign.usage.type === "Assignment" ||
+        sourceAssign.usage.type === "LiteralAssignment"
+      )
+    ) {
+      return;
+    }
+
+    /**
+     * For now, all optimizations only support a single usage of the current variable.
```
sam-goodwin, line 6190: Is there an issue tracking the outstanding tasks?

sam-goodwin (changes requested):
Amazing change. Great job mr compiler man. 

I'm not all the way through but here is some initial feedback.

src/util.ts
```diff
@@ -408,3 +408,153 @@ export const invertBinaryOperator = (op: BinaryOp): BinaryOp => {
...
+      } else if (seg.startsWith("['")) {
+        return seg.substring(2, seg.length - 2);
+      } else if (seg.startsWith("[")) {
+        return Number(seg.substring(1, seg.length - 1));
+      } else {
+        throw new Error("Invalid json path: " + jsonPath);
+      }
+    }) ?? []
+  );
+}
```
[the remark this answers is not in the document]
thantos: Good idea, was also worried about this, will try: https://github.com/dchester/jsonpath#readme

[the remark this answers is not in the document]
thantos: Did a spike on this and found the results disappointing, current state works for now.

https://twitter.com/sussmansa/status/1562523469094330369

test/step-function.test.ts
thantos: yeah. For now at least.  

```ts
  export const DefaultOptimizeOptions: OptimizeOptions = {
    optimizeVariableAssignments: true,
    removeUnreachableStates: true,
    joinConsecutiveChoices: true,
    removeNoOpStates: true,
  };
```

I imagine we'll create new, experimental ones or versions of existing ones and make a breaking change to turn them on.

thantos: > Love this test.

I am glad I wrote it, found that options in SFN were borked this whole time 🤒

test/step-function.test.ts
thantos: Just need to go the ticket to make sure it is all supported. Mostly need to normalize to valid characters in ASL.

src/asl.ts
thantos: The default is the suggested set of optimizations, haha. I can elaborate.

thantos: or this

```ts
   * @default ASLOptimizer.DefaultOptimizeOptions
```

src/asl.ts
thantos: ha, it could :) though there will be some circular references if the SFN types do not move too. leave it for now?

src/asl.ts
thantos: (this is existing code I moved)

I don't think any would matter as falsey. For input path, result, result selector, or parameters to matter, ResultPath would need to be truthy. Output path cannot be null.

```
Result: false,
ResultPath: undefined
```
^ does nothing.

src/asl.ts
thantos: breaks :(

That issue I tweeted about, Intrinsic functions cannot handle json paths that start with a number, underscore, etc, so better to just use valid paths.

src/asl.ts
thantos: On second thought, this file may be why ts is crashing for me. Will break it all up.

src/asl.ts
thantos: In that case it didn't matter because one of them would need to be truthy for the falsey to matter. In this case it matters because the result path is truthy.

src/asl.ts
thantos: https://github.com/functionless/functionless/pull/437

src/asl.ts
thantos: Yes, user should not have to know about these unless they start to break or if we make more experimental ones.

src/asl.ts
thantos: In case someone wanted to implement a different algorithm or if we want to use it for debugging. Doesn't hurt to have the flag.

src/asl.ts
thantos: (this was in there before)

We did not, added one

```ts
      if (input.val !== "a") {
        if (input.val === "b") {
          return "hullo";
        }
      }
      return "woop";
```

```ts
    "if(input.val !== \"a\")": Object {
      "Choices": Array [
        Object {
          "And": Array [
            Object {
              "Not": Object {
                "And": Array [
                  Object {
                    "IsPresent": true,
                    "Variable": "$.input.val",
                  },
                  Object {
                    "And": Array [
                      Object {
                        "IsString": true,
                        "Variable": "$.input.val",
                      },
                      Object {
                        "StringEquals": "a",
                        "Variable": "$.input.val",
                      },
                    ],
                  },
                ],
              },
            },
            Object {
              "And": Array [
                Object {
                  "IsPresent": true,
                  "Variable": "$.input.val",
                },
                Object {
                  "And": Array [
                    Object {
                      "IsString": true,
                      "Variable": "$.input.val",
                    },
                    Object {
                      "StringEquals": "b",
                      "Variable": "$.input.val",
                    },
                  ],
                },
              ],
            },
          ],
          "Next": "return \"hullo\"",
        },
        Object {
          "Next": "return \"wat\"",
          "Not": Object {
            "And": Array [
              Object {
                "IsPresent": true,
                "Variable": "$.input.val",
              },
              Object {
                "And": Array [
                  Object {
                    "IsString": true,
                    "Variable": "$.input.val",
                  },
                  Object {
                    "StringEquals": "a",
                    "Variable": "$.input.val",
                  },
                ],
              },
            ],
          },
        },
      ],
      "Default": "return \"woop\"",
      "Type": "Choice",
    },
```

src/asl.ts
sam-goodwin: Use @link but yes

src/asl.ts
sam-goodwin: Why does ResultPath matter? If Result: false and ResultPath: $.a what would happen

thantos:
```ts
            state.End ||
            state.InputPath ||
            state.OutputPath ||
            state.Parameters ||
            state.Result ||
            state.ResultPath
```

`true`?

src/asl.ts
sam-goodwin: Finding it hard to follow this reasoning ... but not sure what would be a better approach. Maybe add helpers, `isXYZ` that make the code more expressive?

src/asl.ts
sam-goodwin: Does the JSON Path re-write worry you?

PR #428, opened 2022-08-20
feat(sqs): implement Queue Construct with forEach capability

### LINKED ISSUE #58: SQS Queue type-safe integration class
Provide a wrapper class, `Queue<T>`, that models the SQS Queue Functionless integrations.

Must integrates with the following:

1. AppSync
```ts
const queue = new Queue<string>(new aws_sqs.Queue(..))

new AppsyncResolver<{id: string}>($context => {
  queue.sendMessage({
     MessageBody: id,
     MessageAttributes: ..
  });
});
```

2. Step Functions

```ts
new StepFunction(scope, id, (id: string) => {
  queue.sendMessage({
     MessageBody: id,
     MessageAttributes: ..
  })
})
```

3. Event Bridge
```ts
bus
  .when(scope, "rule", event => event.source === "lambda")
  .map(event => event.message)
  .pipe(queue)
```

4. Lambda
```ts
new Function(scope, id, async () => {
  await queue.sendMessage(..);
});
```

sam-goodwin:
Fixes #58 

Adds a `Queue` Construct for SQS with a `forEach` capability.

- [x] decide how we want to layer the implementation, starting with an unopinionated (pure AWS contract, where body is a `string`) and then providing a type-safe `Queue<T>` with serialization abstracted away.
- [x] `forEach`, `forEachBatch`
- [x] `map`, `flatMap`
- [x] `sendMessage`, `sendMessageBatch`, `receiveMessage`, `purge`

Example:
```ts
import { App, Stack } from "aws-cdk-lib";
import { Queue, Function, EventBus, Event, StepFunction } from "functionless";

const app = new App();
const stack = new Stack(app, "queue");

type CartEventEnvelope<E extends CartEvent = CartEvent> = Event<E, E["kind"]>;
type CartEvent = OrderPlacedEvent | OrderCancelledEvent;

interface OrderPlacedEvent {
  kind: "OrderPlacedEvent";
  orderId: string;
}
interface OrderCancelledEvent {
  kind: "OrderCancelledEvent";
  orderId: string;
}

// set up an EventBus containing all of our cart events
const events = new EventBus<CartEventEnvelope>(stack, "Events");

// create a queue to store failed Order Events
const failedOrderQueue = new Queue<OrderPlacedEvent>(
  stack,
  "dead letter queue"
);

// create a queue for processing Order Events
const orderQueue = new Queue<OrderPlacedEvent>(stack, "queue", {
  deadLetterQueue: {
    queue: failedOrderQueue,
    maxReceiveCount: 10,
  },
});

// filter OrderEvents from the Event Bus and route them to the orderQueue
events
  .when(
    stack,
    "Messages",
    (event): event is CartEventEnvelope<OrderPlacedEvent> =>
      event["detail-type"] === "OrderPlacedEvent"
  )
  .map((envelope) => envelope.detail)
  // TODO: support
  .pipe(orderQueue);

// kick off a Step Function to reliably process each order
orderQueue.messages().forEach(async (order) => {
  await processOrder({
    // idempotency on the orderId
    name: order.orderId,
    input: order,
  });
});

// use a Standard Step Function to orchestrate order fulfillment
const processOrder = new StepFunction(
  stack,
  "ProcessOrder",
  async (order: OrderPlacedEvent) => {
    await chargeCard(order);
    await dispatchOrder(order);
  }
);

const chargeCard = new Function(
  stack,
  "PlaceOrder",
  async (order: OrderPlacedEvent) => {
    console.log("Card Charged! (not)", order);
  }
);

const dispatchOrder = new Function(
  stack,
  "DispatchOrder",
  async (order: OrderPlacedEvent) => {
    console.log("Order Dispatched! (not)", order);
  }
);

// process all failed Order Events and re-send them for re-processing
failedOrderQueue.messages().forEach(async (message) => {
  await orderQueue.sendMessage({
    Message: message,
    // naive back-off policy for demonstration purposes only
    DelaySeconds: 60,
  });
});
```


src/queue.ts
```diff
@@ -0,0 +1,719 @@
...
+              ...input,
+              MessageBody: messageBody,
+              QueueUrl: updatedQueueUrl,
+            })
+            .promise();
+
+          return response;
+        },
+      },
+    });
```
thantos, line 408: whats the plan to support SFN?

src/queue.ts
```diff
@@ -0,0 +1,719 @@
...
+          : // this is unsafe - how can we ensure that, when no serializer is provided, then the message is always the raw string?
+            (record.body as unknown as Message),
+      })),
+    });
+  }
+
+  /**
+   * @hidden
+   */
+  public createResponseHandler(): (
```
thantos, line 602: where is this used?

thantos (changes requested):

src/function.ts
```diff
@@ -654,27 +656,52 @@ export class Function<
...
       ),
+      environment: {
+        functionless_infer: Lazy.string({
+          produce: () => {
+            // retrieve and bind all found native integrations. Will fail if the integration does not support native integration.
+            findAllIntegrations().forEach(({ integration, args }) => {
+              new IntegrationImpl(integration).native.bind(this, args);
+            });
+
+            return "DONE";
```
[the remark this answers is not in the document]
sam-goodwin: > nice

Not so sure I would describe it that way haah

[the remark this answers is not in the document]
thantos: you made it work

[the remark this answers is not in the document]
sam-goodwin: Make it work, make it right, make it fast ;)

src/function.ts
```diff
@@ -654,27 +656,52 @@ export class Function<
       onFailure: FunctionBase.normalizeAsyncDestination<OutPayload, Out>(
         onFailure
       ),
+      environment: {
+        functionless_infer: Lazy.string({
+          produce: () => {
+            // retrieve and bind all found native integrations. Will fail if the integration does not support native integration.
+            findAllIntegrations().forEach(({ integration, args }) => {
```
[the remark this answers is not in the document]
sam-goodwin: I tried moving the pre-warm there and it broke serialization - all the functions had the dummy error messag in there instead of code.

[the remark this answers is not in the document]
sam-goodwin: Tried a bunch of different things to try and get the closure serialization in there to support hoisting (i.e. a function referencing something declared after it) but with no luck. Would like to try again but didn't want to block this change on that.

src/integration.ts
```diff
@@ -314,9 +316,26 @@ export function findDeepIntegrations(
...
+          if (ast) {
+            visit(ast);
+          }
+        } else if (Array.isArray(value)) {
+          value.forEach(visitValue);
+        } else if (value && typeof value === "object") {
+          Object.values(value).forEach(visitValue);
+        }
+      })(node.ref());
     }
```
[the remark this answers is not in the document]
sam-goodwin: See the `seen.add(value)`, if they reference the same function it will only be processed once

src/iterable.ts
```diff
@@ -0,0 +1,1631 @@
...
+    ParsedRecord,
+    Response,
+    EventSourceConfig
+  > {
+    return this.flatMap(async (item, record, event, raw) => {
+      let pred = predicate(item, record, event, raw);
+      if (pred instanceof Promise) {
+        pred = await pred;
+      }
+      return pred ? [item] : [];
```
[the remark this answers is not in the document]
sam-goodwin: Well that's very good to know because that actually optimizes the synchronous case

src/iterable.ts
```diff
@@ -0,0 +1,1631 @@
...
+      return [...this.prev.getCallChain(), this.callbackfn];
+    } else {
+      return [this.callbackfn];
+    }
+  }
+
+  /**
+   * @hidden
+   */
+  private getSource(): EventSource<
```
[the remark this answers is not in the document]
sam-goodwin: Wouldn't making it a getter disguise the work it is doing which would have prevented you fro calling out the redundant `getSource`?

src/iterable.ts
```diff
@@ -0,0 +1,1631 @@
...
+  }
+}
+
+function awaitIfPromise<T>(value: T | Promise<T>): Promise<T> {
+  if (value instanceof Promise) {
+    return value;
+  } else {
+    return Promise.resolve(value);
+  }
+}
```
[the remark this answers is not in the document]
sam-goodwin: That's cool, will use this pattern and remove the awaitIfPromise function

src/queue.ts
sam-goodwin: I need to implement it. Forgot twice lol

src/queue.ts
sam-goodwin: In Iterator

src/function.ts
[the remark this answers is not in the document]
thantos: Add a ticket and Todo and I can look after?

src/integration.ts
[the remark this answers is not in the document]
thantos: For each entry point.

src/iterable.ts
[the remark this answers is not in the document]
thantos: Yes. Good point.

src/enumerable.ts
```diff
@@ -0,0 +1,234 @@
...
+  Raw = any
+> = (item: Item, record: Record, event: Event, raw: Raw) => Promise<Processed>;
+
+export type BatchProcessor<Element, Response, Event, Raw> = (
+  batch: Element[],
+  event: Event,
+  raw: Raw
+) => Promise<Response>;
+
+export interface IEnumerable<
```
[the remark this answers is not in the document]
sam-goodwin: I deleted this interface - replaced with IIterable

src/function.ts
```diff
@@ -654,27 +656,52 @@ export class Function<
...
 
-    const integrations = findDeepIntegrations(ast).map(
-      (i) =>
-        <IntegrationInvocation>{
-          args: i.args,
-          integration: i.expr.ref(),
-        }
-    );
-
     // retrieve and bind all found native integrations. Will fail if the integration does not support native integration.
```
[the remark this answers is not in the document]
sam-goodwin: I believe so - `.native` will be `undefined.

src/function.ts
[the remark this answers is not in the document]
sam-goodwin: https://github.com/functionless/functionless/issues/476

src/function.ts
```diff
@@ -654,27 +656,52 @@ export class Function<
...
+            throw Error(
+              `While serializing ${_resource.node.path}:\n\n${e.message}`
+            );
+          } else {
+            throw e;
+          }
+        }
+      })()
+    );
+
```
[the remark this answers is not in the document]
sam-goodwin: I don't understand what you mean. Are you suggesting that `node.addValidation` should be called before kicking off the serialize promise? I can't imagine that being important - validation would only be run during synthesis and after the Construct tree has been fully instantiated.

src/integration.ts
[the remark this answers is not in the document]
sam-goodwin: I don't understand.

src/queue.ts
sam-goodwin: Implemented sendMessage, sendMessageBatch, purge and receiveMessage

src/queue.ts
```diff
@@ -0,0 +1,1052 @@
...
+                      "States.JsonToString(${input.jsonPath}.MessageBody)",
+                  },
+                  Next: ASLGraph.DeferNext,
+                });
+                // it's unfortunate that we have to do this
+                // request.MessageBody = request.MessageBody.value
+                addState({
+                  Type: "Pass",
+                  InputPath: `${input.jsonPath}.MessageBody.value`,
+                  OutputPath: `${input.jsonPath}.MessageBody`,
```
thantos, line 452: ResultPath

OutputPath will erase all of the state, almost always use ResultPath

src/queue.ts
```diff
@@ -0,0 +1,1052 @@
...
+                  throw new SynthError(
+                    ErrorCodes.Invalid_Input,
+                    `the property 'MessageBody' is required in SendMessageRequest`
+                  );
+                }
+                return context.stateWithHeapOutput({
+                  Type: "Task",
+                  Resource: "arn:aws:states:::aws-sdk:sqs:sendMessage",
+                  Parameters: {
+                    ...input.value,
```
thantos, line 492: if Message is a constant, you'll end up with both `Message` and `Message.$`, which will fail.

src/queue.ts
```diff
@@ -0,0 +1,1052 @@
...
+                      Next: "unwrap Message",
+                    },
+                    "unwrap Message": {
+                      Type: "Pass",
+                      InputPath: "$.entry.MessageBody.value",
+                      ResultPath: "$.entry.MessageBody",
+                      OutputPath: "$.entry",
+                      End: true,
+                    },
+                  },
```
thantos, line 626: you should use `context.aslGraphToStates`, which will ensure the state names are globally unique, apply optimization, and you can use ASLGraph.

https://github.com/functionless/functionless/blob/main/src/step-function.ts#L456

src/queue.ts
```diff
@@ -0,0 +1,1052 @@
...
+                Iterator: {
+                  StartAt: "serialize Message",
+                  States: {
+                    "serialize Message": {
+                      Type: "Pass",
+                      Parameters: {
+                        "value.$":
+                          this.serializer.dataType === DataType.Json
+                            ? `States.JsonToString($.entry.MessageBody)`
+                            : "$.entry.MessageBody",
```
thantos, line 614: can you skip this if the map will be a no-op?

src/queue.ts
```diff
@@ -0,0 +1,1052 @@
...
+                // Because of limitations in `Parameters` - need support for `Parameters.$`
+                throw new SynthError(
+                  ErrorCodes.Invalid_Input,
+                  `the 'input' to receiveMessage must be an object literal`
+                );
+              }
+              addState(receiveAndParse(input.value));
+              return {
+                jsonPath: messagesHeap,
+              };
```
thantos, line 716: or you can put `output` on the SubState and return that.

```ts
return {
   stateState: "",
   output: { jsonPath: ... }
}
```

thantos: (almost never use addState, but it does work how you are using it)

src/queue.ts
```diff
@@ -0,0 +1,1052 @@
...
+                      Type: "Choice",
+                      Choices: [
+                        {
+                          IsPresent: true,
+                          Variable: `${messagesHeap}.Messages`,
+                          Next: "parse",
+                        },
+                      ],
+                      Default: ASLGraph.DeferNext,
+                    },
```
thantos, line 762: we remove all unreachable states, so you can add the state to the graph and it will be removed later. This is fine too.

src/queue.ts
```diff
@@ -0,0 +1,1052 @@
...
+                    : ASLGraph.DeferNext,
+              },
+              ...(serializer.dataType === DataType.Json
+                ? {
+                    parseIfMessages: {
+                      Type: "Choice",
+                      Choices: [
+                        {
+                          IsPresent: true,
+                          Variable: `${messagesHeap}.Messages`,
```
thantos, line 757: nit: `...ASL.isPresent()`

thantos (approved):

src/integration.ts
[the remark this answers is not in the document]
sam-goodwin: Because it has to for different functions right?

src/queue.ts
sam-goodwin: I tried this and got a type error - not a valid SubState. Looks like our types don't properly capture the contract.

src/queue.ts
sam-goodwin: Why is that the callee's responsibility

src/function.ts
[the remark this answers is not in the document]
thantos: Hmm. I remember having an issue, but I don't know what. Can you just remove this comment if you are not seeing an issue?

src/queue.ts
thantos: Just is right now. We could process map, parallel, etc at the end to ensure any January created state names are unique with sub-graphs.

src/asl/asl-graph.ts
```diff
@@ -1,5 +1,7 @@
 import { assertNever } from "../assert";
 import { SynthError, ErrorCodes } from "../error-code";
+// @ts-ignore tsdoc
+import type { Expr } from "../expression";
```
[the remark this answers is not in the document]
sam-goodwin: finally!

src/error-code.ts
```diff
@@ -1277,6 +1277,17 @@ export namespace ErrorCodes {
...
+  /**
+   * An Integration was called within a Lambda Function but it does not support the
+   * `native` interface.
+   */
+  export const Integration_does_not_support_native_interface: ErrorCode = {
+    code: 10037,
+    type: ErrorType.ERROR,
+    title:
+      "Integration used in Lambda Function does not support native interface",
+  };
```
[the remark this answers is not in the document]
sam-goodwin: Not that I can find?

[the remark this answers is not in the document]
sam-goodwin: Looks like it throws an `Error` by default - not a SynthError.

src/queue.ts
sam-goodwin: Does it matter if I ignore the return value of `aslGraphToStates`?

src/queue.ts
sam-goodwin: Oh it is, I just refactored the code into two branches. This ternary is no longer needed.

src/queue.ts
sam-goodwin: I thought that we could just chain together states and the underlying machine would flatten them out into a machine. That doesn't seem to be the case?

sam-goodwin: The types suggest I can return `startsAt` and `states` SubState from evalExpr?

image

sam-goodwin: Should I use `jsonPath` or `output: { jsonPath } ` for the heap var?

sam-goodwin: Got it working with this code! I assume your suggestion of `aslGraphToStates` was wrong?

```ts
return context.evalExprToJsonPath(entries, (entries) => {
  const heapVar = context.newHeapVariable();

  return {
    output: {
      jsonPath: heapVar,
    },
    startState: "serialize messages",
    states: {
      "serialize messages": {
        Type: "Map",
        ItemsPath: `${entries.jsonPath}.Entries`,
        Parameters: {
          "entry.$": "$$.Map.Item.Value",
        },
        ResultPath: heapVar,
        Next: "send message batch",
        Iterator: {
          StartAt: "serialize Message",
          States: {
            "serialize Message": {
              Type: "Pass",
              Parameters: {
                "value.$": `States.JsonToString($.entry.MessageBody)`,
              },
              ResultPath: "$.entry.MessageBody",
              Next: "unwrap Message",
            },
            "unwrap Message": {
              Type: "Pass",
              InputPath: "$.entry.MessageBody.value",
              ResultPath: "$.entry.MessageBody",
              OutputPath: "$.entry",
              End: true,
            },
          },
        },
      },
      "send message batch": {
        Type: "Task",
        Resource: "arn:aws:states:::aws-sdk:sqs:sendMessageBatch",
        Parameters: {
          QueueUrl: this.queueUrl,
          "Entries.$": heapVar,
        },
        ResultPath: heapVar,
        Next: ASLGraph.DeferNext,
      },
    },
  };
});
```

src/queue.ts
sam-goodwin: I know it's missing `states` - that's not what I mean. If I set the return type to explicitly a `ASLGraph.SubState`, it doesn't allow the `output` property. That suggested to me that I can't use both at one time. Looks like that's not the case though.

thantos: There is an `ASLGraph.OutputSubState`, which has the `output` property on it. `OutputSubState` and `OutputState` are `NodeResult`s. `SubState` and `State` are not. I wonder if it would be better to add an optional `state` property to `Output` instead of having an output on the `*State`.

src/queue.ts
thantos: Nope, my comment was just about `Iterator`. `aslGraphToStates` is a terminal call in the chain that output the final ASL format. The contents of `Map.Iterator` and `Parallel.Branches[]` are complete ASL graphs (`{StartAt, States}`) (which I have been calling SubGraphs). 

The suggestion was to do:

```ts
return context.evalExprToJsonPath(entries, (entries) => {
  const heapVar = context.newHeapVariable();

  return {
    output: {
      jsonPath: heapVar,
    },
    startState: "serialize messages",
    states: {
      "serialize messages": {
        Type: "Map",
        ItemsPath: `${entries.jsonPath}.Entries`,
        Parameters: {
          "entry.$": "$$.Map.Item.Value",
        },
        ResultPath: heapVar,
        Next: "send message batch",
        Iterator: context.aslGraphToStates({
          startState: "serialize Message",
          states: {
            "serialize Message": {
              Type: "Pass",
              Parameters: {
                "value.$": `States.JsonToString($.entry.MessageBody)`,
              },
              ResultPath: "$.entry.MessageBody",
              Next: "unwrap Message",
            },
            "unwrap Message": {
              Type: "Pass",
              InputPath: "$.entry.MessageBody.value",
              ResultPath: "$.entry.MessageBody",
              OutputPath: "$.entry",
              End: true,
            },
          }),
        },
      },
      "send message batch": {
        Type: "Task",
        Resource: "arn:aws:states:::aws-sdk:sqs:sendMessageBatch",
        Parameters: {
          QueueUrl: this.queueUrl,
          "Entries.$": heapVar,
        },
        ResultPath: heapVar,
        Next: ASLGraph.DeferNext,
      },
    },
  };
});
```

Yours will work, I think, util someone uses batch twice in the same machine or managed to create any other state with the same names.

sam-goodwin: The contents of Iterstor is an isolated machine right?

test/validate.test.ts
```diff
@@ -51,6 +51,7 @@ const skipErrorCodes: ErrorCode[] = [
   ErrorCodes.StepFunction_invalid_filter_syntax,
   // will not support integration specific validation, for now.
   ErrorCodes.Unsupported_AWS_SDK_in_Resource,
+  ErrorCodes.Integration_does_not_support_native_interface,
```
thantos, line 54: This error code no longer exists?

PR #431, opened 2022-08-21
chore: increase allowed memory of node process

sam-goodwin:
Builds are failing because node is running out of memory. To mitigate, this change increases the memory from 2GB to 8GB on the compile and test tasks.


sam-goodwin:
I'm not sure why memory has become a problem, but I suspect SWC is involved. Maybe we should prioritize https://github.com/functionless/ast-reflection/issues/20

thantos (approved):

PR #440, opened 2022-08-25
chore: test performance

thantos:
* Wrap validate test code in a describe to avoid dangling references
* Use docker-reuse to run lambdas in localstack. Uses a single docker container for all tests instead of one per lambda.

### DIFF
--- a/scripts/localstack
+++ b/scripts/localstack
@@ -2,9 +2,9 @@
 
 # Installs and runs localstack
 
-# # run all lambdas in the same container. 
-# export LAMBDA_EXECUTOR=docker-reuse
-# export LAMBDA_STAY_OPEN_MODE=0
+# run all lambdas in the same container. 
+export LAMBDA_EXECUTOR=docker-reuse
+export LAMBDA_STAY_OPEN_MODE=0
 # install LocalStack cli and awslocal
 pip install localstack --upgrade awscli-local[ver1]
 # install CDK Local
--- a/scripts/localstack
+++ b/scripts/localstack
@@ -4,7 +4,6 @@
 
 # run all lambdas in the same container. 
 export LAMBDA_EXECUTOR=docker-reuse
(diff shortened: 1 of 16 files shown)


scripts/localstack
```diff
@@ -2,6 +2,8 @@
 
 # Installs and runs localstack
 
+# run all lambdas in the same container. 
+export LAMBDA_EXECUTOR=docker-reuse
```
sam-goodwin, line 6: Oh that seems promising!

test/validate.test.ts
```diff
@@ -7,96 +7,98 @@ import { makeFunctionlessChecker } from "../src/checker";
...
-describe("all error codes tested", () => {
-  test.concurrent.each(
-    Object.values(ErrorCodes).filter((code) => !skipErrorCodes.includes(code))
-  )("$code: $title", async (code) => {
-    if (!file?.includes(`${code.code}`)) {
-      throw new Error(
-        `validate.test.ts does not emit any errors for ${code.title}`
-      );
-    }
+describe("validate", () => {
```
sam-goodwin, line 10: What does this achieve

thantos: the files and project were being created when the tests all load, but not used and released until the validate tests run and finish. 

By wrapping it, the references are not created until the tests run, reducing the impact on the other tests. I am guessing there was some race condition going on as well since validate would get stuck on my machine after all other suites finished, but not by itself. (related to time the project references are held onto I bet).

thantos:
Memory issues:

* [x] jest hangs - node's max memory setting is higher than host memory - ✅ reoslved - changed 8GB to 4GB max memory (7GB on host)
* [x] `killed` when jest runs out of memory - ✅ seems to happen less often with the right amount of memory.
* [ ] I think lambdas on localstack are taking too much memory when they stay open. Seeing if destroying stacks will free up the lambda resources after each localstack suite.

PR #444, opened 2022-08-25
fix(step-functions): props are not correctly passed to step function

### LINKED ISSUE #441: Updating the logging configuration for ExpressStepFunction does not trigger a redeploy
Steps to reproduce:

1. Deploy an `ExpressStepFunction` with no logging configuration
2. Add a logging configuration.
3. Re-deploy

Expected outcome:

The Step Function should be updated with the new logging config.

Actual outcome:

The logging configuration remains unmodified.

thantos:
Fixes #441

### DIFF
--- a/src/step-function.ts
+++ b/src/step-function.ts
@@ -1579,9 +1579,9 @@ function getStepFunctionArgs<
     | [func: StepFunctionClosure<Payload, Out>]
 ) {
   const props =
-    isFunctionLike(args[0]) || isErr(args[0])
+    isFunctionLike(args[0]) || isErr(args[0]) || typeof args[0] === "function"
       ? {}
-      : (args[1] as StepFunctionProps);
+      : args[0];
   const func = validateFunctionLike(
     args.length > 1 ? args[1] : args[0],
     "StepFunction"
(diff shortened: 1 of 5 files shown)


src/step-function.ts
```diff
@@ -1579,9 +1579,9 @@ function getStepFunctionArgs<
     | [func: StepFunctionClosure<Payload, Out>]
 ) {
   const props =
-    isFunctionLike(args[0]) || isErr(args[0])
+    isFunctionLike(args[0]) || isErr(args[0]) || typeof args[0] === "function"
       ? {}
-      : (args[1] as StepFunctionProps);
+      : args[0];
```
sam-goodwin, line 1584: What case is this catching? When would this be a function?

test/step-function.localstack.test.ts
```diff
@@ -124,6 +124,21 @@ localstackTestSuite("sfnStack", (testResource, _stack, _app) => {
     "hello world"
   );
 
+  test(
+    "props",
```
sam-goodwin, line 128: better test name?

src/step-function.ts
thantos: the actual change was changing the index from `1` to `0`, the function check narrows the types it comes in as without using `as`, it will actually be expressions, but the type on it is a function.

sam-goodwin (approved):

PR #449, opened 2022-08-27
fix(step-functions): compare undefined and null

### LINKED ISSUE #441: Updating the logging configuration for ExpressStepFunction does not trigger a redeploy

thantos:
Fixes #441

`x === undefined` and other comparison patterns.

Root Cause:
* Bad typo (didn't `return`)
* Missing `undefined` case when comparing two variables or variable and a constant.
* Empty `or` logic returned true (should be false)


src/asl.ts
```diff
@@ -5431,6 +5445,24 @@ export namespace ASLGraph {
...
+      if (ASLGraph.isJsonPath(right)) {
+        return ASL.and(
+          ASL.isNotPresent(left.jsonPath),
+          ASL.isNotPresent(right.jsonPath)
+        );
+      } else if (ASLGraph.isLiteralValue(right) && right.value === undefined) {
+        return ASL.isNotPresent(left.jsonPath);
+      }
+    }
+    return undefined;
```
sam-goodwin, line 5463: Is this case expected?

src/asl.ts
```diff
@@ -5261,36 +5270,41 @@ export namespace ASLGraph {
...
     ) {
+      // left is a condition and right is any value
       if (ASLGraph.isConditionOutput(left)) {
         return (
           ASLGraph.booleanCompare(left, right, operator) ?? ASL.falseCondition()
         );
       }
+
+      // returns comparisons if there are valid comparisons to return.
+      // comparisons return undefined when the right is a literal and not the right
```
sam-goodwin, line 5281: > when the right is a literal and not the right

Typo?

src/asl.ts
```diff
@@ -5261,36 +5270,41 @@ export namespace ASLGraph {
...
-            condition
-          )
-        );
-      }
-      return ASL.and(ASL.isPresent(left.jsonPath), condition);
+      return ASL.or(
+        // a === b while a and b are both not defined
+        ASLGraph.undefinedCompare(left, right, operator),
+        // a !== undefined && b !== undefined
+        ASL.and(
```
sam-goodwin, line 5295: I don't understand what the comments mean

sam-goodwin (approved):

src/asl.ts
thantos: Yeah, returns undefined when there is no comparison to add. If there are no comparisons at all, `false` is returned at runtime.

* `x === undefined` => returns conditional (`isNotPresent(x)`)
* `x === "a"` => return undefined
* `x === y` => return conditional (`isNotPresent(x) && isNotPresent(y)`) 
* `x < undefined` => return undefined

src/asl.ts
thantos: Woops, not relevant anymore.

src/asl.ts
```diff
@@ -5261,42 +5270,59 @@ export namespace ASLGraph {
...
-            condition
-          )
-        );
-      }
-      return ASL.and(ASL.isPresent(left.jsonPath), condition);
+      return ASL.or(
+        // for ==/===
+        // if right is undefined literal, check isNotPresent(left)
+        // if right is a variable, check isNotPresent(left) && isNotPresent(right)
+        ASLGraph.undefinedCompare(left, right, operator),
```
sam-goodwin, line 5296: What happens if this returns `undefined`?

thantos: We are trying to collect the comparisons we need, when the values are literals, we only do the comparisons of the type of the literal value. When the values are variable, we must do all of the comparisons to try to find a `true`.

So the `*compare` methods return a Condition when we need a runtime condition and `undefined` when we can assume the comparison is false.

sam-goodwin: I still don't understand what happens when `ASL.or(undefined, ..)`

thantos: ohh, `or` and `and` filter out undefined. if all undefined are passed to `or`, it returns falsy condition, if all undefined are passed to AND it returns a truthy condition.

PR #451, opened 2022-08-29
chore: doc changes

thantos:


website/docs/concepts/step-function/index.md
```diff
@@ -3,9 +3,21 @@
...
+
+```ts
+const serviceCall = new Function(scope, "serviceCall", async (input: string) => {
+  console.log(input);
+});
+new StepFunction(scope, "F", async (payload: { property?: string }) => {
   if (payload.property) {
-    await serviceCall(payload);
+    SFN.waitFor(10);
+    await serviceCall(payload.property);
```
sam-goodwin, line 20: Maybe add a comment here emphasizing how the `serviceCall` Function Construct can be called like an ordinary function

sam-goodwin (approved):

website/docs/concepts/step-function/index.md
thantos: > // the lambda function `serviceCall` is invoked like an ordinary function from the state machine.

sam-goodwin (approved):

Issue #443, opened 2022-08-25
ApiGatewayManagementApi SDK results in an invalid StepFunction state resource


tvanhens:
Example:

```
$AWS.SDK.ApiGatewayManagementApi.postToConnection(
  {
    ConnectionId: item.connectionId.S,
    Data: event.body.message,
  },
  {
    iam: {
      actions: ["execute-api:ManageConnections"],
      resources: ["*"], // look away!
    },
  }
);
```

Error Message on Deployment:

```
Resource handler returned message: "Invalid State Machine Definition: 'SCHEMA_VALIDATION_FAILED: The resource provided arn:aws:states:::aws-sdk:apigatewaymanagementapi:postToConnection is not recognized. The value is not a valid resource ARN, or the resource is not available in this region. at /States/1__$AWS.SDK.ApiGatewayManagementApi.postToConnection({ConnectionId: item.co/Resource' (Service: AWSStepFunctions; Status Code: 400; Error Code: InvalidDefinition; Request ID: 82f4c85c-9448-421c-bca4-a953d97413cf; Proxy: null)" (RequestToken: fef30460-838b-8321-a869-ebadee0ff2ab, HandlerErrorCode: InvalidRequest)
```

sam-goodwin:
Damn looks like AWS does not support it https://docs.aws.amazon.com/step-functions/latest/dg/supported-services-awssdk.html#supported-services-awssdk-list

tvanhens:
So I don't think that is quite true. This is kind of an annoying edge case due to how weird the API Gateway APIs are compared to the rest of AWS.

Connection management actually uses the standard `execute-api` pathway for control operations [see my connection file](https://github.com/tvanhens/shoutlog/blob/master/src/connection.ts#L11-L17).

Step Functions should, in theory, allow this operation with the native API Gateway integration:

https://docs.aws.amazon.com/step-functions/latest/dg/connect-api-gateway.html

Not sure if there is an escape hatch I can use to test this easily.

sam-goodwin:
Oh interesting, that gives me hope then! We should track implementation of the "optimized integrations" for Step Functions - we forgot about them!

RE: escape hatch, you can use the `makeIntegration` helper to implement your own ASL integration.

Here's a link to a relatively complicated one https://github.com/functionless/functionless/blob/main/src/aws.ts#L404-L476

And here's the generic one: https://github.com/functionless/functionless/blob/main/src/aws-sdk/sdk.ts#L67-L114

[...]

PR #456, opened 2022-08-30
fix(step-functions): fail gracefully on unsupported APIs

### LINKED ISSUE #454: Step Functions: AWS SDK Integration should fail with friendly message for invalid service calls
For all unsupported and deprecated calls, fail gracefully.

https://docs.aws.amazon.com/step-functions/latest/dg/connect-ddb.html

thantos:
Fixes #454 
Partial #443

Friendly error when unsupported SFN integration APIs are used.

https://docs.aws.amazon.com/step-functions/latest/dg/supported-services-awssdk.html

### DIFF
--- a/src/aws-sdk/sdk.ts
+++ b/src/aws-sdk/sdk.ts
@@ -59,13 +59,22 @@ class ServiceProxy {
 }
 
 /**
+ * Step Functions AWS Service Integrations support integrations with many services, however, it has a list of
+ * services or methods that are not supported.
+ *
  * https://docs.aws.amazon.com/step-functions/latest/dg/supported-services-awssdk.html
  */
 const SFN_SERVICE_BLOCK_LIST: {
   [key in keyof TSDK]?: true | (keyof TSDK[key])[];
 } = {
-  // note: when lambda is supported by $AWS.SDK
-  // block list this API: https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/ApiGatewayManagementApi.html#postToConnection-property
+  /**
+   * Api Gateway Management API is just a wrapper around the Invoke API.
+   * Use the API Gateway invoke integration instead.
+   * https://github.com/functionless/functionless/issues/443
+   *
+   * note: when lambda is supported by $AWS.SDK block list this API:
+   *       https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/ApiGatewayManagementApi.html#postToConnection-property
+   */
   ApiGatewayManagementApi: true,
   Discovery: ["describeExportConfigurations"],
   CodeDeploy: [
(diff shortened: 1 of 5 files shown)


src/aws-sdk/sdk.ts
```diff
@@ -58,6 +58,55 @@ class ServiceProxy {
   }
 }
 
+/**
+ * Step Functions AWS Service Integrations support integrations with many services, however, it has a list of
+ * services or methods that are not supported.
+ *
+ * https://docs.aws.amazon.com/step-functions/latest/dg/supported-services-awssdk.html
+ */
```
sam-goodwin, line 66: What happens when they add support for one? Are we then blocking users?

test/step-function.test.ts
```diff
@@ -1519,6 +1519,37 @@ test("return AWS.SDK.CloudWatch.describeAlarms dynamic parameters", () => {
...
+          {
+            ConnectionId: "blah",
+            Data: "some data",
+          },
+          {
+            iam: {
+              resources: ["*"],
+            },
+          }
+        );
```
sam-goodwin, line 1540: I thought we were going to polyfill this? Or are we planning to implement a different API call?

src/aws-sdk/sdk.ts
thantos: Yeah, we could add an option to not validate.

sam-goodwin: Maybe these should be warnings, not blocking errors?

test/step-function.test.ts
thantos: I was starting to look at implementing `API.Invoke`. We can add a convenience method for connections along with it, maybe.

`API.postConnection`
`API.getConnection`
`API.deleteConnection`

src/aws-sdk/sdk.ts
```diff
@@ -58,6 +58,55 @@ class ServiceProxy {
...
+  EFS: ["createTags"],
+  ElasticTranscoder: ["testRole"],
+  EMR: ["describeJobFlows"],
+  Iot: [
+    "attachPrincipalPolicy",
+    "listPrincipalPolicies",
+    "detachPrincipalPolicy",
+    "listPolicyPrincipals",
+  ],
+  Lambda: ["invokeAsync"],
```
sam-goodwin, line 102: Oh no, this isn't supported? That seems ... so arbitrary ...

thantos: Its supported via the optimized integrations. Request/Response on a lambda.

> It is also possible to call Lambda asynchronously using the InvocationType parameter, as seen in the following example:

sam-goodwin: Is it possible to polyfill or is the interface botched?

test/step-function.test.ts
thantos: To track.
https://github.com/functionless/functionless/issues/457

sam-goodwin (approved):

PR #459, opened 2022-08-30
fix: support PropAccessExpr and ElementAccessExpr for .DynamoDB.*

### LINKED ISSUE #458: Bug: getTableArgument does not work with PropAccessExpr
Moving to SWC introduced a regression with ReferenceExpr - the ts-transformer used to transform expressions such as PropAccessExpr into a ReferenceExpr if the root was pointing to a free variable.

```ts
const a;
() => {
  a.b; // used to be ReferenceExpr(() => a.b),
}
```

SWC made it so the ReferenceExpr only took the place of an Identiifer, i.e. in this case `PropAccessExpr(ReferenceExpr(() => a), "b")`

Upon updating the samples to SWC, the saga-function broke because getTableArgument assumes the previous behavior: https://github.com/functionless/functionless-samples/blob/main/packages/sagaFunction/src/saga-function.ts#L45

```
SynthError: First argument into putItem should be an input with a property 'Table' that is a Table.

https://functionless.org/docs/error-codes#expected-an-object-literal
    at getTableArgument (/Users/samgoodwin/workspaces/functionless-samples/node_modules/functionless/src/aws.ts:700:8)
    at Object.bind (/Users/samgoodwin/workspaces/functionless-samples/node_modules/functionless/src/aws.ts:634:23)
    at /Users/samgoodwin/workspaces/functionless-samples/node_modules/functionless/src/function.ts:673:15
    at Array.flatMap (<anonymous>)
    at new Function (/Users/samgoodwin/workspaces/functionless-samples/node_modules/functionless/src/function.ts:670:52)
    at new SagaFunction (/Users/samgoodwin/workspaces/functionless-samples/packages/sagaFunction/src/saga-function.ts:25:33)
    at Object.<anonymous> (/Users/samgoodwin/workspaces/functionless-samples/packages/sagaFunction/src/app.ts:22:22)
    at Module._compile (node:internal/modules/cjs/loader:1103:14)
    at Module._compile (/Users/samgoodwin/workspaces/functionless-samples/node_modules/pirates/lib/index.js:136:24)
    at Module._extensions..js (node:internal/modules/cjs/loader:1157:10) {
  code: { code: 10012, type: 'ERROR', title: 'Expected an object literal' }
}
```

sam-goodwin:
Fixes #458

### DIFF
--- a/src/aws.ts
+++ b/src/aws.ts
@@ -39,6 +39,7 @@ import {
   IntegrationCall,
   IntegrationInput,
   makeIntegration,
+  tryFindIntegration,
 } from "./integration";
 import { AnyTable, isTable, ITable } from "./table";
 import type { AnyAsyncFunction } from "./util";
@@ -687,16 +688,8 @@ function getTableArgument(op: string, args: Argument[] | Expr[]) {
     );
   }
 
-  const tableRef = tableProp.expr;
+  const table = tryFindIntegration(tableProp.expr);
 
-  if (!isReferenceExpr(tableRef)) {
-    throw new SynthError(
-      ErrorCodes.Expected_an_object_literal,
-      `First argument into ${op} should be an input with a property 'Table' that is a Table.`
-    );
-  }
-
-  const table = tableRef.ref();
   if (!isTable(table)) {
     throw Error(`'Table' argument should be a Table object.`);
   }
--- a/test/function.test.ts
+++ b/test/function.test.ts
@@ -441,3 +441,24 @@ test("$AWS.DynamoDB.* with non-ReferenceExpr Table property", () => {
     }
   );
 });
+
+test("$AWS.DynamoDB.* with ShorthandPropAssign", () => {
+  const Table = table;
+  new Function(
+    stack,
+    "$AWS.DynamoDB.* with non-ReferenceExpr Table property",
+    async () => {
+      await $AWS.DynamoDB.PutItem({
+        Table,
+        Item: {
+          id: {
+            S: "key",
+          },
+          name: {
+            S: "name",
+          },
+        },
+      });
+    }
+  );
+});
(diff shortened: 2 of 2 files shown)


src/aws.ts
```diff
@@ -687,16 +688,8 @@ function getTableArgument(op: string, args: Argument[] | Expr[]) {
...
+  const table = tryFindIntegration(tableProp.expr);
 
-  if (!isReferenceExpr(tableRef)) {
-    throw new SynthError(
-      ErrorCodes.Expected_an_object_literal,
-      `First argument into ${op} should be an input with a property 'Table' that is a Table.`
-    );
-  }
-
-  const table = tableRef.ref();
```
thantos, line 699: Why don't you just use `evalToConstant`? and then check if it is defined and a `Table`.

thantos: I really don't like `tryFindIntegration`, its better to evaluate than to search. 

For example they may do something like `table.tableArn` and you'd end up returning the `table` part, not the `targetArn` part, which should fail.

thantos (changes requested):

src/aws.ts
sam-goodwin: `evalToCostant` doesn't return a `Table`.

thantos: hmm, I think you are right, we filter out only primitives.

Does `tryFind` search or does it evaluate down to a free variable that is an integration?

sam-goodwin: It does evaluate, following reference expressions and applying expressions to extract the actual value. Maybe poorly named but built for this purpose. We can improve later, this is an easy unblocking fix.

thantos: I hope you are right because of how we use it... but this block concerns me

https://github.com/functionless/functionless/blob/main/src/util.ts#L289

thantos (approved):

src/aws.ts
sam-goodwin: That's evalToConstant?

PR #462, opened 2022-08-30
feat(step-functions): number and string coerce

thantos:
```ts
Number("1") // 1
+"1" // 1
String(1) // "1"
```


src/asl/synth.ts
```diff
@@ -1894,12 +1894,77 @@ export class ASL {
               return { value: !!argOutput.value, containsJsonPath: false };
             }
           });
+        } else if (ref === String) {
+          const [arg] = expr.args;
```
sam-goodwin, line 1898: Do you want to check for `String(undefined)`, i.e. where `expr.args` is an `UndefinedLiteralExpr`?

src/asl/synth.ts
```diff
@@ -1894,12 +1894,77 @@ export class ASL {
...
+            if (ASLGraph.isJsonPath(argOutput) || argOutput.containsJsonPath) {
+              const temp = this.newHeapVariable();
+              const jsonPathStates = this.normalizeOutputToJsonPath(argOutput);
+              const { jsonPath } = ASLGraph.isJsonPath(jsonPathStates)
+                ? jsonPathStates
+                : jsonPathStates.output;
+
+              return {
+                ...ASLGraph.joinSubStates(arg, jsonPathStates, {
+                  // if this was a literal, we know it was an object, not a string
```
sam-goodwin, line 1917: why?

src/asl/synth.ts
```diff
@@ -1894,12 +1894,77 @@ export class ASL {
...
+                    checkString: {
+                      Type: "Choice",
+                      Choices: [{ ...ASL.isString(jsonPath), Next: "assign" }],
+                      Default: "format",
+                    },
+                    format: {
+                      Type: "Pass",
+                      Parameters: {
+                        "str.$": `States.JsonToString(${jsonPath})`,
+                      },
```
sam-goodwin, line 1932: Converting it to a JSON string?

src/asl/synth.ts
```diff
@@ -1894,12 +1894,77 @@ export class ASL {
...
+        } else if (ref === Number) {
+          const [arg] = expr.args;
+          return this.evalToNumber(arg?.expr);
         }
       }
       throw new Error(
-        `call must be a service call or list .slice, .map, .forEach or .filter: ${toStateName(
-          expr
-        )}`
+        `call must be an integration call, list(.slice, .map, .forEach, .filter, or ` +
```
sam-goodwin, line 1964: typeo, `list(`

src/asl/synth.ts
```diff
@@ -2470,6 +2536,139 @@ export class ASL {
...
+                  ),
+                  Next: "zero",
+                },
+                // Number(null) => 0
+                { ...ASL.isNull(output.jsonPath), Next: "zero" },
+                // Number("1") => 1
+                { ...ASL.isString(output.jsonPath), Next: "format" },
+                // Number(1) => 1
+                { ...ASL.isNumeric(output.jsonPath), Next: "assign" },
+                // Number(true) => 1
```
sam-goodwin, line 2598: This reads so well, great job. It's almost coding haha. We should expose it and document it. Great escape hatch DSL.

sam-goodwin (approved):

src/asl/synth.ts
thantos: `containsJsonPath` is only possible for an Object literal

src/asl/synth.ts
thantos: This worked as I expected for all of the tests... but maybe I'm missing something... I'll add more tests. I think the edge case will be that most objects just print `[object object]` instead of getting stringified.

src/asl/synth.ts
thantos: I meant to do that, was trying to show `list(.slice, .map, ...)`

src/asl/synth.ts
thantos: Thanks, love how easy it is.

src/asl/synth.ts
thantos: hmm

image

src/asl/synth.ts
thantos: Currently that will show `Step Functions does not support undefined`, but I can make it work at least.

src/asl/synth.ts
thantos: >                   // if this was a literal, we know it is an object (only objects have containsJsonPath:true), not a string

PR #471, opened 2022-09-02
chore(website): upgrade deps

thantos:


thantos:
Core library fails with typescript 4.8, rolling back to 4.7. Seems to work locally.

sam-goodwin:
> Core library fails with typescript 4.8, rolling back to 4.7. Seems to work locally.

Why does it fail? I'd hate to be locked into an older TS version because of website. Does it affect functionless or only the website's config?

thantos:
> > Core library fails with typescript 4.8, rolling back to 4.7. Seems to work locally.
> 
> Why does it fail? I'd hate to be locked into an older TS version because of website. Does it affect functionless or only the website's config?

No, our core library fails with TS 4.8. We'll need to upgrade to 4.8 and resolve the errors. Website was just the first thing I tried to upgrade to 4.8. Will try with core today.

thantos:
image

sam-goodwin:
> image

Oh, probably the type assertion functions modifying `this` type. Maybe those are way overkill and are slowing down compilation for very little value.

thantos:
> Oh, probably the type assertion functions modifying this type. Maybe those are way overkill and are slowing down compilation for very little value.

Not sure about the value, but definitely one of the main hotspots.

thantos:
Removing the `!` fixed it, ha

website/docs/concepts/terminology.md
```diff
@@ -75,7 +75,7 @@ An Interpreter is a process that runs during [Synthesis Time](#synthesis-time) t
 
 Examples include:
 
-- [Amazon States Language](../api/classes/ASL.md)
+- [Amazon States Language](../api/classes/ASL-1.md)
```
sam-goodwin, line 78: Why did this change

sam-goodwin (approved):

website/docs/concepts/terminology.md
thantos: I am not sure, version of the typedoc was upgraded, something must have changed.

PR #473, opened 2022-09-02
chore: upgrade typescript to 4.8, improve compile performance

thantos:
Fixes type issues which broke in TS 4.8.

Resolved a complex type that was causing an extra second of processing.

### Before
Total time:                  16.91s

```ts
    if (this.parent === undefined) {
      return undefined;
    } else if (is(this.parent)) {
      return this.parent;
    } else {
      return this.parent.findParent(is);
    }
```

### After
Total time:                  16.03s

```ts
    if (this.parent === undefined) {
      return undefined;
    } else if (is(this.parent)) {
      return this.parent as unknown as N;
    } else {
      return this.parent.findParent(is);
    }
```

Evaluating `Parent & {} & N` was taking a second

```
Hot Spots
├─ Check file /home/sussmans/functionless/src/node.ts (1741ms)
│  └─ Compare types 7794 and 1985 (1125ms)
│     └─ Compare types 7794 and 9 (1125ms)
│        ├─ {"id":7794,"kind":"Intersection","count":3,"types":[838,33,1979]}
│        │  ├─ {"id":838,"kind":"TypeParameter","name":"Parent","location":{"path":"/home/sussmans/functionless/src/node.ts","line":76,"char":3}}
│        │  ├─ {"id":33,"kind":"Other","display":"{}"}
│        │  └─ {"id":1979,"kind":"TypeParameter","name":"N","location":{"path":"/home/sussmans/functionless/src/node.ts","line":182,"char":21}}
│        └─ {"id":9,"kind":"Intrinsic","name":"undefined"}
└─ Check file /home/sussmans/functionless/src/asl/synth.ts (1267ms)
```

### DIFF
--- a/src/expression.ts
+++ b/src/expression.ts
@@ -771,7 +771,10 @@ export class TaggedTemplateExpr extends BaseExpr<NodeKind.TaggedTemplateExpr> {
  * // TemplateHead is "".
  * ```
  */
-export class TemplateHead extends BaseNode<NodeKind.TemplateHead> {
+export class TemplateHead extends BaseNode<
+  NodeKind.TemplateHead,
+  TemplateExpr
+> {
   readonly nodeKind = "Node";
 
   constructor(
@@ -795,7 +798,7 @@ export class TemplateHead extends BaseNode<NodeKind.TemplateHead> {
  */
 export class TemplateSpan<
   Literal extends TemplateMiddle | TemplateTail = TemplateMiddle | TemplateTail
-> extends BaseNode<NodeKind.TemplateSpan> {
+> extends BaseNode<NodeKind.TemplateSpan, TemplateExpr> {
   readonly nodeKind = "Node";
 
   constructor(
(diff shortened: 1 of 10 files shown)


thantos:
woops, didn't mean to grab the website change too, fixing...

sam-goodwin (approved):

sam-goodwin:
Fix the pr title?

src/expression.ts
```diff
@@ -815,7 +818,10 @@ export class TemplateSpan<
   }
 }
 
-export class TemplateMiddle extends BaseNode<NodeKind.TemplateMiddle> {
+export class TemplateMiddle extends BaseNode<
+  NodeKind.TemplateMiddle,
+  TemplateSpan
+> {
```
sam-goodwin, line 824: Default types don't work anymore?

src/node.ts
```diff
@@ -57,7 +56,6 @@ export type FunctionlessNode =
   | Err
   | BindingPattern
   | ImportKeyword
-  | NoSubstitutionTemplateLiteralExpr
```
sam-goodwin, line 60: Why did you remove this

src/expression.ts
thantos: Trying to reduce the complexity by specifying types were possible. TemplateMiddle can only be in a TemplateSpan

src/node.ts
thantos: Its also in `Expr`.

src/expression.ts
sam-goodwin: Oh right - good point. This is more precise.

PR #474, opened 2022-09-03
chore: support testing on AWS

### LINKED ISSUE #490: Step Functions: implement `?.` for undefined literals
`?.` on an undefined literal will currently throw `Undefined literal is not supported`.

### LINKED ISSUE #491: Bug: Step Functions truthy logic on an object return undefined.
```ts
if(!maybeObject) { // returns true

}
```

thantos:
Closes #490 Closes #491

* Adds the ability to run all of the localstack tests on AWS.
* renamed `localstack` to `runtime` in various places
* Added the ability to optionally destroy stacks after tests (useful for CI tests)
* Stack names are now based on the github `ref` ENV Variable when present.
* If all localstack tests fail, skip, or not `only`, deployment is skipped.
* Added assume credentials step to the build and release workflows.
* Assume credentials uses the github OIDC connector to grant access without using secrets. This role is defined by a CDK stack in the private `aws_org` repo.

* [x] Step Functions tests running concurrntly
* [x] Build on github succeeding
* [x] Clean up disgusting runtime test code
* [x] Document testing options
* [x] increase lambda concurrency from 10 to 1000 on CI account 
* [x] Self-destructing stack
* [x] Restore localstack code locally


thantos:
I expect all of this to explode for many reasons...

thantos:
Good to know, if a github action is in valid, it automatically passes!

sam-goodwin:
This change is top priority now - our tests have ground us to a halt.

.projenrc.ts
```diff
@@ -69,6 +71,26 @@ class CustomTypescriptProject extends typescript.TypeScriptProject {
...
+    return [
+      ...super.renderWorkflowSetup(options),
+      // https://github.com/aws-actions/configure-aws-credentials#sample-iam-role-cloudformation-template
+      // the aws-org stacks create an OIDC provider for github.
+      {
+        name: "Configure AWS Credentials",
+        uses: "aws-actions/configure-aws-credentials@v1",
+        with: {
+          "role-to-assume":
+            "arn:aws:iam::593491530938:role/githubActionStack-githubactionroleA106E4DC-14SHKLVA61IN4",
```
sam-goodwin, line 87: How does github actions authenticate to assume this role?

.projenrc.ts
```diff
@@ -218,25 +239,44 @@ project.compileTask.prependExec(
...
-project.testTask.prependExec("./scripts/localstack");
-project.testTask.exec("localstack stop");
+// project.testTask.prependExec("./scripts/localstack");
+// project.testTask.exec("localstack stop");
 
 project.testTask.env("NODE_OPTIONS", "--max-old-space-size=4096");
-project.testTask.env("DEFAULT_REGION", "ap-northeast-1");
-project.testTask.env("AWS_ACCOUNT_ID", "000000000000");
-project.testTask.env("AWS_ACCESS_KEY_ID", "test");
-project.testTask.env("AWS_SECRET_ACCESS_KEY", "test");
```
sam-goodwin, line 233: What were these doing before?

.projenrc.ts
```diff
@@ -282,4 +322,6 @@ project.eslint!.addOverride({
 project.prettier!.addIgnorePattern("coverage");
 project.prettier!.addIgnorePattern("lib");
 
+project.buildWorkflow?.addPostBuildSteps;
```
sam-goodwin, line 325: Did you plan on doing something here?

CONTRIBUTING.md
```diff
@@ -93,6 +93,13 @@ test-app$ yarn install
 test-app$ yarn watch
 ```
 
+### Test Options
+
+| Env                 | Description                                                                                                           |
+| ------------------- | --------------------------------------------------------------------------------------------------------------------- |
+| TEST_DEPLOY_TARGET  | Deploy to `AWS` (with default credential chain) or `LOCALSTACK`, Default: `LOCALSTACK`                                |
+| TEST_DESTROY_STACKS | When truthy, will destroy any stacks created when the runtime test suite completes, Default: False (unless `CI=true`) |
```
sam-goodwin, line 101: Curious if we can trigger this when the PR is closed? That way new commits trigger updates to the stack? Or is that too risky? What's the plan for isolation?

test/eventbus.localstack.test.ts
```diff
@@ -63,6 +63,8 @@ localstackTestSuite("eventBusStack", (testResource) => {
           });
         }
       );
+      bus.resource.grantPutEventsTo(role);
+      table.resource.grantReadData(role);
```
sam-goodwin, line 67: Oh do we now need to grant permissions to the role used by the tests?

sam-goodwin: Not necessary for this iteration - but perhaps we could use integrations here? Output integration which produces a client on the other side.

.projenrc.ts
thantos: OIDC uses github itself as the identity and in the role we assert which repos can assume the role.

```ts
    const provider = new OpenIdConnectProvider(this, "githubProvider", {
      url: "https://token.actions.githubusercontent.com",
      clientIds: ["sts.amazonaws.com"],
      thumbprints: ["6938fd4d98bab03faadb97b34396831e3780aea1"],
    });
    const principal = new OpenIdConnectPrincipal(provider, {
      StringLike: {
        "token.actions.githubusercontent.com:sub": "repo:functionless/*",
      },
    });
```

.projenrc.ts
thantos: setting the env creds to hit localstack.

CONTRIBUTING.md
thantos: I was thinking of using the self-destructing stack (via step functions) where a machine starts and decided if it should destruct the stack based on the last deployment time. That way, an active PR will maintain it's stack and a stale one will not use up resources (with a higher cold start time).

Going to play with this today.

test/eventbus.localstack.test.ts
thantos: We totally could generate the test script using the serializer and integrations. Would be an interesting experiment. 

But yeah, the new annoyance is that our tests will need to make real AWS calls using a real role that actually needs permissions to make the calls.

The other option is that we make catch all roles for the tests, but I like the idea of being least privilege to set a good example. 

Luckily, the core resource like sfn and lambda can be wired using the core test logic, any extra resources will need to grant explicitly (until we start to use integrations).

thantos: https://github.com/functionless/functionless/issues/485

thantos:
Tests are failing due to Lambda concurrent invocations limits on the CI AWS account. @sam-goodwin will need to up the limit to progress.

.projenrc.ts
sam-goodwin: That is cool!

CONTRIBUTING.md
sam-goodwin: What would the timeout be on that step function?

thantos: The default I have right now is 12 hours, but we can adjust to anything under a year.

.projenrc.ts
sam-goodwin: Did you deploy that OIDC policy with CloudFormation?

test/function.localstack.test.ts
```diff
@@ -51,7 +54,11 @@ interface TestFunctionBase {
     expected: OO extends void
       ? null
       : OO | ((context: Outputs) => OO extends void ? null : O),
-    payload?: I | ((context: Outputs) => I)
+    payload?: I | ((context: Outputs) => I),
+    /**
+     * When true uses jest's {@link jest.Matchers.toMatchObject} logic instead of {@link jest.Matchers.toEqual}.
+     */
+    matchObject?: boolean
```
sam-goodwin, line 61: Could pass the reference to the function instead.

test/self-destructor.ts
```diff
@@ -0,0 +1,123 @@
...
+export interface SelfDestructorProps extends SelfDestructorMachineProps {}
+
+export interface SelfDestructorMachineProps {
+  /**
+   * Default: 43200 (12 hours)
+   */
+  selfDestructAfterSeconds?: number;
+}
+
+export class SelfDestructor extends Construct {
```
sam-goodwin, line 19: Document

test/self-destructor.ts
```diff
@@ -0,0 +1,123 @@
...
+    const dateAddSeconds = new Function(
+      this,
+      "timestampOffsetBeforeCurrent",
+      async (input: {
+        timestamp: Date | string;
+        offsetSeconds: number;
+      }): Promise<string> => {
+        const date = new Date(input.timestamp);
+        date.setSeconds(date.getSeconds() + input.offsetSeconds);
+        return date.toISOString();
```
sam-goodwin, line 45: Nit: create a new Date

test/self-destructor.ts
```diff
@@ -0,0 +1,123 @@
...
+          const offsetTime = await dateAddSeconds({
+            timestamp: stackChangeTime,
+            offsetSeconds: input.selfDestructAfterSeconds ?? 43200,
+          });
+
+          // set the base time to the destruct time, if these are the same after selfDestructAfterSeconds, destroy the stack
+          destructTime = stackChangeTime as unknown as string;
+
+          // wait until selfDestructAfterSeconds + stackChangeTime
+          $SFN.waitUntil(offsetTime);
```
sam-goodwin, line 100: I still wish this was an await. It's always jarring

sam-goodwin (approved):

test/function.localstack.test.ts
thantos: Ended up just repurposing the expected method to accept the payload and run arbitrary logic.

```ts
expected: T | (context, payload) => void
```

Where a value will just use `.toEquals`

test/self-destructor.ts
thantos: though could be convinced otherwise...

Durable functions uses await on timers.

https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-timers?tabs=csharp#usage-for-delay

So does temporal

https://docs.temporal.io/workflows/#intrinsic-non-deterministic-logic

PR #482, opened 2022-09-06
feat: compiler register hook

tvanhens:
Fixes #


register.js
```diff
@@ -0,0 +1,26 @@
...
+    parser: {
+      syntax: "typescript",
+      dynamicImport: false,
+      decorators: false,
+      hidden: {
+        jest: true,
+      },
+    },
+    transform: null,
+    target: "es2021",
```
sam-goodwin, line 14: How did we choose this? Any reason not to jump straight to 2022?

sam-goodwin (approved):

register.js
tvanhens: I cargo-culted it from the `test-app` directory like the grown up developer I am 🤣 

I'll update it to es2022

register.js
```diff
@@ -0,0 +1,26 @@
+const register = require("@swc/register/lib/node").default;
+
+register({
```
sam-goodwin, line 3: Will we need to allow users to specify any of these configuration properties?

sam-goodwin (approved):

register.js
tvanhens: Yes, but breaking this into a separate file and allowing the config to be overwritten seems to be causing the error I was running into yesterday. As it stands, users can still overwrite the config by including a `.swcrc`.

tvanhens: My inclination is to get the critical path in place first to unblock the getting started section of the website and then come back and figure out the problem with making this composable. Doing so is a non-breaking change with the hook in place so I don't think it makes sense to block finishing the getting started guide.

test-app/.swcrc
```diff
@@ -1,29 +0,0 @@
-{
```
sam-goodwin, line 1: Did you forget to remove the root `.swcrc`?

thantos (approved):

test-app/.swcrc
tvanhens: It's still needed for testing. Will remove when I add a jest loader.

PR #483, opened 2022-09-08
chore: update reference to $SFN symbol in docs

Grundlefleck:
Minor fix to Step Function documentation. I was following along and couldn't find the `SFN` symbol. Presume it was renamed to `$SFN` at some point and docs were not updated.


thantos (approved):
Nope, its always been `$SFN`, just a typo, thanks! (Note: we need to typecheck our docs)

PR #488, opened 2022-09-09
feat: step functions retry

### LINKED ISSUE #486: Step Functions: Retry
```ts
$SFN.retry(async () => {
   ....
})
```

Will create a parallel task with a single branch.

Limits:

1. Cannot update mutable variables outside of the parallel/retry, like the `$SFN.map` and `$SFN.parallel` functions.
2. will re-run any of the expressions in the parallel, adding some cost over just retrying the task.

thantos:
Closes #486 

```ts
$SFN.retry(async () => {
   return await func();
});

$SFN.retry([{ ErrorEquals: ["MyError"], MaxAttempts: 2, IntervalSeconds: 4, BackoffRate: 1 }], async () => {
   throw new StepFunctionError("MyError", "cause");
});
```

### DIFF
--- a/src/step-function.ts
+++ b/src/step-function.ts
@@ -544,17 +544,17 @@ export namespace $SFN {
           };
         }
 
-        const temp = context.newHeapVariable();
+        const heap = context.newHeapVariable();
 
         return {
           Type: "Parallel",
           Branches: [context.aslGraphToStates(functionStates)],
           Next: ASLGraph.DeferNext,
           Retry: retryValues,
-          ResultPath: temp,
+          ResultPath: heap,
           Parameters: context.cloneLexicalScopeParameters(body.expr),
           output: {
-            jsonPath: `${temp}[0]`,
+            jsonPath: `${heap}[0]`,
           },
         };
 
@@ -575,7 +575,7 @@ export namespace $SFN {
             retryValue.constant.length === 0
           ) {
             throw new SynthError(
-              ErrorCodes.Invalid_Input,
+              ErrorCodes.Step_Function_Retry_Invalid_Input,
               "Expected $SFN.retry retry argument to be an array literal which contains only constant values."
             );
           }
@@ -596,14 +596,14 @@ export namespace $SFN {
 
           if (!ErrorEquals || ErrorEquals.some((x) => typeof x !== "string")) {
             throw new SynthError(
-              ErrorCodes.Invalid_Input,
+              ErrorCodes.Step_Function_Retry_Invalid_Input,
               "$SFN.retry retry object must be an array literal of string with at least one value"
             );
           }
 
           if (Object.keys(rest).length !== 0) {
             throw new SynthError(
-              ErrorCodes.Invalid_Input,
+              ErrorCodes.Step_Function_Retry_Invalid_Input,
               "$SFN.retry retry object must only contain keys: ErrorEquals, BackoffRate, IntervalSeconds, or MaxAttempts"
             );
           }
--- a/src/step-function.ts
+++ b/src/step-function.ts
@@ -10,7 +10,7 @@ import { StepFunctions } from "aws-sdk";
 import { Construct } from "constructs";
 import { ApiGatewayVtlIntegration } from "./api";
 import { AppSyncVtlIntegration } from "./appsync";
-import { ASL, ASLGraph, StateMachine, States } from "./asl";
+import { ASL, ASLGraph, Retry, StateMachine, States } from "./asl";
 import { assertDefined } from "./assert";
 import { FunctionLike } from "./declaration";
 import { ErrorCodes, SynthError } from "./error-code";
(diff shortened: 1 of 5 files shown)


src/step-function.ts
```diff
@@ -466,6 +465,152 @@ export namespace $SFN {
...
+            ErrorCodes.Invalid_Input,
+            "Expected $SFN.retry callback argument to be a function."
+          );
+        }
+
+        const retryValues = processRetries(retries?.expr);
+
+        const functionStates = context.evalStmt(body.expr.body, {
+          End: true,
+          ResultPath: "$",
```
sam-goodwin, line 533: What does this do?

src/step-function.ts
```diff
@@ -466,6 +465,152 @@ export namespace $SFN {
...
+   *
+   * Caveat: `$SFN.retry` does not maintain the state of mutated variables within the callback.
+   *
+   * ```ts
+   * let a = 1;
+   * await $SFN.retry(() => { a = 2; return a * 4; })
+   * return a;
+   * ```
+   *
+   * The return value will be `1`.
```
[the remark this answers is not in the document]
sam-goodwin: Oh is this because you're using a Parallel?

src/step-function.ts
```diff
@@ -466,6 +465,152 @@ export namespace $SFN {
...
+
+        const functionStates = context.evalStmt(body.expr.body, {
+          End: true,
+          ResultPath: "$",
+        });
+
+        /**
+         * If there are no states for the function, just return an empty state, nothing to do here.
+         */
+        if (!functionStates) {
```
sam-goodwin, line 539: What if there are no states after optimization?

src/step-function.ts
```diff
@@ -466,6 +465,152 @@ export namespace $SFN {
...
+          return {
+            Type: "Pass",
+            Next: ASLGraph.DeferNext,
+            output: { value: context.context.null, containsJsonPath: false },
+          };
+        }
+
+        const temp = context.newHeapVariable();
```
sam-goodwin, line 547: nit: better name for `temp`?  Is it temp? It seems to the be the result of the inner block

src/step-function.ts
```diff
@@ -466,6 +465,152 @@ export namespace $SFN {
...
+          const retryValue = evalToConstant(expr);
+
+          if (
+            !retryValue ||
+            !Array.isArray(retryValue.constant) ||
+            retryValue.constant.length === 0
+          ) {
+            throw new SynthError(
+              ErrorCodes.Invalid_Input,
+              "Expected $SFN.retry retry argument to be an array literal which contains only constant values."
```
sam-goodwin, line 579: The error message doesn't explain that the array can't be empty.

I think this case deserves its own ErrorCode with explanation.

src/step-function.ts
```diff
@@ -466,6 +465,152 @@ export namespace $SFN {
...
+              ErrorCodes.Invalid_Input,
+              "$SFN.retry retry object must be an array literal of string with at least one value"
+            );
+          }
+
+          if (Object.keys(rest).length !== 0) {
+            throw new SynthError(
+              ErrorCodes.Invalid_Input,
+              "$SFN.retry retry object must only contain keys: ErrorEquals, BackoffRate, IntervalSeconds, or MaxAttempts"
+            );
```
sam-goodwin, line 608: Let's create an ErrorCode for this. Lots to explain.

sam-goodwin (changes requested):

src/step-function.ts
```diff
@@ -466,6 +465,152 @@ export namespace $SFN {
...
+  }
+
+  /**
+   * Apply Step Function's built in Retry logic to a function.
+   *
+   * ```ts
+   * new ExpressStepFunction(this, "F", (id: string) => {
+   *   const results = $SFN.retry(
+   *     async () => { await task1(id) }
+   *   )
```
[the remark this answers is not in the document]
thantos: It puts everything into a Parallel, so yeah, any error in the body will be used (and then the error matchers are applied)

src/step-function.ts
thantos: Determines the behavior of `return`. In our case, set the return value to be the whole state and end (return from the parallel).

src/step-function.ts
thantos: Currently the optimization never removes the `StartAt` state. In the future it may be smart enough to remove Map and Parallel states that can be empty.

src/step-function.ts
thantos: its the heap variable we return to and the future states can use to refer to this output. I had been calling them `heap`, but have started to use `temp`. Will update

src/step-function.ts
thantos: I have been trying to reduce error codes related to specific integrations, but I guess this is a common enough case.

src/step-function.ts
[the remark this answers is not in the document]
thantos: yeah. $SFN.map and $SFN.Parallel have the same limitation.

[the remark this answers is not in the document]
thantos: I've thought of a few ways to do this, but they all fall over if the mutations need to be reflected through retry attempts. 

I added a few options in #486 , seems like if we had multiplication abilities we would implement our own retry at the cost of additional transitions and states.

For now we have this limit.

sam-goodwin (approved):

PR #494, opened 2022-09-12
fix: ForOfStmt, ForInStmt, ForStmt, WhileStmt, DoStmt ES conformance

sam-goodwin:
Partial https://github.com/functionless/functionless/issues/374

- [x] change `block: BlockStmt` to `stmt: Stmt` - our prior AST was too narrow, requiring everything be a BlockSmt
```ts
while (true) <stmt>
``` 
- [x] support optional chaining on calls
```ts
call.?(..args)
```
- [x] Update `initializer` in for-loops to allow `VariableDeclList` - this is because `const | let | var` is stored on the `VariableDeclList`, not a `VariableDecl`. For backwards compatibility purposes, we will continue to allow VariableDecl until ast-reflection is updated to always write VariableDeclList.

### DIFF
--- a/src/asl/synth.ts
+++ b/src/asl/synth.ts
@@ -558,7 +558,7 @@ export class ASL {
       }
     } else if (isForOfStmt(stmt) || isForInStmt(stmt)) {
       return this.evalExprToSubState(stmt.expr, (output) => {
-        const body = this.evalStmt(stmt.body, returnPass);
+        const body = this.evalStmt(stmt.stmt, returnPass);
 
         // assigns either a constant or json path to a new variable
         const assignTempState = this.assignValue(undefined, output);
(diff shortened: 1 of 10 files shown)


thantos (approved):

src/asl/synth.ts
```diff
@@ -622,9 +622,21 @@ export class ASL {
...
+              : isIdentifier(stmt.initializer)
+              ? this.evalAssignment(stmt.initializer, {
+                  jsonPath: `${tempArrayPath}[0]`,
+                })!
+              : (() => {
+                  throw new SynthError(
+                    ErrorCodes.Unsupported_Feature,
+                    `expression ${stmt.initializer.nodeKind} is not supported as the initializer in a ForInStmt`
+                  );
+                })();
```
sam-goodwin, line 639: @thantos thoughts on this?

PR #505, opened 2022-09-14
chore: simplify tests

thantos:
1. reduce the number of resources created by the lambda and step functions tests
   1. Create common resources when the behavior is shared across tests
   2. Share a common resource role when the test does not involve granting inferred IAM policies
   3. Merged some SFN test cases
   4. Changes in resources created
      1. Step Functions 194 -> 88
      2. Lambda 109 -> 66
   5. Changes in roles created (CI account was hitting the 1000 role limit)
      1. Step Functions 111 -> 33
      2. Lambda 56 -> 32 
2. Make use of the functional resource creation pattern in the tests
3. Fixed a bug in step function templates where the machine failed on special characters and added a test for it. 
    4. ```
       `{`\\${something}\\`}`
       ```

```ts
  const stringFunction = standardFunction<undefined, string>(
    stack,
    "stringFunction",
    async () => "hi"
  );
```

```ts
  function sfnTestLambda<I, O>(
    id: string,
    closure: FunctionClosure<I, O>
  ): Function<I, O>;
  function sfnTestLambda<I, O>(
    scope: Construct,
    id: string,
    closure: FunctionClosure<I, O>
  ): Function<I, O>;
  function sfnTestLambda<I, O>(
    ...args:
      | [id: string, closure: FunctionClosure<I, O>]
      | [scope: Construct, id: string, closure: FunctionClosure<I, O>]
  ) {
    const [scope, id, closure] =
      args.length === 2 ? [stack, args[0], args[1]] : args;
    return new Function<I, O>(scope, id, localstackClientConfig, closure);
  }


const lambda = sfnTestLambda("myLambda", async () => {...});
```


src/asl/synth.ts
```diff
@@ -1771,6 +1773,15 @@ export class ASL {
             jsonPath: `${tempHeap}.string`,
           },
         };
+
+        // https://docs.aws.amazon.com/step-functions/latest/dg/amazon-states-language-intrinsic-functions.html#amazon-states-language-intrinsic-functions-escapes
+        function escapeFormatString(literal: ASLGraph.LiteralValue) {
+          if (typeof literal.value === "string") {
+            return literal.value.replace(/[\}\{\'}]/g, "\\$&");
```
sam-goodwin, line 1780: I don't understand what this fixes

test/function.localstack.test.ts
```diff
@@ -152,14 +210,7 @@ runtimeTestSuite("functionStack", (testResource, _stack, _app) => {
...
         const val = "a";
-        return new Function(
-          parent,
-          "function",
-          {
-            timeout: Duration.seconds(20),
-          },
-          async () => val
-        );
+        return standardFunction(parent, async () => val);
```
sam-goodwin, line 213: Could be confused with standard step functions lol.

sam-goodwin: Much cleaner though

sam-goodwin: Maybe `functionWithSharedRole`

sam-goodwin (approved):
Great improvement

src/asl/synth.ts
thantos: Escaping special characters (see link above)


`{${a}}` => 
   Before - `States.Format('{{}}', $.a)` - fails because `{` and `}` are special characters and `{{}}` is invalid.
   After - `States.Format('\{{}\}', $.a)` - succeeds!  `{helloworld}`

test/function.localstack.test.ts
thantos: Ha, sure.

src/asl/synth.ts
sam-goodwin: How does that regex achieve that?

sam-goodwin: \\$& ???

thantos: Added a comment.

PR #515, opened 2022-09-17
feat(step-functions): object spread and fix bind in SFN map/foreach

### LINKED ISSUE #501: Bug: $SFN.ForEach bound argument uses the wrong source path
```json
        "ResultPath": "$.ResourceARN",
        "InputPath": "$$.Map.Item.Value['ResourceARN']"
```

Should use a parameter that is set to `$$.Map.Item.Value`.

thantos:
Fixes #501

* Object spread using JsonMerge
* `$SFN.ForEach` and `$SFN.Map`'s parameter binding used $$.Map.* which is not present in the body
* Improved intrinsic function rendering on ASLGraph
   * added helpers to render 
   * added specific interface and wrapper for `IntrinsicFunction`
   * Added algorithm which can avoid the limit of 10 intrinsic functions per expression by spreading the tree out over multiple states and Parameter properties.

```ts
const x = {
  ....anyObject
};
```

```ts
$SFN.ForEach(async ({ prop }) => { await func(prop) })
```

* [ ] Take a pass at cleaning up the code.
* [ ] Update validation


src/queue.ts
```diff
@@ -613,11 +616,13 @@ abstract class BaseQueue<Message>
                   },
                   ResultPath: heapVar,
                   Next: "send message batch",
-                  Iterator: {
-                    StartAt: "serialize Message",
-                    States: {
+                  Iterator: context.aslGraphToStates({
```
sam-goodwin, line 619: Why is this necessary? The inside of an iterator is an isolated state machine, no?

src/step-function.ts
```diff
@@ -370,7 +370,7 @@ export interface $SFN {
 /**
  * @see https://docs.aws.amazon.com/step-functions/latest/dg/amazon-states-language-intrinsic-functions.html#asl-intrsc-func-hash-calc
  */
-export type HashAlgorithm = "MD5" | "SHA-1" | "SHA-256" | "SHA-384" | "SHA-512";
+export type HashAlgorithm = ASLGraph.HashAlgorithm;
```
sam-goodwin, line 373: Why are there two types?

src/step-function.ts
```diff
@@ -800,15 +780,12 @@ export const $SFN = {
             assertLiteralStringOrJsonPath(
               algorithmOut,
               "States.Hash",
-              "algorithm"
+              "algorithm",
+              ["MD5", "SHA-1", "SHA-256", "SHA-384", "SHA-512"]
```
sam-goodwin, line 784: Should this be defined as a constant?

test/step-function.localstack.test.ts
```diff
@@ -70,6 +70,9 @@ runtimeTestSuite<
           res instanceof StepFunction ? [res, {}] : [res.sfn, res.outputs];
         funcRes.resource.grantStartExecution(role);
         funcRes.resource.grantRead(role);
+
+        console.log(JSON.stringify(res.definition, null, 2));
```
sam-goodwin, line 74: Leaving this in?

test/step-function.localstack.test.ts
```diff
@@ -448,6 +457,114 @@ runtimeTestSuite<
...
+          ...input,
+          ...input,
+          ...input,
+          ...input,
+          ...input,
+          ...input,
+          ...input,
+          ...input,
+          ...input,
+          ...{ x: 7, y: 8, z: 9, a: 1, m: 4 },
```
sam-goodwin, line 542: Do we want to support dynamic keys and values like we discussed with json format?

test/step-function.test.ts
```diff
@@ -398,6 +398,123 @@ test("spread constant array and object", () => {
   expect(normalizeDefinition(definition)).toMatchSnapshot();
 });
 
+test("depth", () => {
```
sam-goodwin, line 401: More descriptive test name?

sam-goodwin (changes requested):

src/queue.ts
thantos: I think state names need to be globally unique, can add a test, but it also worked better with the new intrinsic API.

src/step-function.ts
thantos: Didn't want to mix the more pubic facing one with the aslgraph one

test/step-function.localstack.test.ts
thantos: No, I'm a bit nervous about the formatting with Json paths so wanted to be leave it as a separate task.

src/queue.ts
sam-goodwin: I would be surprised if that were true.

sam-goodwin: You can't transition to states inside the iterator from outside

thantos: > States are elements in your state machine. A state is referred to by its name, which can be any string, but which must be unique within the scope of the entire state machine.

sam-goodwin: Did you test it?

src/step-function.ts
sam-goodwin: Problem would then be when the user jumps to the definition and has to go look at the private definition. I'd vote for having only the public interface

test/step-function.localstack.test.ts
sam-goodwin: What makes you nervous? Fine with it being a separate task - that's reasonable.

src/queue.ts
thantos: Looks like sendmessagebatch and receive were broken after one of my last PRs. (there were no tests)

But after fixing that...

```ts
      return new StepFunction(scope, "q", async () => {
        await queue.sendMessage({ MessageBody: { id: "hello" } });
        await queue.sendMessageBatch({
          Entries: [{ MessageBody: { id: "hello" }, Id: "1" }],
        });
        await queue.sendMessageBatch({
          Entries: [{ MessageBody: { id: "hello" }, Id: "2" }],
        });
      });
```

```
InvalidDefinition: Invalid State Machine Definition: 'DUPLICATE_STATE_NAME: Duplicate State name: serialize Message at /States/1__await queue.sendMessageBatch({Entries: [{MessageBody: {id: "hello"}, Id: 1/Iterator/States/serialize Message, DUPLICATE_STATE_NAME: Duplicate State name: unwrap Message at /States/1__await queue.sendMessageBatch({Entries: [{MessageBody: {id: "hello"}, Id: 1/Iterator/States/unwrap Message'
```

sam-goodwin: Cool, good to know. Seems silly to me as you would think scoping rules would apply. Thanks for checking and fixing.

test/step-function.localstack.test.ts
thantos: Finding more limits in the size of a parameters, formatting issues, maybe it'll be fine. Need to write some tests.

sam-goodwin (approved):

src/queue.ts
thantos: np, will add a new test with this PR.

sam-goodwin (approved):

sam-goodwin (approved):

PR #516, opened 2022-09-19
feat(step-functions): support string.split

thantos:
String split!

### DIFF
--- a/src/asl/synth.ts
+++ b/src/asl/synth.ts
@@ -1974,6 +1974,56 @@ export class ASL {
               : ASLGraph.intrinsicStringToJson(output)
           );
         });
+      } else if (isSplit(expr)) {
+        const [splitter, limit] = expr.args;
+
+        if (limit) {
+          throw new SynthError(
+            ErrorCodes.Unsupported_Feature,
+            "Step Functions does not support the limit parameter on String.split"
+          );
+        } else if (!splitter) {
+          throw new SynthError(
+            ErrorCodes.Invalid_Input,
+            "Step Functions String Split splitter argument is required"
+          );
+        }
+
+        return this.evalContext(expr, ({ evalExprToJsonPathOrLiteral }) => {
+          const valueOut = evalExprToJsonPathOrLiteral(expr.expr.expr);
+          const splitterOut = evalExprToJsonPathOrLiteral(splitter.expr);
+
+          if (ASLGraph.isLiteralValue(valueOut)) {
+            if (!ASLGraph.isLiteralString(valueOut)) {
+              throw new SynthError(
+                ErrorCodes.Invalid_Input,
+                "Step Functions String Split must be on a reference or string literal."
+              );
+            }
+          }
+
+          if (ASLGraph.isLiteralValue(splitterOut)) {
+            if (!ASLGraph.isLiteralString(splitterOut)) {
+              throw new SynthError(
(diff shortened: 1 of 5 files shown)


src/asl/synth.ts
```diff
@@ -1974,6 +1974,56 @@ export class ASL {
...
+          throw new SynthError(
+            ErrorCodes.Unsupported_Feature,
+            "Step Functions does not support the limit parameter on String.split"
+          );
+        } else if (!splitter) {
+          throw new SynthError(
+            ErrorCodes.Invalid_Input,
+            "Step Functions String Split splitter argument is required"
+          );
+        }
```
sam-goodwin, line 1990: Is it really called splitter and not delimiter? Haha

src/asl/synth.ts
```diff
@@ -1974,6 +1974,56 @@ export class ASL {
...
+
+        return this.evalContext(expr, ({ evalExprToJsonPathOrLiteral }) => {
+          const valueOut = evalExprToJsonPathOrLiteral(expr.expr.expr);
+          const splitterOut = evalExprToJsonPathOrLiteral(splitter.expr);
+
+          if (ASLGraph.isLiteralValue(valueOut)) {
+            if (!ASLGraph.isLiteralString(valueOut)) {
```
sam-goodwin, line 1997: Nit: use &&

src/asl/synth.ts
```diff
@@ -1974,6 +1974,56 @@ export class ASL {
...
         });
+      } else if (isSplit(expr)) {
+        const [splitter, limit] = expr.args;
+
+        if (limit) {
+          throw new SynthError(
+            ErrorCodes.Unsupported_Feature,
+            "Step Functions does not support the limit parameter on String.split"
+          );
+        } else if (!splitter) {
```
sam-goodwin, line 1985: Can this be supported when it's all literal strings and no references?

src/util.ts
```diff
@@ -286,6 +286,9 @@ export const evalToConstant = (expr: Expr): Constant | undefined => {
     }
   } else if (isPropAccessExpr(expr)) {
     const obj = evalToConstant(expr.expr)?.constant as any;
+    if (typeof obj !== "object") {
```
sam-goodwin, line 289: This will match null as typeof null is "object"

sam-goodwin (approved):

sam-goodwin:
Do we need any documentation in the website for these new features?

src/util.ts
thantos: I actually will remove this, we should be catching the callExpr before evaluating the property access on the string anyways. This was added because it fixed an issue caused by a bug I had (trying to eval the `split` not the value.)

src/asl/synth.ts
thantos: What is the use case where this would actually matter? When the constant is a reference resolvable at runtime? I'll try it.

src/asl/synth.ts
thantos: yes. Typescript/ECMA calls it separator or splitter.

thantos:
> Do we need any documentation in the website for these new features?

We don't currently have any documentation for supported operations. I was going to add a Caveat due to Step Functions removing empty strings from the output.

thantos:
https://github.com/functionless/functionless/issues/517

PR #518, opened 2022-09-20
feat(step-functions): task utility

### LINKED ISSUE #467: Feature: escape hatch for Step Function Tasks
Functionless should offer an escape hatch for StepFunction Tasks so that users can drop down to a lower level to write tasks that functionless does not yet support.

One way this could work is by exposing a `$AWS.Sfn.Task({...})` API. This would treat a Task like any other integration but allow `Resource` and `Parameters` to be specified.

Example:

```typescript
$AWS.Sfn.Task({
  Resource: "arn:aws:states:::apigateway:invoke",
  Parameters: {
    ApiEndpoint: "example.execute-api.us-east-1.amazonaws.com",
    Method: "GET",
    Headers: {
      key: ["value1", "value2"],
    },
    Stage: "prod",
    Path: "bills",
    QueryParameters: {
      billId: ["123456"],
    },
    RequestBody: {},
    AuthType: "NO_AUTH",
  },
});
```

thantos:
Closes #467 
Partial #79 

* Support $SFN.task from #467
   * can invoke any task  with access to (most of) the bare state.
   * Catch, Next, ResultPath, OutputPath, and ResultSelector are computed
   * Parameters, Timeout, and Heartbeat are generated from the simple Parameters input
   * Resource, Retry, Comment can use constant values to populate the task
* Added sendTaskSuccess, sendTaskFailure, sendTaskHeartbeat integrations to step functions, lambda, and appsync.
* Fixed a bug where the context argument (second argument in closure) would break `$SFN.map` and `$SFN.foreach`
* Showed example in tests of using $SFN.task with `StepFunction.sendTaskSuccess` to block and resume a machine from lambda and step functions.
* Added typecheck to pre-commit hook
* Small changes to shaved seconds off the tsc compile time (22s -> ~19s)


src/asl/asl-graph.ts
```diff
@@ -2037,13 +2046,102 @@ export namespace ASLGraph {
...
+    } else if (ASLGraph.isLiteralArray(value)) {
+      return stringifyLiteralArray(value);
+    } else if (ASLGraph.isLiteralObject(value)) {
+      return stringifyLiteralObject(value);
+    }
+    return ASLGraph.isLiteralString(value)
+      ? value
+      : ASLGraph.literalValue(String(value.value));
+
+    function stringifyLiteralObject(
```
sam-goodwin, line 2082: Add some docs. Seems important

src/asl/asl-graph.ts
```diff
@@ -2037,13 +2046,102 @@ export namespace ASLGraph {
...
+      value: ASLGraph.LiteralValue<Record<string, ASLGraph.LiteralValueType>>
+    ): ASLGraph.IntrinsicFunction | ASLGraph.LiteralValue<string> {
+      if (!value.containsJsonPath) {
+        return ASLGraph.literalValue(JSON.stringify(value.value), false);
+      }
+      const entries = Object.entries(value.value).map(([key, value]) => {
+        if (key.endsWith(".$")) {
+          // assuming the these values are
```
sam-goodwin, line 2090: Are what?

src/asl/asl-graph.ts
```diff
@@ -2037,13 +2046,102 @@ export namespace ASLGraph {
...
+            ASLGraph.intrinsicJsonToString(ASLGraph.jsonPath(value as string)),
+          ] as const;
+        }
+        return [key, stringifyLiteral(ASLGraph.literalValue(value))] as const;
+      });
+
+      if (entries.every(([_, value]) => ASLGraph.isLiteralValue(value))) {
+        // just stringify the original value.
+        return ASLGraph.literalValue(JSON.stringify(value.value));
+      }
```
sam-goodwin, line 2102: Can you explain how this works? Looks like a nice optimization but why can you serialize the value.value here

src/asl/asl-graph.ts
```diff
@@ -2037,13 +2046,102 @@ export namespace ASLGraph {
...
+                    : `'${ASLGraph.escapeFormatLiteralString(value)}'`
+                }`
+            )
+            .join(",")}\\}`
+        ),
+        ...entries
+          .map(([_, value]) => value)
+          .filter(ASLGraph.isIntrinsicFunction)
+      );
+    }
```
sam-goodwin, line 2120: Very impressive

src/asl/asl-graph.ts
```diff
@@ -2037,13 +2046,102 @@ export namespace ASLGraph {
...
+      // States.Format('["a", 1, {}{}]', $.var1, $.var2)
+      return ASLGraph.intrinsicFormat(
+        stringifyElements
+          .map((s) =>
+            ASLGraph.isLiteralValue(s)
+              ? `'${escapeFormatLiteralString(s)}'`
+              : "{}"
+          )
+          .join(","),
+        ...stringifyElements.filter(ASLGraph.isIntrinsicFunction)
```
sam-goodwin, line 2141: Are you worried about this being invalid json after formatting?

src/asl/synth.ts
```diff
@@ -2265,7 +2269,11 @@ export class ASL {
             prop: PropAssignExpr | SpreadAssignExpr
           ): ASLGraph.LiteralValue | ASLGraph.JsonPath {
             const output = evalExprToJsonPathOrLiteral(prop.expr);
-            return ASLGraph.isJsonPath(output) ? assignValue(output) : output;
+            return ASLGraph.isJsonPath(output) &&
+              // paths at $$ are immutable
+              !output.jsonPath.startsWith("$$.")
+              ? assignValue(output)
+              : output;
```
sam-goodwin, line 2276: I don't understand the significance of this branch

sam-goodwin (changes requested):

src/asl/asl-graph.ts
thantos: sigh... half thoughts.

assuming that the keys will contain json path and not intrinsic functions.

I am starting to think of a re-write where literal values contain the JsonPath and IntrinsicFunction structs instead of strings and then resolving at the end. That way we don't need the "containsJsonPath", basically leave all resolution to the end.

Also thinking of introducing intrinsic function as an Output.  This would allow for optimal programs because intrinsic functions take 2 states to resolve today, but negative states when they are composed.

for example as array returns an intrinsic function

```ts
{
    v: [input.value]
}
```

```ts
// format the array
{
    Parameters: { "arr.$": "States.Array($.value)" },
    ResultPath: "$.t",
    output: { jsonPath: "$.t.arr" }
}
```

```ts
// create the object with array prop
{
    Parameters: { "v.$": "$.t.arr" }
}
```

But if Array returned an intrinsic function instead...

```ts
{
    output: { funcName: "States.Array", args: [{ josnPath: "$.value" }] }
}
```

```ts
{
    Parameters: { "v.$": "States.Array($.value)" }
}
```

src/asl/asl-graph.ts
thantos: This works off the assumption that the stringify functions only return LiteralValue when all of the children are also literals.

If one value is a jsonPath, nested stringify will start to return interinsicFunctions. 

Will document.

src/asl/asl-graph.ts
thantos: Works well so far... I am sure we'll find some limits at some point...

src/asl/synth.ts
thantos: `evalSafeOutput` references jsonPath values to ensure they do not change between the time when they are supposed to be set in javascript and when they are actually set in stepfunctions.

```ts
let x = 0;
{
    a: x,
    b: (x = 1)
}
``` 

If we don't reference x, it may be 1 at `a` and not 0.

So we do...

```ts
let x = 0;
let v1 = x;
{
    a: v1,
    b: (x = 1)
}
``` 

However, `$$` json paths are the step functions context object which cannot be assigned by the user (and the values we expose do not change between states).

It is safe to use them directly.

```ts
let x = 0;
{
    z: context.Execution.Name,
    a: x,
    b: (x = 1)
}
``` 

becomes  (we don't change the AST, using it for illustration)

```ts
let x = 0;
let v1 = x;
{
    z: context.Execution.Name,
    a: v1,
    b: (x = 1)
}
```

sam-goodwin (approved):

PR #520, opened 2022-09-23
chore: support ast-reflection s-expressions 

thantos:
Required for: https://github.com/functionless/ast-reflection/pull/32

Maintains backwards compatibility with previous register and bind functions.

AST-Reflection is changing from register and bind methods to using s-expression like sequence expressions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,16 +79,16 @@
     "uuid": "^8.3.2"
   },
   "peerDependencies": {
-    "@aws-cdk/aws-appsync-alpha": "*",
     "aws-cdk-lib": "^2.28.1",
     "aws-sdk": "^2",
     "constructs": "^10.0.0",
     "esbuild": "^0.15.6",
     "typesafe-dynamodb": "^0.1.5",
-    "typescript": "^4.8.2"
+    "typescript": "^4.8.2",
+    "@aws-cdk/aws-appsync-alpha": "*"
   },
   "dependencies": {
-    "@functionless/ast-reflection": "^0.3.2",
+    "@functionless/ast-reflection": "^0.3.1",
     "@functionless/nodejs-closure-serializer": "^0.1.2",
     "@swc/cli": "^0.1.57",
     "@swc/core": "1.2.245",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1047,7 +1047,7 @@
     minimatch "^3.1.2"
     strip-json-comments "^3.1.1"
 
-"@functionless/ast-reflection@^0.3.2":
+"@functionless/ast-reflection@^0.3.1":
   version "0.3.2"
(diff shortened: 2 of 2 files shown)


src/function.ts
```diff
@@ -1044,7 +1051,120 @@ export async function serialize(
...
+                                ? p.name.text
+                                : "UNKNOWN",
+                              p.initializer,
+                            ])
+                        );
+                        if (!args || !_this || !func) {
+                          throw new SynthError(
+                            ErrorCodes.Unexpected_Error,
+                            "Compilation Error: found an invalid register command. Check the versions of AST-Reflection and Functionless."
+                          );
```
sam-goodwin, line 1137: Why not just make it a no-op?

sam-goodwin (changes requested):

src/function.ts
thantos: My thought was: The user wouldn't know it is an issue until runtime. If we cannot re-construct the command then we should fail.

sam-goodwin: But how do you know it's not just someone else using similar looking syntax? The compiler should not produce an invalid contract right?

thantos: We check for our flag on the left first. The chance someone has stash_salt=bind is very low. 

That leaves all parsing errors as a bug or mis-matched package version ( with breaking changes)

src/compile.ts
```diff
@@ -88,9 +90,46 @@ export const RegisterFunctionName = `register_${FunctionlessSalt}`;
...
+ * @deprecated - ast-reflection now uses the sequence commands below.
  */
 export const BindFunctionName = `bind_${FunctionlessSalt}`;
 
+/**
+ * AST-Reflection injected s-expression commands.
+ *
+ * (stash=[COMMAND], arguments...)
+ */
+// (stash=REGISTER,stash=func,stash[]=ast,stash)
```
sam-goodwin, line 103: I think a lone "REGISTER" string literals would be less common than an assignment

thantos: That is what I wanted too, but the string gets stripped.

swc-config.js
```diff
@@ -16,8 +16,9 @@ exports.config = {
       plugins: [["@functionless/ast-reflection", {}]],
     },
   },
-  minify: true,
+  minify: false,
```
sam-goodwin, line 19: Did you mean to check this in? It can cause problems?

swc-config.js
```diff
@@ -16,8 +16,9 @@ exports.config = {
...
   sourceMaps: "inline",
+  inlineSourcesContent: false,
```
sam-goodwin, line 21: I think this has an impact where identifiers that have been renamed cannot be re-mapped.

sam-goodwin (approved):

PR #530, opened 2022-10-03
fix: convert to turborepo monorepo

tvanhens:


.github/workflows/release.yml
```diff
@@ -36,7 +35,7 @@ jobs:
       - name: Install dependencies
         run: yarn install --check-files --frozen-lockfile
       - name: release
-        run: npx projen release
+        run: yarn turbo run release --filter=functionless
```
tvanhens, line 38: Release is run via turbo so all the references are resolved as if functionless was still the root.

.husky/pre-commit
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env sh
```
tvanhens, line 1: Since we are no longer using projen, I introduced husky to manage hooks. Husky automatically registers git hooks on `yarn install`.

apps/test-app/package.json
```diff
@@ -1,8 +1,9 @@
 {
-  "name": "test",
+  "name": "test-app",
+  "version": "0.0.0",
   "private": true,
   "scripts": {
-    "build": "tsc && functionless",
+    "typecheck": "tsc --noEmit && functionless",
```
tvanhens, line 6: Renamed for consistency when running with turbo.

apps/test-app/package.json
```diff
@@ -15,10 +16,10 @@
     "@graphql-codegen/typescript": "^2.4.8",
     "@graphql-codegen/typescript-resolvers": "^2.6.1",
     "@types/node": "^14",
-    "aws-cdk-lib": "2.43.1",
     "aws-cdk": "2.43.1",
+    "aws-cdk-lib": "2.43.1",
     "constructs": "10.0.0",
-    "functionless": "file:../",
+    "functionless": "*",
```
tvanhens, line 22: `yarn install` in the root of the project will automatically treat this as a linked dependency.

apps/website/package.json
```diff
@@ -12,7 +12,7 @@
     "serve": "docusaurus serve",
     "write-translations": "docusaurus write-translations",
     "write-heading-ids": "docusaurus write-heading-ids",
-    "typecheck": "tsc"
+    "typecheck-bak": "tsc --noEmit"
```
tvanhens, line 15: These types were not getting checked and there are quite a few errors. I renamed to `-bak` for now so it does not get run by turbo automatically. Will rename back to `typecheck` once the issues are fixed.

packages/core/.eslintrc.js
```diff
@@ -0,0 +1,123 @@
...
+    jest: true,
+    node: true,
+  },
+  root: true,
+  plugins: ["@typescript-eslint", "import", "prettier"],
+  parser: "@typescript-eslint/parser",
+  parserOptions: {
+    ecmaVersion: 2018,
+    sourceType: "module",
+    tsconfigRootDir: __dirname,
```
tvanhens, line 12: Eslint needs the relative root to resolve tsconfig. This required the file to be switched to `*.js` rather than `*.json`.

packages/core/.projen/tasks.json
```diff
@@ -92,35 +89,11 @@
         {
           "exec": "ts-node ./scripts/sdk-gen.ts"
         },
-        {
-          "exec": "yarn link && cd ./test-app && yarn link functionless"
```
tvanhens, line 96: This is no longer needed due to being able to leverage yarn workspaces.

packages/core/.projen/tasks.json
```diff
@@ -211,14 +181,6 @@
         }
       ]
     },
-    "test:app": {
```
tvanhens, line 214: This is run as a separate pipeline step now to benefit from turborepo caching once we're able to optimize the configuration.

.husky/pre-commit
sam-goodwin: Husky always annoyed me because it doesn't align with how I use squash commits. The commits in a pr are irrelevant for the most part. How do you envision the workflow using husky?

apps/test-app/package.json
sam-goodwin: How will this work for a non-private package? The version range could matter and there may be peer dependencies

sam-goodwin (commented):
Looking good

.github/workflows/build.yml
```diff
@@ -31,50 +28,6 @@ jobs:
...
-          git config user.name "github-actions"
-          git config user.email "github-actions@github.com"
-      - name: Push changes
-        run: |2-
-            git add .
-            git commit -s -m "chore: self mutation"
-            git push origin HEAD:${{ github.event.pull_request.head.ref }}
+        run: yarn turbo run build --filter=functionless
+      - name: synth
+        run: yarn turbo run synth --filter=test-app
```
tvanhens, line 33: These all should run together, but doing so requires untangling projen's tasks. Will fix in a followup PR.

.husky/pre-commit
tvanhens: Not sure what you mean. All husky is doing right now is installing an equivalent pre-commit hook to the one projen had set up.

sam-goodwin: What does that hook do? I was referring to hooks that validate the commit title.

tvanhens: It runs `yarn typecheck` and `yarn lint-staged`. This behavior was already present. This was only included to maintain the same functionality. It shouldn't change the workflow.

thantos: Husky just runs hooks, you have to add things to the hooks, we just have lint-staged.

sam-goodwin: Got it. I think I didn't understand what husky did - I've only seen it used to validate commit titles. Makes sense that it's running the linters. No problem

thantos: Husky is fine, do you have it all orchestrated for husky 6? it annoyed me that you need to `init` per user to get husky working after v6.

.github/workflows/build.yml
thantos: What is "these" here?

apps/test-app/package.json
tvanhens: This is outside the scope of turborepo and depends on the versioning/publishing solution we choose.

See [versioning and publishing][1] for more info.

[1]: https://turborepo.org/docs/handbook/publishing-packages/versioning-and-publishing

apps/website/package.json
sam-goodwin: The website doesn't compile? That's strange - I thought docusaurus would validate types

.husky/pre-commit
tvanhens: > Husky is fine, do you have it all orchestrated for husky 6? it annoyed me that you need to `init` per user to get husky working after v6.

Interesting, it just worked when I tried it, didn't have me run init. Let me double check the docs.

thantos:
settings.json should stay at the top?

turbo.json
```diff
@@ -0,0 +1,16 @@
+{
+  "$schema": "https://turborepo.org/schema.json",
+  "pipeline": {
+    "release": {},
+    "build": {},
+    "compile": {
+      "dependsOn": ["^compile"]
+    },
```
thantos, line 8: compile depends on not compile?

.husky/pre-commit
tvanhens: > Husky is fine, do you have it all orchestrated for husky 6? it annoyed me that you need to `init` per user to get husky working after v6.

As far as I can tell, it will auto install, at least with yarn.

.github/workflows/build.yml
tvanhens: Sorry should read "This should run together with the previous step".

turbo.json
tvanhens: > compile depends on not compile?

This threw me for a loop as well. The carrot means for all my dependencies. So a given workspace requires all its dependencies to be compiled first.

https://turborepo.org/docs/core-concepts/running-tasks#in-a-different-workspace

tvanhens:
> settings.json should stay at the top

Yeah settings should. I think `launch.json` makes sense on a package basis though.

sam-goodwin (approved):

apps/website/package.json
tvanhens: It will build but fails type checking. I plan on fixing the issues in a subsequent PR.

thantos (approved):

PR #531, opened 2022-10-03
fix: release task failing due to missing projen dep

tvanhens:
Projen eject does not include the necessary dependencies for the release task.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -69,6 +69,7 @@
     "json-schema": "^0.4.0",
     "npm-check-updates": "^15",
     "prettier": "^2.7.1",
+    "projen": "^0.63.4",
     "promptly": "^3.2.0",
     "proxy-agent": "^5.0.0",
     "standard-version": "^9",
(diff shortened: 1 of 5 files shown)


packages/core/package.json
```diff
@@ -69,6 +69,7 @@
     "json-schema": "^0.4.0",
     "npm-check-updates": "^15",
     "prettier": "^2.7.1",
+    "projen": "^0.63.4",
```
sam-goodwin, line 72: Oh interesting, the ejected projen still has a dependency on itself?

sam-goodwin (approved):

packages/core/package.json
tvanhens: Yeah the release tasks reference built-in functions which are not included in the eject. We can kill this dep after setting up changesets.

thantos: https://github.com/functionless/functionless/actions/runs/3178163655/jobs/5179381959

This seems unrelated to the issue?

tvanhens: At first glance, yes, but the run-tasks script the eject process generates is doing dynamic lookup magic. It is trying to find this definition:
https://github.com/projen/projen/blob/main/src/release/bump-version.task.ts

sam-goodwin (approved):

PR #533, opened 2022-10-04
chore: ts project references and working typescript on all projects

thantos:
* add typescript package references to support a single watch/build of the package.
* website types compile now
* other clean up


apps/website/src/components/home/features/compose/composeFeature.tsx
```diff
@@ -8,9 +8,9 @@ export const ComposeFeature = () => {
   const [codeVisible, setCodeVisible] = useState(false);
   return (
     <FeatureSection
-      feature={compose}
```
thantos, line 11: Did this ever work @cfraz89 ? feature wasn't a field on the object.

apps/website/src/theme/Navbar/Layout/index.tsx
```diff
@@ -1,12 +1,12 @@
 import { useThemeConfig } from "@docusaurus/theme-common";
-import {
+const {
   useHideableNavbar,
   useNavbarMobileSidebar,
-} from "@docusaurus/theme-common/internal";
+} = require("@docusaurus/theme-common/internal");
```
thantos, line 5: docusaurus does not expose types for `/internal` needed to require them.

apps/website/tsconfig.json
```diff
@@ -26,9 +27,10 @@
     "target": "ES2019",
     "declarationMap": true,
     "baseUrl": ".",
-    "types": ["mdx"],
-    "allowJs": true
+    "types": ["mdx", "@docusaurus/theme-classic"],
```
sam-goodwin, line 30: Is this scalable? Do we always have to put everything in here?

package.json
```diff
@@ -11,10 +11,15 @@
...
-    "apps/**/*",
-    "packages/**/*"
-  ],
+  "workspaces": {
+    "packages": [
+      "apps/**/*",
+      "packages/**/*"
+    ],
+    "nohoist": [
+      "**/@types/jest"
```
thantos, line 20: jest/swc failed on the lack of local `@types/jest`, no-hoisting seems to fix.

tsconfig-base.json
```diff
@@ -0,0 +1,33 @@
...
+    "resolveJsonModule": true,
+    "strict": true,
+    "strictNullChecks": true,
+    "strictPropertyInitialization": true,
+    "stripInternal": true,
+    "target": "ES2019",
+    "declarationMap": true,
+    "noUncheckedIndexedAccess": true,
+    "skipLibCheck": true,
+    "types": []
```
sam-goodwin, line 29: Is empty the default?

sam-goodwin (approved):

apps/website/src/theme/Navbar/Layout/index.tsx
cfraz89: I'd keep it as an import, and just @ts-expect-error it

apps/website/src/theme/Navbar/MobileSidebar/Header/index.tsx
```diff
@@ -1,4 +1,4 @@
-import { useNavbarMobileSidebar } from "@docusaurus/theme-common/internal";
+const { useNavbarMobileSidebar } = require("@docusaurus/theme-common/internal");
```
cfraz89, line 1: same here

apps/website/src/theme/Navbar/MobileSidebar/PrimaryMenu/index.tsx
```diff
@@ -1,8 +1,6 @@
-import { isContext } from "vm";
 import Link from "@docusaurus/Link";
-import { useNavbarMobileSidebar } from "@docusaurus/theme-common/internal";
+const { useNavbarMobileSidebar } = require("@docusaurus/theme-common/internal");
```
cfraz89, line 2: ditto

apps/website/tsconfig.json
cfraz89: just for '@docusaurus' imports, they're provided virtually by docusaurus' webpack config

cfraz89 (changes requested):

apps/website/tsconfig.json
thantos: I don't think so, but docusaurus has special magic types, maybe if we stop hoisting this type it will work?

tsconfig-base.json
thantos: I stole this from the typescript package where they also explicitly have this empty. I think default is all `@types`.... will try to remove.

apps/website/tsconfig.json
thantos: Tried to remove, tsc was not happy. Other types work fine.

tvanhens:
Looks like this is breaking the website build. Not sure why just yet.

thantos:
> Looks like this is breaking the website build. Not sure why just yet.

Me neither, it was working before my second commit

tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
```
tvanhens, line 1: Why is this separate from `tsconfig-base.json`? Seems like this one is not connected to the rest of the configs.

thantos: `tsconfig.json` is the root of the project tree.

`tsconfig-base.json` are just common ts-configurations shared between the configs.

You could copy the contents of `tsconfig-base` into each of the `tsconfigs` to see the same result.

sam-goodwin:
Is this ready to merge?

apps/website/src/components/home/features/compose/composeFeature.tsx
cfraz89: Oh you can delete that whole compose folder, its legacy code now

PR #534, opened 2022-10-04
feat: use lerna for versioning and publishing

tvanhens:
Todos:
- [x] Remove test changelog files
- [x] Delete test releases


CHANGELOG.md
```diff
@@ -0,0 +1,26 @@
+# Change Log
```
sam-goodwin, line 1: Can we back date this to all versions?

.github/workflows/release.yml
```diff
@@ -1,18 +1,15 @@
-
```
sam-goodwin, line 1: I think this file is still read only which makes GitHub think it's a generated file. Might want to change that?

.github/workflows/release.yml
```diff
@@ -34,58 +31,10 @@ jobs:
...
-        with:
-          node-version: 14.x
-      - name: Download build artifacts
-        uses: actions/download-artifact@v3
-        with:
-          name: build-artifact
-          path: dist
-      - name: Release
-        run: errout=$(mktemp); gh release create $(cat dist/releasetag.txt) -R $GITHUB_REPOSITORY -F dist/changelog.md -t $(cat dist/releasetag.txt) --target $GITHUB_REF 2> $errout && true; exitcode=$?; if [ $exitcode -ne 0 ] && ! grep -q "Release.tag_name already exists" $errout; then cat $errout; exit $exitcode; fi
+      - run: echo //registry.npmjs.org/:_authToken=${NPM_TOKEN} > .npmrc
```
sam-goodwin, line 34: Just checking: does this result in the token being in the logs? I don't think so but worth a check

sam-goodwin (approved):

.github/workflows/release.yml
tvanhens: It shouldn't and doesn't appear to be from the test runs.

CHANGELOG.md
tvanhens: > Can we back date this to all versions?

Possibly but it will take some investigation.

.github/workflows/release.yml
tvanhens: it was the `.gitattributes` file that projen generated. I removed it and it seems to have fixed the issue.

PR #536, opened 2022-10-05
chore: migrate fl-exp to functionless repo

sam-goodwin:
Fixes #


package.json
```diff
@@ -6,6 +6,7 @@
   "author": "Functionless Team",
   "license": "Apache-2.0",
   "scripts": {
+    "compile": "turbo run compile",
```
tvanhens, line 9: Can remove this. Made `build` the compile task.

packages/fl-exp/package.json
```diff
@@ -0,0 +1,63 @@
...
+  "types": "./lib/index.d.ts",
+  "bin": {
+    "fl": "./bin/cli.js"
+  },
+  "files": [
+    "bin",
+    "lib"
+  ],
+  "scripts": {
+    "compile": "tsc --build",
```
tvanhens, line 14: Can drop the compile task. All packages are consistently using `build` now.

packages/fl-exp/package.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "fl-exp",
```
tvanhens, line 2: Should we rename to `@functionless/fl-exp` to test package deployment? Forget if we transferred ownership of `fl-exp`, think that is still on my account.

packages/fl-exp/package.json
```diff
@@ -0,0 +1,63 @@
...
+    "@types/express": "^4.17.14",
+    "@types/jest": "^29.0.2",
+    "@types/node": "^18.7.18",
+    "@types/toposort": "^2.0.3",
+    "aws-cdk": "^2.43.1",
+    "aws-cdk-lib": "^2.43.1",
+    "aws-sdk": "^2.1215.0",
+    "constructs": "^10.1.111",
+    "esbuild": "^0.15.7",
+    "functionless": "^0.23.1",
```
tvanhens, line 46: This dependency should be `"*"` to pick it up from the workspace.

apps/fl-exp-simple/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "@types/jest": "^29.0.2",
+    "@types/node": "^18.7.18",
+    "@types/uuid": "^8.3.4",
+    "aws-cdk": "*",
+    "aws-cdk-lib": "*",
+    "aws-sdk": "^2.1216.0",
+    "constructs": "^10.1.111",
+    "esbuild": "^0.15.7",
+    "fl-exp": "*",
+    "functionless": "^0.23.1",
```
tvanhens, line 26: Should be `*` to pick up workspace.

packages/fl-exp/package.json
sam-goodwin: Will do - that will also test the deployment of a package that does not exist. I've seen that fail some times so good to check.

packages/fl-exp/package.json
sam-goodwin: All of them or just `functionless`?

apps/fl-exp-simple/package.json
sam-goodwin: All of them or just `functionless`?

packages/fl-exp/package.json
sam-goodwin: Is it wise to have a public facing package use `*` in its dependencies? Or are we centrally managing them somewhere?

apps/fl-exp-simple/package.json
sam-goodwin: Oh i see, these are devDependencies. Looks like we're missing some `dependencies` in fl-exp.

sam-goodwin: Adding this peer to `fl-exp`

```
  "peerDependencies": {
    "functionless": "^0.24.0"
  },
```
  
Will rely on transitive peers from `functionless` for the rest.

sam-goodwin: Actually, why isn't `functionless` a dependency and not a peer?

sam-goodwin: Changing `constructs: "*"` caused a ton of type errors. Had to pin everything to `10.0.0` or else i would end up with `10.111.0` in the root `node_modules` and `10.0.0` in the child `node_modules` causing type errors.

sam-goodwin: Based on my experience, `*`, causes problems with sharing dependencies. Pattern that I see working is to use the same fixed version in `devDependencies`, e.g. `esbuild: "0.15.9"` and then specify a range in the peer dependency, e.g. `^0.15.9`. Using `*` results in the latest version always being installed in the root.

packages/@functionless/fl-exp/src/credentials.ts
```diff
@@ -0,0 +1,17 @@
+import { CredentialProviderChain } from "aws-sdk";
+
+// export function getCredentials() {
+//   return new SsoCredentials({
+//     profile: "sam",
+//   });
+// }
```
thantos, line 7: nit: delete dead code

tvanhens (approved):

PR #537, opened 2022-10-05
chore: CI build cache sharing and improvements

tvanhens:
Fixes #


sam-goodwin (approved):

.github/workflows/build.yml
```diff
@@ -1,31 +1,46 @@
 name: build
 on:
   pull_request: {}
-  workflow_dispatch: {}
 jobs:
-  build:
+  test:
     runs-on: ubuntu-latest
+    concurrency: ci-${{github.job}}-${{ github.ref }}
```
sam-goodwin: what does `ci-...` mean for concurrency?

.github/actions/build/action.yml
```diff
@@ -0,0 +1,35 @@
...
+      run: echo "::set-output name=dir::$(yarn cache dir)"
+      shell: bash
+
+    - uses: actions/cache@v3
+      id: yarn-cache
+      with:
+        path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+        key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-yarn-
```
sam-goodwin, line 17: is the `yarn-` missing a suffic?

.github/actions/build/action.yml
```diff
@@ -0,0 +1,35 @@
...
+
+    - uses: actions/cache@v3
+      id: turbo-cache
+      with:
+        path: .turbo-cache
+        key: turbo-${{ github.sha }}
+        restore-keys: |
+          turbo-
```
sam-goodwin, line 25: curious how these keys/prefixes work? Any docs you can link to?

.github/workflows/build.yml
tvanhens: It's the concurrency key. `ci` is just a static string. I gave it a better name. Only 1 task with a given concurrency key can run at a time. This prevents two test runs in the same branch from clobbering each other.

sam-goodwin (approved):

.github/actions/build/action.yml
tvanhens: `key` is the id of the cache that will be created when the workflow completes with the contents of path.
`restore-keys` is a list of prefixes to restore from. If a cache key matches a restore prefix, it will be loaded. If there are multiple matches, the most recent one will be pulled.

https://github.com/actions/cache

.github/actions/build/action.yml
tvanhens: Nope see below, it is meant to be a prefix. This will pull caches event if they don't match exactly since there is a high likelihood many of the dependencies will be the same.

PR #540, opened 2022-10-06
chore: @functionless/aws, util, aws-util

sam-goodwin:
- [x] refactor `fl-exp/interface` into `@functionless/aws`
- [x] refactor `fl-exp/src/memoize` and `/util` into `@functionless/util` - all non AWS utilities go here
- [x] refactor `fl-exp/src/credentials.ts` into `@functionless/aws-util`  - all dependencies that depend on AWS SDK go here

BREAKING CHANGE: move `fl-exp/interface` to `@functionless/aws`


apps/fl-exp-simple/package.json
```diff
@@ -14,12 +14,13 @@
...
     "@aws-cdk/aws-appsync-alpha": "2.44.0-alpha.0",
     "@functionless/ast-reflection": "^0.3.1",
+    "@functionless/aws": "*",
     "@functionless/aws-constructs": "*",
     "@functionless/fl-exp": "*",
     "@functionless/language-service": "^0.0.4",
     "@functionless/register": "*",
     "@types/jest": "^29.0.3",
-    "@types/node": "^18.7.18",
+    "@types/node": "^18.0.0",
```
thantos, line 23: nit: this version should reflect our minimum engine target (ex: `"^16"`).

If we allow node types to be higher, then we may use features not yet in consumer's node version.

thantos (approved):

apps/fl-exp-simple/package.json
sam-goodwin: Moved them all to `^16`

PR #543, opened 2022-10-06
fix: reset versions in imported packages and run build on PR

tvanhens:
Fixes #


.github/workflows/build.yml
```diff
@@ -38,6 +38,11 @@ jobs:
       - name: Install dependencies
         run: yarn install --check-files
 
+      - name: Build Packages
+        run: yarn turbo run build -- --cache-dir=.turbo-cache
```
sam-goodwin, line 42: Any reason not to run `yarn build` to centralize the command logic?

sam-goodwin (approved):

.github/workflows/build.yml
tvanhens: Fixing this as a part of my caching PR.

PR #547, opened 2022-10-06
chore: create-functionless

sam-goodwin:
Fixes #


tvanhens:
Taking over and moving to #554

PR #549, opened 2022-10-06
chore: move prettier config to project root

tvanhens:
Fixes #


apps/test-app/src/generated-types.ts
```diff
@@ -1,11 +1,24 @@
-import { GraphQLResolveInfo } from 'graphql';
```
sam-goodwin: Might want to ignore this file too. It's generated

tvanhens: Yeah I just rolled back that commit. I'm not going to backfill prettier on already committed files since the lint-staged task will fix them up if they ever get edited again.

package.json
```diff
@@ -28,5 +28,10 @@
     "lerna": "^5.5.4",
     "lint-staged": "^13.0.3",
     "turbo": "^1.5.5"
+  },
+  "lint-staged": {
+    "*.{tsx,jsx,ts,js,md,css,yml,json}": [
+      "prettier --write"
```
thantos, line 34: Why prettier and not eslint? We have rules in eslint like calling out the debugher command and removing `.only` when left in. If we want to decouple them, we can remove the wrapper and run prettier and eslint separately.

thantos (changes requested):

package.json
tvanhens: Goal is to do both. Having trouble getting eslint to work across the project but working on that next.

tvanhens: Right now neither is being run project wide. Prettier just works out of the box so figured it made sense to quickly get that in place while working through eslint.

PR #550, opened 2022-10-07
feat: formation

thantos:
V0.1 of `@functionless/formation`

* [x] Modules Hanlders
* [ ] Hand Rolled Resources (not supported by CC)
   * [ ] Event Bus
      * [x] Create
      * [ ] Update
      * [ ] Delete
   * [ ] Event Bus Rule (not supported by CC)
      * [x] Create
      * [ ] Update
      * [ ] Delete
   * [ ] Policy (not supported by CC)
      * [x] Create
      * [x] Update
      * [ ] Delete
   * [ ] Managed policy (not supported by CC)
      * [x] Create
      * [x] Update
      * [ ] Delete
   * [ ] Queue Policy (not supported by CC)
      * [x] Create
      * [x] Update
      * [ ] Delete
   * [ ] Queue (optimization) 
      * [x] Create
      * [ ] Update
      * [ ] Delete
   * [ ] Role (optimization)
      * [x] Create
      * [ ] Update
      * [ ] Delete
   * [ ] Secret (not supported by CC)
      * [x] Create
      * [ ] Update
      * [ ] Delete
   * [ ] Event Source Mapping (optimization)
      * [x] Create
      * [ ] Update
      * [ ] Delete
   * [ ] more...
* [ ] All Functionless Runtime Tests pass with Formless
    * [x] Lambda
    * [x] Queue - Event Bus Rule was broken
    * [ ] Sfn
    * [ ] Secret
    * [ ] Table
    * [ ] Api
* [ ] Plan
    * [ ] Compute
    * [ ] Render (cli)
* [ ] Improve CLI Output
* [ ] Support Delete
* [ ] Support Statefulness via tags and or stateful saving
* [ ] Bi-furcate consistent vs referenceable completion
* [ ] Metrics
   * [x] Basic (per resource process time, n count, average)
   * [ ] Record
   * [ ] trace logs - https://github.com/wix-incubator/trace-event-lib
* [ ] Support correct Ref logic for all resources - some reasons do not use ARN for REF like Queue (uses QueueUrl)
* [ ] create external package and cli with a really clever name


sam-goodwin:
Not turbo formation?

.projenrc.ts
```diff
@@ -0,0 +1,391 @@
+import * as fs from "fs";
```
sam-goodwin, line 1: Delete this file?

packages/@functionless/ast/src/ensure.ts
```diff
@@ -69,7 +69,7 @@ export function ensure<Assert extends Assertion>(
         .map((kind) =>
           typeof kind === "number" ? getNodeKindName(kind) : kind
         )
-        .join(", ")}`
+        .join(", ")}, found: ${item}`
```
sam-goodwin, line 72: Will this print an object usefully?

packages/@functionless/aws-constructs/test/queue.localstack.test.ts
```diff
@@ -273,7 +264,7 @@ runtimeTestSuite<{
       return {
         queue,
         outputs: {
-          busName: addr.node.addr,
+          busName: bus.resource.eventBusName,
```
sam-goodwin, line 267: Oops lol. How did that ever work

packages/@functionless/aws-constructs/test/runtime.ts
```diff
@@ -385,18 +408,46 @@ export function runtimeTestSuite<
         deployOutputs: s,
       }));
     }
+
+    async function nodeCfnDeploy(
```
sam-goodwin, line 412: Formation deploy?

packages/@functionless/formation/package.json
```diff
@@ -0,0 +1,99 @@
+{
+  "name": "@functionless/formation",
+  "bin": {
+    "node-cfn": "bin/node-cfn.js",
+    "node-cfn.js": "bin/node-cfn.js"
```
sam-goodwin, line 5: Time to rename these?

packages/@functionless/formation/src/aws.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+  async stsClient(options: ClientOptions): Promise<sts.STSClient> {
+    return new sts.STSClient(options);
+  }
+  async ecrClient(options: ClientOptions): Promise<ecr_v2> {
+    return new ecr_v2(options);
+  }
+  async secretsManagerClient(
+    options: ClientOptions
+  ): Promise<secrets_manager_v2> {
```
sam-goodwin, line 52: Should these be memoized ?

packages/@functionless/formation/src/aws.ts
```diff
@@ -0,0 +1,109 @@
...
+  async assumeRole(
+    region: string | undefined,
+    roleArn: string,
+    externalId?: string
+  ): Promise<Credentials> {
+    return new ChainableTemporaryCredentials({
+      params: {
+        RoleArn: roleArn,
+        ExternalId: externalId,
+        RoleSessionName: `node-cfn-${safeUsername()}`,
```
sam-goodwin, line 88: formation

packages/@functionless/formation/src/graph.ts
```diff
@@ -0,0 +1,254 @@
...
+const emptySet = ImmutableSet<string>([]);
+
+function findCircularReferences(graph: DependencyGraph): string[] {
+  const circularReferences: string[] = [];
+  for (const [logicalId, references] of Object.entries(graph)) {
+    for (const reference of references) {
+      const isCircular = transitReference(reference);
+      if (isCircular) {
+        circularReferences.push(logicalId);
+        break;
```
sam-goodwin, line 176: Should you continue and find all?

packages/@functionless/formation/src/resource-provider.ts
```diff
@@ -0,0 +1,140 @@
...
+      resource: PhysicalResource<Properties>;
+    } & ResourceOperationResultMetadata)
+>;
+
+export interface ResourceProviderRetryConfig {
+  canRetry: ("CREATE" | "DELETE" | "UPDATE")[] | true;
+}
+
+/**
+ * TODO: support optional snapshot.
```
sam-goodwin, line 67: What do you mean

packages/@functionless/formation/src/resource-provider.ts
```diff
@@ -0,0 +1,140 @@
...
+  "AWS::IAM::Policy": (props) => new InlinePolicyProvider(props),
+  "AWS::IAM::ManagedPolicy": (props) => new ManagedPolicyProvider(props),
+  "AWS::SQS::QueuePolicy": (props) => new QueuePolicyProvider(props),
+  "AWS::SQS::Queue": (props) => new QueueProvider(props),
+  "AWS::IAM::Role": (props) => new RoleProvider(props),
+  "AWS::SecretsManager::Secret": (props) => new SecretProvider(props),
+  "AWS::CDK::Metadata": new NoOpProvider(),
+  "AWS::Lambda::EventSourceMapping": (props) =>
+    new EventSourceMappingProvider(props),
+  [DEFAULT_RESOURCE_PROVIDER_KEY]: (props) => new CloudControlProvider(props),
```
sam-goodwin, line 103: Would it a be better to have this logic handled explicitly by the caller? Strange to see it as a magic key

packages/@functionless/formation/src/resource-providers/cloud-control.ts
```diff
@@ -0,0 +1,208 @@
...
+export class CloudControlProvider
+  implements ResourceProvider<PhysicalProperties>
+{
+  readonly Type = "AWS::SQS::QueuePolicy";
+  private controlClient: control.CloudControlClient;
+
+  constructor(props: ResourceProviderProps) {
+    this.controlClient = new control.CloudControlClient(props.sdkConfig);
+  }
+  retry: ResourceProviderRetryConfig = { canRetry: true };
```
sam-goodwin, line 34: Set this in the constructor? Make it read only?

packages/@functionless/formation/src/resource.ts
```diff
@@ -0,0 +1,116 @@
+// @ts-ignore - imported for type-doc
+import type {
+  CreateResourceCommand,
+  UpdateResourceCommand,
+} from "@aws-sdk/client-cloudcontrol";
+
+// @ts-ignore - imported for type-doc
+import { Expression } from "./expression";
+// @ts-ignore - imported for type-doc
```
sam-goodwin, line 9: Latest ts versions no longer need this

.projenrc.ts
thantos: hmm, must have come from the merge? deleted

packages/@functionless/aws-constructs/test/queue.localstack.test.ts
thantos: It did work, but the bus was a singleton.

packages/@functionless/formation/package.json
thantos: likely, just copied them over

packages/@functionless/formation/src/aws.ts
thantos: I suspect CDK does memoize them internally, but new options can be passed in on each call, so I didn't memorize them.

packages/@functionless/formation/src/resource-provider.ts
thantos: Some resources support snapshots on delete. My thought was the interface would have an optional `snapshot` callback that can be implemented by a provider when they want to support snapshots.

Additionally, they should be able to force a snapshot on delete for cases like RDS.

packages/@functionless/formation/src/resource-provider.ts
thantos: these are defaults, the user can pass in any map of providers or merge into the defaults. 

What interface were you looking for?

packages/@functionless/formation/src/graph.ts
sam-goodwin: I can comment on my own code too. It's all being committed

packages/@functionless/formation/src/resource-providers/cloud-control.ts
thantos: Moved it to the top and made it readonly. No reason to set in the constructor if it is a constant.

sam-goodwin: Why even have it if it's a constant? I was just thinking for consistency. It's jarring code

packages/@functionless/formation/src/resource-provider.ts
sam-goodwin: I would expect this map to only have resource types. What resource is the default? Is there logic to specially handle that key?

packages/@functionless/formation/src/resource-providers/cloud-control.ts
thantos: Its configuration per provider. Not all providers can idempotently be retried on all operations.

packages/@functionless/formation/src/resource-provider.ts
thantos: Default is the provider we try when there is no explicit key for the current resource in the map. It seemed easier to define the default providers this way, at least until a more complex fallback system was created.

Other options: 
* a map of key to providers and a parameter for "defaultProvider"
* a pipeline approach where a single provider is given which can have providers nested within.
* a map of key to providers then an array of fallback providers.
* a single handler given by the consumer which returns the provider given a key and some config

packages/@functionless/formation/src/graph.ts
thantos: I re-wrote all of this anyways, lol. It was re-doing a lot of work.

sam-goodwin: Can it be removed?

packages/@functionless/ast/src/ensure.ts
thantos: Not always, but it did when I was testing. Will stringify.

sam-goodwin:
> > Not turbo formation?
> 
> 
> 
> My though was to have a nested package (`@functionless/formation`) and then a top level package that gets published `???-formation` (like we are doing with `functionless`).

If we have multiple packages, they should all share the same branding. We need to agree on the branding. I think turbo formation was winning last time we chatted. We can bring it up again with the team

PR #553, opened 2022-10-10
chore: use tsc -b for build and run with turbo

tvanhens:
Fixes #

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,12 +7,13 @@
   "license": "Apache-2.0",
   "scripts": {
     "clean:deep": "git clean -fqdx .",
-    "build:website": "turbo run build --filter=website",
+    "build:website": "turbo run build:website",
     "lint": "turbo run lint",
     "typecheck": "turbo run typecheck",
     "build": "turbo run build",
     "test": "turbo run test",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "dev": "turbo run dev --parallel"
   },
   "workspaces": {
     "packages": [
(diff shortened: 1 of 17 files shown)


package.json
```diff
@@ -7,12 +7,13 @@
   "license": "Apache-2.0",
   "scripts": {
     "clean:deep": "git clean -fqdx .",
-    "build:website": "turbo run build --filter=website",
+    "build:website": "turbo run build:website",
     "lint": "turbo run lint",
-    "typecheck": "turbo run typecheck",
+    "typecheck": "tsc -b",
```
sam-goodwin, line 12: should this have `--noEmit`

packages/@functionless/ast/package.json
```diff
@@ -5,6 +5,7 @@
   "types": "lib/index.d.ts",
   "scripts": {
     "build": "tsc -b",
+    "dev": "tsc --watch",
```
sam-goodwin, line 8: We no longer need this in each of the child packages. We can instead run a single `tsc -w` from the root

sam-goodwin (changes requested):

package.json
tvanhens: The Polaris example has it emit not sure why.

packages/@functionless/ast/package.json
tvanhens: Shouldn't we be building the same way locally that it will build in ci?

sam-goodwin: Not sure it would make a difference? If the references are set up properly, tsc -w would be optimal performance and identical behavior.

sam-goodwin (approved):

PR #567, opened 2022-10-14
chore: remove private dependencies from aggregate packages

sam-goodwin:
@functionless/test (which is a private package) was in our public package's dependencies

- [x] remove `@functionles/test`
- [x] stop running doctor scripts on `yarn install`

### DIFF
--- a/apps/test-app/package.json
+++ b/apps/test-app/package.json
@@ -16,7 +16,7 @@
     "@graphql-codegen/typescript-resolvers": "^2.6.1",
     "@graphql-codegen/typescript": "^2.4.8",
     "@types/node": "^16",
-    "functionless": "^0.27.1",
+    "functionless": "*",
     "aws-cdk-lib": "2.44.0",
     "aws-cdk": "2.44.0",
     "constructs": "10.0.0",
--- a/apps/test-app/package.json
+++ b/apps/test-app/package.json
@@ -16,7 +16,7 @@
     "@graphql-codegen/typescript-resolvers": "^2.6.1",
     "@graphql-codegen/typescript": "^2.4.8",
     "@types/node": "^16",
-    "functionless": "*",
+    "functionless": "^0.28.0",
     "aws-cdk-lib": "2.44.0",
     "aws-cdk": "2.44.0",
     "constructs": "10.0.0",
--- a/apps/website/package.json
+++ b/apps/website/package.json
@@ -63,7 +63,6 @@
     "@functionless/serde": "*",
     "@functionless/serialize-closure": "*",
     "@functionless/swc-config": "*",
-    "@functionless/test": "*",
     "@functionless/util": "*",
     "@functionless/validate": "*",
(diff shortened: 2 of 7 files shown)


apps/test-app/package.json
```diff
@@ -16,7 +16,7 @@
     "@graphql-codegen/typescript-resolvers": "^2.6.1",
     "@graphql-codegen/typescript": "^2.4.8",
     "@types/node": "^16",
-    "functionless": "^0.27.1",
```
thantos, line 19: need to merge tyler's change?

apps/website/package.json
```diff
@@ -104,7 +103,11 @@
     "@types/react": "^17.0.0"
   },
   "browserslist": {
-    "production": [">0.5%", "not dead", "not op_mini all"],
+    "production": [
+      ">0.5%",
+      "not dead",
+      "not op_mini all"
+    ],
```
thantos, line 110: this is becoming a problem, prettier and lerna or some script keep re-formatting json documents.

apps/test-app/package.json
tvanhens, line 19: is this intentional?

sam-goodwin: it is not

scripts/doctor-aggregate-packages.mjs
```diff
@@ -68,6 +68,7 @@ async function patch(pkgJsonPath, pkgJson, packages, dependenciesKey) {
     ...(pkgJson[dependenciesKey] ?? {}),
     ...Object.fromEntries(
       packages
+        .filter((pkg) => pkg.private !== true)
```
thantos, line 71: Should this be based on if the current project is itself private?

apps/test-app/package.json
sam-goodwin: weird, that's not in my local

thantos (changes requested):

apps/test-app/package.json
sam-goodwin: Oh actually, it was intentional - it was causing a problem where differnet versions of functionless were installed from NPM and not fetched from the monorepo

sam-goodwin: Fixed.

apps/website/package.json
sam-goodwin: I disabled them. They will only run manually now.

scripts/doctor-aggregate-packages.mjs
sam-goodwin: That could make sense but the script is really just to automate the docs and aggregate functionless package. We can remove this doctor script probably.

apps/website/package.json
sam-goodwin: This new formatting is what VS code does automatically. We should check it in. It shouldn't auto change any more.

thantos (approved):

